PR #1, opened 2018-07-05
Initial commit of conversion of P4Runtime Spec from GooleDoc format i…

chrispsommers:
…nto Madoko (.mdk). Some additional work remains. The PDF output has a few formatting problems, the HTML looks pretty good. See the README under v1/docs for outstanding items.

This content is as of the shared gDoc spec on around June 20th so any changes made since then need to be hand-merged in.


README.md
```diff
@@ -1,2 +1,30 @@
-# p4runtime
-Specification documents for the P4Runtime control-plane API
+# P4Rruntime Specification
```
antoninbas: typo s/P4Rruntime/P4Runtime
I found this typo a few times in this PR so I recommend sed'ing it / grep'int it

README.md
```diff
@@ -1,2 +1,30 @@
...
+This directory contains protobuf files, specifications and related artifacts for all versions of the P4Runtime API. A separate directory is assigned to each major version, e.g. v1. Therein, tags are used to mark minor and patch release versions.
```
antoninbas: Can we wrap the lines at 80 characters in all the files?

README.md
chrispsommers: ok, thanks

README.md
chrispsommers: I can't convert the whole file to 80-character wide. For  normal "prose" I can do this manually I suppose, but I can't do it to the tables, because Madoko is very fussy about making you put all content for a given row in one line. Using \ as a line-continuation does not seem to work. Perhaps someone knows better than I. The only alternative I am aware of in theses cases is to express the table as HTML, which sort of defeats the purpose. In fact I hand-converted all the HTML output following auto-conversion into markdown tables. Going forward we should consider the efficacy of Madoko tables when  constructing some of our content. Perhaps using native LaTex is the way to go in certain cases.

antoninbas: No worries. I can merge this and I will take care of the 80-character wrap manually as I read the spec. Can you do it now for the non-Madoko files?

chrispsommers: Antonin, I've already done it and am reviewing prior to a new commit. Which non-Madoko files are you thinking of, the READMEs?

antoninbas: Yes the READMEs
I'll review your commit as soon as it is pushed

PR #2, opened 2018-07-17
Add proto definitions for P4Runtime and P4Info

antoninbas:
Copied from p4lang/PI


antoninbas:
@chrispsommers added the protobuf IDLs

chrispsommers:
@antoninbas It's good to get this in the repo. What are the implications of code under PI? Will it live on but migrate to referencing this repo? I guess all the APIs will have to change to the semantically-versioned APIs.

antoninbas:
@wmohsin I updated the PR with the directory structure we discussed. PTAL.

v1/README.md
```diff
@@ -1,4 +0,0 @@
-# P4Runtime Specification Version 1
-
-This directory contains protobuf files, specifications and related artifacts for
```
chrispsommers: Should we modify this text in the top-level README and possibly add some READMEs into the `proto` directory structure at suitable places?
> This directory contains protobuf files, specifications and related artifacts for all versions of the P4Runtime API. A separate directory is assigned to each major version, e.g. v1. Therein, tags are used to mark minor and patch release versions.

antoninbas: Thanks for pointing this out, I'm going to update the README before I merge

PR #7, opened 2018-07-19
Improve syntax highlighting for code blocks

antoninbas:
We do not use P4 as the default language for every code block. We now
have extra syntax files for .proto, .cpp and .pb.txt. The C++ syntax
highlighter was taken from the Madoko github. There is probably a lot of
room for improvement for the proto ones.

I added an FAQ to the docs' README to keep track of some Madoko
"specificities".

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -21,7 +21,6 @@ title,titlenote,titlefooter,authors,h1,h2,h3,h4,h5 {
   font-weight: bold;
 }
 pre, code {
-  language: p4;
   font-family: monospace;
   font-size: 10pt;
 }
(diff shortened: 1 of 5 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -134,6 +165,35 @@ proto {
 }
 }
 
+@if html {
+prototext {
+  replace: "~ Begin Prototextblock&nl;\
+                 ````prototext&nl;&source;&nl;````&nl;\
+                 ~ End Prototextblock";
+  border: solid;
```
wmohsin: I find a solid border to make code snippets it look like tables (obviously, this is very personal taste).
What other option is available?
How hard would it be include the generated html/pdf in the PR to visualize these changes?

wmohsin: SG

PR #11, opened 2018-07-23
Validate Protobuf files as part of CI

antoninbas:
Run protoc on the P4Runtime Protobuf files to ensure they are still
correct in case they are modified by a PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,13 @@
 services:
   - docker
 
+install:
+  - docker build -t p4runtime-ci -f CI/Dockerfile .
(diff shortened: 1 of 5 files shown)


chrispsommers:
Hi Antonin, great idea. Is it possible to run the validation from the command line assuming one has checked out the PR into their workspace (and has docker, of course)? It would be nice to try it without Jenkins to see how it works.  If so, can you provide a sample command? Thanks.

antoninbas:
@chrispsommers done. PTAL.

PR #15, opened 2018-07-25
Clean-up tables in specification

### LINKED ISSUE #12: Clean-up all tables
They are currently not rendering very well, especially in the generated PDF.

antoninbas:
I have given-up on fixing the tables in the generated PDF. It may be
possible by injecting some Tex directly in the Madoko source file, but
it sounds tedious and hard to maintain. So instead, I have cleaned-up
the tables that could be cleaned-up easily (i.e. these that don't have
long lines of text) using the Madoko online editor, and I have replaced
all other tables with lists. I did prefer the tables (in HTML) but the
lists do look okay IMO and they should be easier to modify & maintain.

Fixes #12

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,13 @@ follows:
 
 Git tags are used to mark minor and patch release versions.
 
+## Reading the latest version of the documentation
+
+The HTML for the latest version (working draft) of the P4Runtime v1
+specification is available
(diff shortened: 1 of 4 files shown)


antoninbas:
The built spec for this PR is available here: https://s3-us-west-2.amazonaws.com/p4runtime/travis/antonin/clean-up-spec-tables/P4Runtime-Spec.html

PR #28, opened 2018-08-02
Use Madoko bibliography support for references

### LINKED ISSUE #25: Find a more "automatic" way to manage references
It would be great to have a less manual way of doing this (no repeat of the link, in-order numbering)
I don't know if there is a convenient way to do it in Madoko, maybe we can abuse the bibliography support.

antoninbas:
Madoko uses BibTeX for bibliography which supports online / web
references. We move all the references to a .bib file and include it in
the Madoko source. The end-result looks very nice and references will
definitely be easier to maintain this way.

It seems that dvipng is used to render "math" inside BibTex strings (we
need it to render P4_16 in subscript notation). For people building
locally, dvipng will need to be installed.

Fixes #25

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -117,6 +117,12 @@ extra pass when trying to fit an overfull box. See
 [StackExchange](https://tex.stackexchange.com/a/241355) for more
 information. Not sure what the "optimal" setting is.
 
+### bibref_no_title.js
+
+It is a jQuery script that removes the `title` attribute for bibliography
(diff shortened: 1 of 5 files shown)


antoninbas:
@chrispsommers you make me work hard :)
The titles were not correct anyway so the only solution I found was to remove them altogether (which wasn't super easy)

chrispsommers:
@antoninbas You keep raising the bar for this document! Well done.

PR #30, opened 2018-08-03
Update version number to 1.0.0-rc2 for release

antoninbas:


antoninbas:
I am aware that the link to the tagged protos in the spec is different from other references, but I wanted it this way (I wanted the link to show inline).

antoninbas:
Nope, it's a chicken and egg problem :/
Hopefully I got it right, we will know pretty soon

antoninbas:
I got the link right!
Will update the links on p4.org now.

chrispsommers:
Yup, verified it - nice!

Issue #21, opened 2018-07-30
Include source for diagrams / figures


antoninbas:
So that people can edit them
Do not use images as they are kind of blurry

chrispsommers:
I agree with this idea but it'll take some effort to redraw them. Many of the diagrams (the Ref Arch ones anyway) were created directly in gDoc to solve the "what kind of graphic editor shall we use?" question. The source could not be extracted during conversion to Madoko, so screen captures were done. To abide by a policy to include source, they all need to be redrawn from scratch in an agreed-upon tool. BTW, PowerPoint was used for PSA spec. I'd think we'd prefer to encourage an "open" tool though; PowerPoint is licensed and might exclude some potential contributors. Perhaps between PowerPoint and LibreOffice, and their flexible import/export options, we can cover all the bases.

jnfoster:
TikZ is a really nice LaTeX-based package that can be invoked from Madoko.

It does take some time to code up each figure, but the scripting language
is fairly powerful and you can write macros to avoid having to repeat
common idioms.

The resulting figures are beautiful and don't require proprietary software
to build.

-N

PR #34, opened 2018-08-12
Chrispsommers/redraw figures

chrispsommers:
#21. Redrew all diagrams in OfficeLibre Impress, which is freeware.  The master file is .odg format which can be imported into other programs. It provides basic Visio/PowerPoint-like drawing capabilities. Connector lines are handled quite well which is important for block diagrams. Each image was exported individually in both .png and .svg formats. Madoko's HTML rendering uses .svg and is preferred. LaTex cannot render .svg into the PDF output so .png is specified as a secondary format in the .mdk document source. Note, if need be, the .svg files are importable as master images into other programs.


antoninbas:
A different idea: we can split all the diagrams in different LibreOffice files, then have the Makefile generate the .svg and .png for all of them using the LibreOffice command-line:
```
soffice --headless --convert-to svg test.odg
soffice --headless --convert-to png test.odg
```

chrispsommers:
@antoninbas , thanks for the careful review, I will fix the errors and discrepancies. I also like your idea to render the images during `make`, I will experiment with that and if I'm satisfied I'll make the changes you proposed and leave the incorporation of `soffice` into the build dockerfile up to you.

antoninbas:
Yes, if the `soffice` command line gives satisfactory results, I'm happy to install `soffice` in the docker

antoninbas:
@chrispsommers when you're ready you can use
```
docker run -v `pwd`/docs/v1:/usr/src/p4-spec p4lang/p4rt-madoko:latest make
```
to build the spec. This new docker image will be used for Travis and includes `soffice`. Let me know if the rendered images look good.

chrispsommers:
@antoninbas I think after you merge https://github.com/p4lang/p4runtime/pull/41 this will build OK and you can merge it in. I already rebased previously, hopefully it can be merged as-is.

chrispsommers:
@antoninbas I've rebased and pushed, CI passed, PTAL at the built docs. Thanks!

chrispsommers:
@antoninbas My pleasure, I think this is a good document workflow model to share with others.

Issue #14, opened 2018-07-25
Bytestring length rules and upgrade compatibility issues


teverman:
Under "General principles for message formatting" in the "Bytestrings" subsection, the server is required to return an error if the P4Info-defined bitwidth, rounded to bytes, is not equivalent to the bytestring length in the P4Runtime message.  Enforcement of this requirement leads to awkward upgrade constraints.

Suppose a match field X exists, and X has a P4 program specified width of 9 bits.  Now suppose a new version of the P4 program changes the width to 10 bits.  Since the client and server are likely to upgrade at different times, a 10-bit-aware server will accept requests with a 9-bit X from an older client.  Similarly, a server expecting a 9-bit X will accept client requests with a 10-bit X, subject to the constraint that the client cannot use the full range of X until the server also upgrades to the P4 config with 10-bit X.  The same thing holds if X expands to 11, 12, 13, 14, 15, or 16 bits.

Unfortunately, if at some time X needs to expand to 17 bits, compatibility issues between client and server arise.  If the server upgrades first, it must reject any X from old clients that continue to send X as 2 bytes in a P4Runtime request, even though the request value fits within the P4Info-specified width of 17 bits.  The (p4_bitwidth + 7 / 8) computation will yield an expected byte size of 3 on the server, but the older client will be sending a string of 2 bytes, leading to an INVALID_ARGUMENT error from the server.

As shown above, the current byte string size requirement leads to an anomaly where expansion of bitwidths can occur gracefully across client and server versions as long as the expansion does not cross byte boundaries.  However, once that boundary is crossed, it becomes difficult to gracefully roll out new versions of the P4 program.

An improved approach would require the server to assure that the bytestring provided by the client translates to an integer value that fits within the specified bitwidth.  The server is free to accept bytestrings that appear to be padded with leading zeroes.  The server can also accept bytestrings that omit leading zeroes.

antoninbas:
I think we should discuss this at the meeting next Wednesday.

That seems reasonable to me, but I think the rationale for restricting the size of the value to the byte-width of the P4 field was to ensure we had a canonical representation for these values. We would lose that with this proposed change. AFAIK, the canonical representation was to guarantee read-write symmetry without necessarily having to store all the proto messages at the server.

teverman:
Agree on a meeting discussion.

The requirement to encode the field in a bytestring with length
(p4_bitwidth + 7 / 8) at the sender can persist, with the receiver free to
[...]

PR #36, opened 2018-08-14
Relaxing bytestring constraints

teverman:


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1456,41 +1456,132 @@ types (32-bit and 64-bit words). The P4 language does not put any limit on the
...
-
-For all binary strings, P4Runtime uses big-endian (&ie; network) byte-order. If
-the P4 integer type is not byte-aligned (`p4runtime_binary_string_length !=
-p4_bitwidth * 8`), P4Runtime expects the client to pad the value with unset bits
-before the most-significant bit of the value.
+`bytes` Protobuf type. The correct bitwidth - as per the P4 program - of each
+integer variable exposed through P4Runtime is specified in the P4Info message.
 
+The canonical binary string representation uses the shortest string that
+fits the encoded integer value.  This representation achieves three goals:
```
antoninbas: extra space

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1456,41 +1456,132 @@ types (32-bit and 64-bit words). The P4 language does not put any limit on the
...
+~ End CPP
+
+Binary strings with the byte length computed as `encoded_string_size` promote
+P4Runtime read-write symmetry in both client-to-server requests and
+server-to-client replies.
+
+Upon receiving a binary string, the P4Runtime server does not impose any
+restrictions on the length of the string itself.  Instead, the server verifies
+that the highest-order non-zero bit position in the string is within the
+bounds established by the P4Info bitwidth.  If `p4_bitwidth` is the P4Info-
```
antoninbas: another extra space. It seems to happen after every period?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1456,41 +1456,132 @@ types (32-bit and 64-bit words). The P4 language does not put any limit on the
...
+specified bitwidth for a P4 object value, then the server computes the number of
+binary string bytes required to represent the maximum integer of `p4_bitwidth`
+bits as:
+~ Begin CPP
+binary_string_bytes = (p4_bitwidth + 7) / 8;
+~ End CPP
+The server's decision
```
antoninbas: why the line break?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1456,41 +1456,132 @@ types (32-bit and 64-bit words). The P4 language does not put any limit on the
...
+  to the P4Info-specified bitwidth.
+
+* It supports [read-write symmetry](#sec-read-write-symmetry).
+
+* It helps facilitate non-disruptive P4 program updates.
+
+The shortest possible binary string for an integer value with `high_bit`
+as the most-significant non-zero bit in the value is computed as:
+~ Begin CPP
+encoded_string_size = (high_bit + 8) / 8;
```
antoninbas: if the value is 0, what is `high_bit`?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1456,41 +1456,132 @@ types (32-bit and 64-bit words). The P4 language does not put any limit on the
...
+  `binary_string_bytes`, then the server must also consider the P4 integer
+  type's byte alignment.  Byte-aligned strings that satisfy the expression:
+~ Begin CPP
+  (binary_string_bytes * 8 == p4_bitwidth)
+~ End CPP
+  are always valid.  For non-aligned strings, the server requires
+  the client to fill the high-order pad bits with zeroes, and the server
+  rejects any binary strings with non-zero pad bits.
+
+* If the string's byte length is zero, the server always rejects the string.
```
antoninbas: Yes I think this is better than an empty string.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1456,41 +1456,132 @@ types (32-bit and 64-bit words). The P4 language does not put any limit on the
...
-| `int<8>`  | 99 (0x63)      | `\x63`                  |
-| `int<8>`  | -99 (-0x63)    | `\x9d`                  |
-| `int<12>` | -739 (-0x2e3)  | `\x0d\x1d`              |
-| `int<16>` | 0 (0x0)        | `x00\x00`               |
-|-----------|----------------|-------------------------|
+complement bitwise representation as P4.  Table [#tab-valid-bytestring-encoding]
+shows various examples of integer values that the server accepts as valid
+P4Runtime binary strings according to the criteria in the list above.
+
+~ TableFigure { #tab-valid-bytestring-encoding; \
```
antoninbas: I checked that table and it renders fine in PDF, although we are almost at the limit :)

docs/v1/P4Runtime-Spec.mdk
antoninbas: @teverman so I just found out that double spacing was a thing, sorry about that
I don't know if we should try to keep thing uniform at this level, it doesn't matter in the great scheme of things :)

teverman: I'm definitely old school here out of habit :-)  Since the document is right-justified, I can't tell if Madoko eats the extra space upon conversion to PDF, but there are a few places where it looks like it does.

docs/v1/P4Runtime-Spec.mdk
teverman: I think I rewrote or did some editing on my original version of this statement and never bothered to join the text lines following the edit.

docs/v1/P4Runtime-Spec.mdk
teverman: Good point - will address in next commit.

docs/v1/P4Runtime-Spec.mdk
teverman: Yes - I rendered in PDF before creating the pull request.

teverman:
Agree about impact on 9.1.1.  I'll fix it, but I need some clarification first.

I find the beginning of the second paragraph under 9.1.1 to be misleading, or I am misunderstanding:

"For "don't care" matches, the P4Runtime client must omit the match value when
building the match repeated field of the `TableEntry` message."

Taken literally, this can be interpreted as "don't care" matches need a match entry with field_id, but the value field of FieldMatch is not present.  Isn't this more correctly stated as:

"For "don't care" matches, the P4Runtime client must omit the field's entire `FieldMatch` entry when
building the `match` repeated field of the `TableEntry` message."

This is how we are implemented today, and IIRC I looked at bmv2 first for consistency.

Maybe it helps to clarify with an example.  Suppose a table has two match fields, field ID 1 is TERNARY, and field ID 2 is EXACT.  To insert a table entry with ternary field ID 1 as a don't care, the repeated match field has no entry for field_id == 1:

    table_entry {
      table_id: 33554439
      match {
        # field_id 1 is not present to use don't care ternary value.
        field_id: 2
        exact {
          value: "\x60"
        }
      }
      action {
        action_profile_member_id: 20
      }
    }

antoninbas:
@teverman good point. Yes your phrasing is much better and it would be great if you could include your example when you edit the section (maybe add the corresponding P4 table declaration to accompany the P4Runtime message).

antoninbas:
Fixes #14

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1456,41 +1456,134 @@ types (32-bit and 64-bit words). The P4 language does not put any limit on the
...
+| `bit<12>` | `\x10\x63`              |
+| `bit<12>` | `\x01\x00\x63`          |
+| `bit<12>` | `\x00\x40\x63`          |
+| `int<8>`  | `\xff\x9d`              |
+| `int<12>` | `\x8d\x1d`              |
+| `int<16>` | `empty string`          |
+|-----------|-------------------------|
 ~
 
+As the preceding examples illustrate, a P4Runtime server must accept a wide
```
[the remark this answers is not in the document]
antoninbas: I may be wrong but I think the idea was to be backward-compatible with exiting implementations. @teverman ?

teverman:
Yes, Antonin.  That is the problem.  If my changes were for a brand new
document with no prior implementations, we wouldn't need to be friendly
toward leading padding bytes.

docs/v1/P4Runtime-Spec.mdk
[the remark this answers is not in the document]
konne-google: Makes sense. Thanks for the clarification!

konne-google:
Makes sense. Thanks for the clarification!

PR #46, opened 2018-09-13
Clarify server behavior for malformed actions

### LINKED ISSUE #33: Clarify server behavior for malformed actions
Most of the spec defines a specific error for the server to return upon encountering protocol violations by the client.  Section 9.1.2 "Action Specification" does not.  The spec should define specific responses for these situations:

TableEntry INSERT or MODIFY is missing an expected Action ID, action profile member or group ID.

Action ID does not belong to the P4Info set for the table.

Action parameter is missing from the P4Runtime action.

Parameter ID is invalid according to the P4Info.

Parameter value is missing.

antoninbas:
Fixes #33

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2009,7 +2009,7 @@ pair of table entries that the same packet matches.
 The `match` and `priority` fields are used to uniquely identify an entry within
 a table. Therefore, these fields cannot be modified after the entry has been
 inserted and must be provided for `MODIFY` and `DELETE` updates. When deleting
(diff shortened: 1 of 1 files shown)


antoninbas:
@teverman couldn't log into Github so he sent me the following feedback by email:

> My only comment concerns this text for action parameter values.
"if a parameter value is missing, or if the value provided for one of the parameters does not have the correct format"
Since parameter values are protobuf bytes type, I don't know what it means to have an invalid parameter format, but I suggest rewording to say "if the value provided for one of the parameters does not conform to <insert bytestring reference here>"

I'm going to update this PR based on his feedback and proceed with merging.

Issue #49, opened 2018-09-13
Fix incorrect statement in spec that device port number 0 is invalid in PSA


antoninbas:
I don't believe that PSA has that restriction and I actually found an example in the PSA spec that mentions port 0 as a valid port. While it makes sense that the SDN port 0 be invalid, there is no reason for device port 0 to be invalid. I know of hardware targets for which it is not the case.

jafingerhut:
At least for port ids, that was the original motivating use case for numeric translation between control plane and data plane values in a P4Runtime server, so making it valid in the data plane but invalid in the control plane is straightforward.

PR #50, opened 2018-09-13
Tentative support for new (user-defined) types in P4Info

antoninbas:
We update P4DataTypeSpec to be able to represent user-defined types in
P4Info. Translation (when the control-plane view differes from the
underlying data-plane type) may be required for new types through the
@p4runtime_translation annotation and this information is preserved in
P4Info. As decided previously, we do not break backward-compatibility
for action parameters and match fields. For these we do not use
P4DataTypeSpec but instead we add a new Protobuf field ("type_name")
which is set if the action parameter / match field is a user-defined
type (which must map to an underlying bit vector in this case).

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -365,6 +365,9 @@ The following are not in scope of this specification document:
 * target
   : The hardware or software entity which "executes" the P4 pipeline and hosts
     the P4Runtime Service; often used interchangeably with "device".
+* URI
+  : Uniform Resource Identifier; a string of characters designed for unambiguous
+    identification of resources.
 
 # Reference Architecture { #sec-reference-architecture}
 Figure [#fig-reference-architecture] represents the P4Runtime Reference
@@ -971,6 +974,10 @@ control plane. This message contains the following fields:
 
     * `doc`, a `Documentation` message describing this match field.
 
+    * `type_name`, which indicates whether the action parameter has a
+      [user-defined type](#sec-user-defined-types); this is useful for
+      [translation](#sec-psa-metadata-translation).
+
 * `action_refs`, a repeated `ActionRef` field representing the set of possible
   actions for this table. The `ActionRef` message is used to reference an action
   specified in the same P4Info message and it includes the following fields:
@@ -1041,6 +1048,9 @@ The `Action` message defines the following fields:
       annotation associated to this parameter.
     * `bitwidth`, an `int32` value set to the size in bits of this parameter.
     * `doc`, which describes this parameter using a `Documentation` message.
+    * `type_name`, which indicates whether the action parameter has a
+      [user-defined type](#sec-user-defined-types); this is useful for
+      [translation](#sec-psa-metadata-translation).
 
 ### `ActionProfile`
 
@@ -1697,7 +1707,8 @@ client can provide a value for a P4~16~ expression. `P4DataTypeSpec` describes
 the compile-time of the expression as a Protobuf `oneof`, which can be:
 
 * a string representing the name of the type in case of a named type (`struct`,
-  `header`, `header_union`, `enum` or `serializable_enum`),
+  `header`, `header_union`, `enum`, `serializable_enum` or user-defined "new"
+  `type`),
 
 * an empty Protobuf message for `bool` and `error`, or
 
@@ -1844,15 +1855,93 @@ entry value. When providing serializable enum values through `P4Data`, one can
 either use the enum entry's name (`enum` human-readable string field) or its
 assigned value (`enum_value` bytestring field).
 
+### User-defined types { #sec-user-defined-types}
+
+P4~16~ enables programmers to introduce new types [@P4NewTypes]. While similar
+to `typedef`, this mechanism introduces in fact a new type, which is not
+strictly synonym with the original type. It is important to preserve this
+distinction in the P4Info message, in particular for the purposes of
+[translation](#sec-psa-metadata-translation). When introducing a new type, the
+declaration can be annotated with `@p4runtime_translation` to indicate that the
+type exposed to the P4Runtime client is different from the original P4 type. One
+important use-case if for [port numbers](#sec-translation-of-port-numbers),
+whose underlying dataplane representation may vary on different targets, but for
+which it may be convenient to present a unified representation and numbering
+scheme to the control-plane. **The `@p4runtime_translation` annotation can only
+be used if the underlying P4 built-in type is a fixed-width unsigned bitstring
+type (`bit<W>`)** and the type exposed to the control-plane will also be a
+fixed-width unsigned bitstring, with a potentially different bitwidth. It takes
+two parameters: a *URI* (Uniform Resource Identifier) which uniquely identifies
+the translation being performed on entities of the new type to the P4Runtime
+server and the *bitwidth* of the bitstring type exposed to the control-plane. It
+is recommended that the URI includes at least the P4 architecture name and the
+type name.
+
+User-defined types are specified using the `P4NewTypeSpec` message, which has
+the following fields:
+
+* `original_type`, which specifies the underlying type used in the P4 `type`
+  declaration, using the generic `P4DataTypeSpec` message. The `original_type`
+  can correspond to a P4~16~ built-in type or to another user-defined type.
+
+* `translation` of type `P4NewTypeTranslation`, which is set if and only if the
+  P4 type declaration was annotated with `@p4runtime_translation`. The
+  `P4NewTypeTranslation` message itself has two fields - `uri` and
+  `sdn_bitwidth`, which map to the two input parameters to the annotation.
+
+* `annotations`, a repeated field of strings, each one representing a P4
+  annotation associated to the type declaration.
+
+For example, an architecture - in this case PSA - may introduce a new type for
+port numbers:
+~ Begin P4Example
+@p4runtime_translation("p4.org/psa/v1/PortId_t", 32)
+type bit<9> PortId_t;
+~ End P4Example
+
+In this case, the P4Info message would include the following `P4TypeInfo`
+message:
+
+~ Begin Prototext
+type_info {
+  new_types {
+    key: "PortId_t"
+    value {  # P4NewTypeSpec
+      original_type {  # P4DataTypeSpec
+        type_spec {
+          bit {
+            bitwidth: 9
+          }
+        }
+      }
+      translation {  # P4NewTypeTranslation
+        uri: "p4.org/psa/v1/PortId_t"
+        sdn_bitwidth: 32
+      }
+    }
+  }
+}
+~ End Prototext
+
+Note that a P4 compiler may provide a mechanism external to the language to
+specify if and how a user-defined type is to be translated (e.g. through some
+configuration file passed on the command-line when invoking the compiler). This
+mechanism should take precedence over `@p4runtime_translation` to enable users
+to overwrite annotations included as part of the P4 architecture definition.
+
 ### Trade-off for v1.0 Release
 
 For the v1.0 release of P4Runtime, it was decided not to replace occurrences of
-`bytes` with `P4Data` in the `p4.FieldMatch` message, which is used to represent
-table and Value Set entries. This is to avoid breaking pre-release
+`bytes` with `P4Data` in the `p4.v1.FieldMatch` message, which is used to
+represent table and Value Set entries. This is to avoid breaking pre-release
 implementations of P4Runtime. Similarly it has been decided to keep using
 `bytes` to provide action parameter values. However `P4Data` is used whenever
 appropriate for PSA externs and we encourage the use of `P4Data` in
 architecture-specific extensions.
+
+In order to support [translation](#sec-psa-metadata-translation) for action
+parameters and match fields, we include a `type_name` field in
+`p4.config.v1.MatchField` and `p4.config.v1.Action.Param`.
 
 # P4 Entity Messages { #sec-p4-entity-msgs}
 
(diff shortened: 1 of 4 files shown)


proto/p4/config/v1/p4types.proto
```diff
@@ -183,4 +185,22 @@ message P4ErrorTypeSpec {
...
+  // P4Runtime agent to perform value-mapping appropriately when required. It is
+  // recommended that the URI includes at least the P4 architecture name and the
+  // type name.
+  string uri = 1;
+  int32 sdn_bitwidth = 2;
+}
+
+// New types introduced with the "type" keyword
+message P4NewTypeSpec {
+  P4DataTypeSpec original_type = 1;
```
antoninbas: One potential issue is that `original_type` may be different across different targets / backends (e.g. if the device port representation is different). To keep P4Info target independent, it may be better to omit `original_type` for translated entities.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1844,16 +1855,94 @@ entry value. When providing serializable enum values through `P4Data`, one can
 either use the enum entry's name (`enum` human-readable string field) or its
 assigned value (`enum_value` bytestring field).
 
+### User-defined types { #sec-user-defined-types}
+
+P4~16~ enables programmers to introduce new types [@P4NewTypes]. While similar
+to `typedef`, this mechanism introduces in fact a new type, which is not
+strictly synonym with the original type. It is important to preserve this
```
jafingerhut: grammar typo: 'not strictly synonym' -> 'not a strict synonym' or perhaps 'not strictly a synonym'

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1844,16 +1855,94 @@ entry value. When providing serializable enum values through `P4Data`, one can
...
+distinction in the P4Info message, in particular for the purposes of
+[translation](#sec-psa-metadata-translation). When introducing a new type, the
+declaration can be annotated with `@p4runtime_translation` to indicate that the
+type exposed to the P4Runtime client is different from the original P4 type. One
+important use-case if for [port numbers](#sec-translation-of-port-numbers),
```
jafingerhut: typo: 'if' -> 'is'

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1844,16 +1855,94 @@ entry value. When providing serializable enum values through `P4Data`, one can
...
+whose underlying dataplane representation may vary on different targets, but for
+which it may be convenient to present a unified representation and numbering
+scheme to the control-plane. **The `@p4runtime_translation` annotation can only
+be used if the underlying P4 built-in type is a fixed-width unsigned bitstring
+type (`bit<W>`)** and the type exposed to the control-plane will also be a
+fixed-width unsigned bitstring, with a potentially different bitwidth. It takes
```
jafingerhut: I guess the restriction to `bit<W>` types might be generalized in later P4Runtime API spec versions after 1.0?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1844,16 +1855,94 @@ entry value. When providing serializable enum values through `P4Data`, one can
...
+server and the *bitwidth* of the bitstring type exposed to the control-plane. It
+is recommended that the URI includes at least the P4 architecture name and the
+type name.
+
+User-defined types are specified using the `P4NewTypeSpec` message, which has
+the following fields:
+
+* `original_type`, which specifies the underlying type used in the P4 `type`
+  declaration, using the generic `P4DataTypeSpec` message. The `original_type`
+  can correspond to a P4~16~ built-in type or to another user-defined type.
```
jafingerhut: As I think Antonin commented elsewhere, at least for the current 7 PSA types that we want to allow to have bit widths that can vary across targets, including `original_type` for those would conflict with the goal of making the P4Info file the same for all targets.

The set of types with that property we expect to change very slowly, if at all (on the time scale of releases of new versions of the PSA spec).

One possibility: For such types define a special value for the `original_type` field that is easily distinguishable from all other type definitions, which means "target-dependent", with no further details.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1844,16 +1855,94 @@ entry value. When providing serializable enum values through `P4Data`, one can
...
+
+* `translation` of type `P4NewTypeTranslation`, which is set if and only if the
+  P4 type declaration was annotated with `@p4runtime_translation`. The
+  `P4NewTypeTranslation` message itself has two fields - `uri` and
+  `sdn_bitwidth`, which map to the two input parameters to the annotation.
```
jafingerhut: I'm not a great one for coming up with names, but would `p4runtime_bitwidth` be appropriate here instead of `sdn_bitwidth`?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1844,16 +1855,94 @@ entry value. When providing serializable enum values through `P4Data`, one can
...
+
+* `annotations`, a repeated field of strings, each one representing a P4
+  annotation associated to the type declaration.
+
+For example, an architecture - in this case PSA - may introduce a new type for
+port numbers:
+~ Begin P4Example
+@p4runtime_translation("p4.org/psa/v1/PortId_t", 32)
```
jafingerhut: I'm not deeply familiar with URIs, but at least the Wikipedia page makes it appear that the "scheme:" part of it is required at the beginning.  I have no strong opinion for whether the p4runtime_translation parameter needs a scheme included, or whether it should be assumed to always be "http://"

proto/p4/config/v1/p4types.proto
jafingerhut: I mentioned this in another comment, but also replying here to keep it with the previous comment: perhaps define a unique and easily recognized value for P4DataTypeSpec that means "target-dependent"?  So far as I know right now that value would only be legal in the context of the value of the `original_type` key for one of the PSA target-dependent type names, and would be nonsense to use that value in most or all other contexts.

antoninbas: I thought it might make more sense to use a union (Protobuf `oneof`) so that we either include `original_type` (in the absence of translation) or `translated_type` (in the case of translation). I have made the change.

docs/v1/P4Runtime-Spec.mdk
antoninbas: I tried to emulate the notion of URL for a Protobuf `Any` message (https://developers.google.com/protocol-buffers/docs/proto3#any). I believe URI is more appropriate than URL for this, since for the `Any` message they seem to use the URL purely as an identifier.
We can probably be a bit liberal and use URI interchangeably with "URI suffix" in the spec. According to the same Wikipedia page, this seems to be a common practice.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Nothing off the top of my head.  I was more asking the question aloud whether we are thinking we are restricting ourselves from allowing it in the future.  I am pretty sure we are not restricting ourselves in that way, if/when someone comes up with a strong enough use case.

I do think that it is more likely that a P4 `type` might be used in the future with the underlying type being a struct, with no @p4runtime_translation annotation, e.g. for being able to name sub-fields of a larger value.  Again, as long as the P4 Info can be generalized so that original_type elsewhere in this proposal can be any P4 type we want to be able to handle, no problem.  Also not urgent for P4Runtime version 1.0.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Thanks for checking.  Seems fine to me.

antoninbas:
@jafingerhut Could I get a second review on this? Please let me know if there is something I didn't reply to or address. Thanks.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -971,6 +974,10 @@ control plane. This message contains the following fields:
 
     * `doc`, a `Documentation` message describing this match field.
 
+    * `type_name`, which indicates whether the action parameter has a
```
jafingerhut: typo?  'action parameter' -> 'match field'

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1697,7 +1707,8 @@ client can provide a value for a P4~16~ expression. `P4DataTypeSpec` describes
 the compile-time of the expression as a Protobuf `oneof`, which can be:
 
 * a string representing the name of the type in case of a named type (`struct`,
-  `header`, `header_union`, `enum` or `serializable_enum`),
+  `header`, `header_union`, `enum`, `serializable_enum` or user-defined "new"
+  `type`),
```
jafingerhut: This comment is not for this change, but for original line 1697, new line 1707 above where it says "the compile-time of".  Maybe that should be "the compile-time type of" ?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1844,16 +1855,100 @@ entry value. When providing serializable enum values through `P4Data`, one can
...
+  new_types {
+    key: "PortId_t"
+    value {  # P4NewTypeSpec
+      original_type {  # P4DataTypeSpec
+        type_spec {
+          bit {
+            bitwidth: 9
+          }
+        }
+      }
```
jafingerhut: Sorry if I missed this in my first review, but if the goal is to have P4Info contents be identical for all P4 target devices, isn't the presence of `original_type` going to conflict with that goal?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3942,18 +4037,26 @@ In order to support the above SDN use case, P4Runtime requires translation of
...
-for all P4 entities of the special PSA port types use a controller-defined
-32-bit type instead of the dataplane bitwidth defined in the P4 program. The
-following PSA port metadata types are defined in *psa.p4* for the PSA device in
-Switch 1.
+purpose, PSA defines the port metadata field type using special [user-defined P4
+types](#sec-user-defined-types), namely `PortId_t` and `PortInHeader_t`, instead
+of standard P4 bitstrings. The P4Info for all P4 entities of the special PSA
+port types use a controller-defined 32-bit type instead of the dataplane
+bitwidth defined in the P4 program. The following PSA port metadata types are
+defined in *psa.p4* for the PSA device in Switch 1.
```
jafingerhut: If I am reading this correctly, this is saying that all of the PSA custom width types will be 32 bits wide in P4Runtime messages.  I guess I had been assuming that the "*InHeader_t" bit widths in the latest working PSA spec would be the P4Runtime widths, and those are not all 32 bits wide.  The timestamp one doesn't even fit in 32 bits, but the rest of them do, so no big deal (except for the timestamp) if they all end up being 32 bits in P4Runtime messages.

Here is a link to the PSA spec Madoko source for quick reference: https://github.com/p4lang/p4-spec/blob/master/p4-16/psa/psa.p4#L75-L92

Note especially the comment block just before those types, which may be wrong and need updating, which I'd like to do when we decide the desired approach:
```
/* Note: All of the widths for types with `InHeader` in their name are
 * intended only to carry values of the corresponding types in packet
 * headers between a controller and a PSA device.  The widths are
 * intended to be large enough for all PSA devices, so that a
 * controller can fill in headers with these fields in a common way
 * for all PSA devices.
 *
 * All widths must be a multiple of 8, so that any subset of these
 * fields may be used in a single P4 header definition, even on P4
 * implementations that restrict headers to contain fields with a
 * total length that is a multiple of 8 bits. */
typedef bit<32> PortIdInHeader_t;
typedef bit<32> MulticastGroupInHeader_t;
typedef bit<16> CloneSessionIdInHeader_t;
typedef bit<8>  ClassOfServiceInHeader_t;
typedef bit<16> PacketLengthInHeader_t;
typedef bit<16> EgressInstanceInHeader_t;
typedef bit<64> TimestampInHeader_t;
```

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3942,18 +4037,26 @@ In order to support the above SDN use case, P4Runtime requires translation of
...
 as needed. Such translation is enabled by identifying a P4 entity (match field,
 action parameter or header field) as being a PSA port metadata type. For this
-purpose, PSA defines the port metadata field type using special P4 types, namely
-`PortId_t` and `PortInHeader_t`, instead of standard P4 bitstrings. The P4Info
-for all P4 entities of the special PSA port types use a controller-defined
-32-bit type instead of the dataplane bitwidth defined in the P4 program. The
-following PSA port metadata types are defined in *psa.p4* for the PSA device in
-Switch 1.
+purpose, PSA defines the port metadata field type using special [user-defined P4
+types](#sec-user-defined-types), namely `PortId_t` and `PortInHeader_t`, instead
```
jafingerhut: typo: `PortInHeader_t` -> `PortIdInHeader_t`.  Yeah, that type name is unwieldy, but at least it is only in to/from-controller packet headers :)

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3942,18 +4037,26 @@ In order to support the above SDN use case, P4Runtime requires translation of
...
+@p4runtime_translation("p4.org/psa/v1/PortIdInHeader_t", 32)
 type PortIdInHeader_t bit<32>;
 ~ End P4Example
 
+The first argument to the `@p4runtime_translation` annotation is a URI that
+indicates to the P4Runtime server which numerical mapping - provided by the
+out-of-band switch configuration mechanism - to use to translate between the SDN
+value and the dataplane value. The second argument is the bitwidth of the SDN
+representation of the transltaed entity (32-bit in the case of ports).
+
```
jafingerhut: Do we have a separate open issue to track the comment below of "PSA device port number 0 ... invalid"?  I know your brought this up at a P4 API work group meeting.  I believe the _data plane_ should allow ingress and egress port number 0 to be valid, even if the P4Runtime messages use numeric translation so that 0 is considered invalid there.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Actually, doesn't this example violate the comments mentioned later that at most one of original_type and translated_type should be present?  Should "translated" below be "translated_type"?

jafingerhut:
@antoninbas I have done a second review now.  I've been more thorough with the text than the Google Protobuf code.

docs/v1/P4Runtime-Spec.mdk
antoninbas: Yes, we have an issue, you even commented there :)
https://github.com/p4lang/p4runtime/issues/49

docs/v1/P4Runtime-Spec.mdk
antoninbas: It may not be super clear the way Samar wrote it, but it does say "port types"; it only applies to `PortId_t` and `PortIdInHeader_t`.

docs/v1/P4Runtime-Spec.mdk
antoninbas: My bad. I updated the text and the .proto, but not the example.

antoninbas:
@jafingerhut thanks for the thorough review. PTAL and approve if I didn't miss anything so I can merge.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: I think I was maybe reading a bit too quickly and missed the "port types", but they are mentioned and no need to change anything further here that I can see.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: They say that memory is the second thing to go :)

PR #59, opened 2018-09-21
Adds ability to associate a cookie to the forwarding-pipeline config

ccascone:
This change also allow a client to verify the config on the target by asking the server to return only the cookie, minimizing the amount of data in the response message (vs. returning the full P4 blob).


proto/p4/v1/p4runtime.proto
```diff
@@ -639,10 +639,26 @@ message ForwardingPipelineConfig {
...
 message GetForwardingPipelineConfigRequest {
+  // Specifies the fields to populate in the response.
+  enum ResponseType {
+    // Default behaviour. Returns a ForwardingPipelineConfig with all fields set
+    // as stored by the target.
+    ALL = 0;
+    // Reply by setting only the cookie field, omitting all other fields.
+    COOKIE_ONLY = 1;
+    // Reply by setting only the p4info field.
+    P4INFO_ONLY = 2;
```
antoninbas: Maybe this could return the cookie as well?

proto/p4/v1/p4runtime.proto
```diff
@@ -639,10 +639,26 @@ message ForwardingPipelineConfig {
   config.v1.P4Info p4info = 1;
   // Target-specific P4 configuration.
   bytes p4_device_config = 2;
+  // Metadata (cookie) opaque to the target. A control plane may use this field
+  // to uniquely identify this config. There are no restrictions on how such
+  // value is computed, or where this is stored on the target, as long as it is
+  // returned with a `GetForwardingPipelineConfig` RPC.
```
antoninbas: Remove ticks here

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3778,10 +3786,10 @@ The action is the type of configuration action requested, it can be one of:
   target.
 
 The `config` field is a message of type `ForwardingPipelineConfig` that carries
-the P4Info and the opaque target-dependent forwarding configuration data or
-device config, generated by the P4 compiler for the target. See the
-[Forwarding-Pipeline Configuration](#sec-p4-fwd-pipe-config) section for
-details.
+the P4Info, the opaque target-dependent forwarding-pipeline configuration data
+(e.g., generated by the P4 compiler for the target), and the cookie to uniquely
```
antoninbas: remove comma after e.g.
use `&eg;` in the madoko source

antoninbas: I would remove the comma before and

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3795,34 +3803,56 @@ the `GetForwardingPipelineConfig RPC`. The request is defined as:
...
+* `ALL`: returns a `ForwardingPipelineConfig` with all fields
+  set as stored by the target. This is the default behaviour if the
+  `response_type` field is not set.
+
+* `COOKIE_ONLY`: reply by setting only the `cookie` field in the
+  `ForwardingPipelineConfig`, omitting all other fields. This mechanisms can be
+  used by a controller to verify that a config is the expected one, while
+  minimizing the amount of data in the response message.
+
+* `P4INFO_ONLY`: reply by setting only the `p4info` field.
```
antoninbas: As I commented in the proto source, I think this mode should probably include the cookie (all modes should proabbly include the cookie IMO)

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3795,34 +3803,56 @@ the `GetForwardingPipelineConfig RPC`. The request is defined as:
...
 Once a forwarding-pipeline config is installed on the device (either via
 `SetFowardingPipelineConfig` or a different mechanism), some P4Runtime servers
 may not support retrieval of the target-dependent config, in which case
-`config.p4_device_config` will be empty / unset in the response. However, all
-P4Runtime servers are required to return P4Info in this scenario.
+`config.p4_device_config` will be empty / unset in the response, even if
+`response_type` in the request was set to `ALL`. However, all P4Runtime servers
+are required to return the P4Info and cookie in this scenario. If the config is
+installed on the device with a mechanism other than
+`SetFowardingPipelineConfig`, the value of the returned cookie is left
```
antoninbas: what about: the value of the returned cookie will be the default Protobuf value (0).

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3795,34 +3803,56 @@ the `GetForwardingPipelineConfig RPC`. The request is defined as:
 
 ~ Begin Proto
 message GetForwardingPipelineConfigRequest {
+  enum ResponseType {
+    ALL = 0;
+    COOKIE_ONLY = 1;
+    P4INFO_ONLY = 2;
```
teverman: Should the controller have the option to get the p4_device_config only, without the p4info?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3795,34 +3803,56 @@ the `GetForwardingPipelineConfig RPC`. The request is defined as:
 
 ~ Begin Proto
 message GetForwardingPipelineConfigRequest {
+  enum ResponseType {
+    ALL = 0;
```
teverman: ALL will be the default value, so a server will see this when the client (say an older version prior to the response type addition) does not specify the type.  In this case, the client probably hasn't provided a cookie.  What should the server do when it has no cookie to return, default the cookie to zero?

Similarly, if a newer client provides a cookie to an older version of the server, the server won't store the cookie, and won't have a cookie to provide in the response.  The server will also not be aware of the response type presence, and thus it may respond with more information than the client expects.  Client and server need to handle these cases for upgrade compatibility.

There is another situation where client and server are both aware of the response type, but SetForwardingPipelineConfig was done before the update occurred, so once again, the server will have no cookie to provide if requested.  All of these situations are leading me to believe that the cookie ought to be wrapped in its own sub-message so that it is obvious whether the server has provided a known value in proto3.

docs/v1/P4Runtime-Spec.mdk
teverman: Need to account for older versions, as I described above.

docs/v1/P4Runtime-Spec.mdk
teverman: Cookie in sub-message to clearly identify cases where a cookie is not present.

docs/v1/P4Runtime-Spec.mdk
antoninbas: There isn't much harm in supporting this mode as well. However I don't think there is really a requirement for this case. @ccascone's concern - that lead to this proposed change - was the potential size of the `p4_device_config` on some targets.

docs/v1/P4Runtime-Spec.mdk
ccascone: Thanks for this comment. I agree on wrapping the cookie in a sub-message. This also helps to clarify the case where a pipeline-config was installed with a mechanism other than `SetForwardingPipelineConfig`, in which case the returned cookie will be unset. In general, I made the use of a cookie field optional.

docs/v1/P4Runtime-Spec.mdk
ccascone: I added a `DEVICE_CONFIG_AND_COOKIE`. I can envision cases in which the device config has a size smaller than the p4info, e.g. for a target that can support only a limited set of possible device config / P4 programs, the config field might be used to encode the ID of one of these configs already available in the device.

docs/v1/P4Runtime-Spec.mdk
ccascone: Changed modes to always return the cookie (is present). Older versions should be supported by `ForwardingPipelineConfig.Cookie` sub-message.

docs/v1/P4Runtime-Spec.mdk
ccascone: Cookie is now unset if one was not present in the `SetForwardingPipelineConfig` RPC, or if the pipeline has been installed via other mechanisms.

docs/v1/P4Runtime-Spec.mdk
antoninbas: Still unconvinced that this is a useful mode to have (even for small device configs). However it is not any kind of burden to implement so I don't have any strong objection.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1368,6 +1372,16 @@ The `p4_device_config` is opaque binary data which contains the target-specific
 configuration to realize the P4 program. The P4 program running on a target is
 changed by loading a new `FowardingPipelineConfig` on that target.
 
+The `cookie` field is opaque data which may be used by a control plane to
+uniquely identify a forwarding-pipeline configuration among others managed by
+the same control plane. For example, a controller can compute its value using a
+hash function over the P4Info and target-specific binary data. However, there
```
teverman: Nit: and -> and/or (client could choose to hash either or both)

antoninbas:
Fixes #65

PR #60, opened 2018-09-24
Add section on extending P4Runtime for non-PSA architectures in spec

antoninbas:
I didn't want to keep this section too long, it mostly gathers
information which is otherwise spread through the specification. This is
based on my experience extending P4Runtime to support Barefoot-specific
architectures.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -242,7 +242,7 @@ P4Runtime:
   Switch Architecture (PSA) externs (&eg; Counters, Meters, Action Profiles,
   ...).
 * Runtime control of architecture-specific (non-PSA) externs, through an
-  extensions mechanism.
+  extension mechanism.
 * Basic session management for Software-Defined Networking (SDN) use-cases,
   including support for controller replication to enable control-plane
   redundancy.
@@ -690,7 +690,7 @@ controller describes the role configuration, &ie; which operations, P4 entities,
 behaviors, etc. are in the scope of a given role. An unset `role.config` implies
 "full pipeline access" as well (similar to the default role explained above). In
 order to support different role definition schemes, `role.config` is defined as
-a `Any` Protobuf message [@ProtoAny]. Such schemes are out-of-scope of this
+an `Any` Protobuf message [@ProtoAny]. Such schemes are out-of-scope of this
 document. When partitioning of the control plane is desired, the P4Runtime
 client(s) and server need to agree on a role definition scheme in an
 out-of-band fashion.
@@ -864,7 +864,7 @@ message PkgInfo {
   string description = 2;
 }
 ~ End Proto
-## ID Allocation for P4Info Objects
+## ID Allocation for P4Info Objects { #sec-id-allocation}
 
 P4Info objects receive a unique ID, which is used to identify the object in
 P4Runtime messages. IDs are 32-bit unsigned integers which are assigned by the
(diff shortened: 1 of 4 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1342,10 +1346,45 @@ The `Digest` message defines the following fields:
...
+
+`Extern` messages are used to specify all extern instances across all extern
+types for a non-PSA architecture. This is useful when extending P4Runtime to
+support a new architetcure. Each architecture-specific extern type corresponds
+to at most one `Extern` message instance in P4Info. The `Extern` message defines
+the following fields:
+
+* `extern_type_id`, a 32-bit unsigned integer which uniquely identifies the
+  extern type in the context of the architecture. It must be in the [reserved
+  range](#sec-id-allocation) `[0x81, 0xfe]`. Note that this value does not need
```
jafingerhut: Is the entire 32-bit value restricted to the range [0x81, 0xfe]?  Or merely the most significant 8 bits or something like that?  Seems like very narrow restriction for a 32-bit value, unless it is 32-bits just for convenience and only values in range [0, 255] are expected to be supported.

jafingerhut: Except for this small probably-easily-resolved question, which is likely to be more about my ignorance of the wider context than something wrong with the changes, this LGTM.  I am probably the least steeped in the details of P4Runtime of the 3 people you asked to review this, though.

chrispsommers:
@jafingerhut  I only had those nits to report, that's why I "approved" with LGTM.  Thanks for clarifying.

antoninbas:
@chrispsommers I'm also confused. Did you leave any comments apart from approving?

docs/v1/P4Runtime-Spec.mdk
antoninbas: This is the type id so yes it is restricted to this range since it is used as the most significant 8 bits of the instance id for each extern instance of this type. It is 32 bits just for convenience (the value actually comes from an enum member).

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1027,6 +1027,10 @@ control plane. This message contains the following fields:
 * `is_const_table`, a boolean flag indicating that the table is filled with
   static entries and cannot be modified by the control plane at runtime.
 
+* `other_properties`, a `Any` Protobuf message [@ProtoAny] to embed
```
chrispsommers: nit: an `Any`

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1342,10 +1346,45 @@ The `Digest` message defines the following fields:
...
+* `extern_type_name`, which specifies the fully-qualified P4 name of the extern
+  type.
+
+* `instances`, a repeated field of `ExternInstance` Protobuf messages, with each
+  entry corresponding to a separate P4 instance of the extern. The
+  `ExternInstance` in turn defines the following fields:
+
+  * `preamble`, a `Preamble` message with the ID, name, and alias of this digest
+    instance.
+  * `info`, a `Any` Protobuf message [@ProtoAny] which is used to embed
```
chrispsommers: nit: an `Any`

chrispsommers:
@antoninbas @jafingerhut  My comments didn't get submitted initially, my bad.

antoninbas:
Fixed typos, please take another look

PR #61, opened 2018-09-24
Remove meter color-awareness information from P4Info

### LINKED ISSUE #58: Clarify the meaning of COLOR_AWARE meter in the spec
And in particular in relation with the PSA specification
When is the meter type set to `COLOR_AWARE`?

antoninbas:
There doesn't seem to be any point in keeping this in P4Info at the
moment. First the semantics are murky because in PSA each execute call
on a meter is either color-aware or color-blind (i.e. it is not an
attribute of the meter object itself). Second, at the moment this
information cannot be leveraged in any way by a P4 client since
P4Runtime does not expose any knob to enable / disable color-awareness
on a per meter entry basis at runtime. Even if there was such a knob, it
would be hard to reconciliate it with PSA semantics: in PSA the same
meter can be "executed" in 2 different actions, in color-aware mode in
one and in color-blind mode in the other; these 2 action can use the
same indirect meter entry.

Fixes #58

### DIFF
--- a/proto/p4/config/v1/p4info.proto
+++ b/proto/p4/config/v1/p4info.proto
@@ -249,12 +249,13 @@ message ActionProfile {
 }
 
 message CounterSpec {
-  // Corresponds to 'type' attribute for counter in P4 spec.
+  // Corresponds to 'type' constructor parameter for Counter / DirectCounter in
+  // PSA
   enum Unit {
     UNSPECIFIED = 0;
     BYTES = 1;
     PACKETS = 2;
-    BOTH = 3;  // not part of the P4 spec yet but will be in the future
+    BOTH = 3;
   }
   Unit unit = 1;
 }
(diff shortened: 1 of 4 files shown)


antoninbas:
@ccascone this fixes an inconsistency between P4Runtime & PSA that you pointed out a while back

proto/p4/config/v1/p4info.proto
```diff
@@ -274,17 +275,12 @@ message DirectCounter {
...
     BYTES = 1;
     PACKETS = 2;
   }
-  // Not part of the P4 spec yet but will be in the future.
-  enum Type {
-    COLOR_UNAWARE = 0;  // default value
-    COLOR_AWARE = 1;
-  }
   Unit unit = 1;
   Type type = 2;
```
teverman: This pull request deletes the 'Type' enum, but the type field is still present and defined as 'Type'.

antoninbas: Thanks for catching this Tom. It made me realize that the Travis CI build was not exiting properly in case of protoc failure so I was able to fix that. PTAL.

PR #63, opened 2018-09-24
Update the names used in CloneSessionId documentation to match proto …

jafingerhut:
…file

There were also a few old uses of `replica_id` replaced with
`instance` in a multicast group example.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2954,10 +2954,10 @@ entity {
   packet_replication_engine_entry {
     multicast_group_entry {
       muticast_group_id : 1
-      replicas { egress_port : 5 replica_id: 1 }
-      replicas { egress_port : 12 replica_id: 2 }
-      replicas { egress_port : 18 replica_id: 3 }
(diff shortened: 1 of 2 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3033,10 +3035,10 @@ type: INSERT
 entity {
   packet_replication_engine_entry {
     clone_session_entry {
-      clone_session_id : 100
-      replicas { egress_port : 0xFFFFFFFD replica_id: 1 } # to CPU
+      session_id : 100
+      replicas { egress_port : 0xFFFFFFFD instance: 1 } # to CPU
       class_of_service : 2
-      packet_length_bytes : 4094
+      packet_length_bytes : 4096
```
antoninbas: in theory truncating at 4094 is fine, but I respect your love for power of 2s :)

jafingerhut: I made that change to match the 4096 numbers mentioned text after the code example.  I suppose I could have updated the text to 4094 instead.  Just picked one.

antoninbas: sgtm

PR #66, opened 2018-09-27
Move to canonical representation for serializable enums

### LINKED ISSUE #64: Does choice of representing serializable enums violate the read-write symmetry principle?
Section 8.5.5 titled "enum, serializable enum and error" says:

"When providing serializable enum values through P4Data, one can either use the enum entry's name (enum human-readable string field) or its assigned value (enum_value bytestring field)."

That seems to imply one of two things to me:

(1) it breaks read-write symmetry, because if a client chooses to write with the string, but reads back an integer (or vice versa), the messages are not equal

or

(2) read-write symmetry is preserved, because when comparing the read to the write data, one must consider a numeric value and its corresponding string value for a serializable enum to be equal.

or maybe there is a 3rd possibility I'm not thinking of.

antoninbas:
In order to facilitate read-write symmetry. If we only have one valid
representation, it needs to be the bitvector one, as not all possible
numeric values have to be named.

Fixes #64

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1881,11 +1881,15 @@ therefore use a human-readable string in `P4Data` to represent `enum` and
 `error` values.
 
 Serializable `enum` types have an underlying fixed-width unsigned integer
-representation (`bit<W>`). Integer values must be assigned to each member entry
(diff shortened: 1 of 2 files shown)


antoninbas:
Brought up at the 10/24/2018 P4 API WG meeting, merging now

PR #69, opened 2018-10-15
Fix error-report figure

### LINKED ISSUE #68: Fix figure 7 in specification
I believe the top level message should be labelled `grpc::Status`, not `google::rpc::Status` (which is for the second level)

antoninbas:
It seems that the outer-most message wasn't correct. It is a
grpc::Status object (wrapping a google.rpc.Status message, itself
including a repeated field of p4.v1.Error messages).

Fixes #68


antoninbas:
Link to view generated spec: https://s3-us-west-2.amazonaws.com/p4runtime/travis/antonin/fix-error-report-figure/P4Runtime-Spec.html#sec-error-reporting-messages

PR #70, opened 2018-10-15
Clarify controller role definition

### LINKED ISSUE #56: Include more information about what a role is
This is based on question I get on p4.org.
In the "Master-Slave Arbitration and Controller Replication" section I suggest adding something like this:

> The definition of a role (out-of-scope) may also include the following:
x Ability to send `PacketOut` messages to the server
x Ability to receive `PacketIn` messages from the server, along with filtering / subscription mechanism based on the value of the `PacketMetadata` fields for each `PacketIn`

As well as

> Each master controller will only receive `DigestList` and `IdleTimeoutNotification` messages if the corresponding P4 object is included in its role.

Thoughts?

antoninbas:
With some information about what a role definition may include.

Fixes #56

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -713,7 +713,7 @@ scheme in an out-of-band fashion.
 
     1. If `device_id` does not match any of the devices known to the P4Runtime
        server, the server shall terminate the stream by returning a
-       `FAILED_PRECONDITION` error.
+       `NOT_FOUND` error.
 
     2. If the `election_id` is already used by another controller for the same
        (`device_id`, `role_id`), the P4Runtime server shall terminate the stream
@@ -759,6 +759,9 @@ scheme in an out-of-band fashion.
        if the client becomes master, in which case the server also follows the
        "mastership change rules" described in the [following
        section](#sec-master-change).
+
+If `role.config` does not match the "out-of-band" scheme previously agreed upon,
+the server must return an `INVALID_ARGUMENT` error.
 
 ## Mastership Change { #sec-master-change}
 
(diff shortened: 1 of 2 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -686,14 +686,25 @@ to 0.
...
+* Whether the controller is able to send `PacketOut` messages, along with a
+  filtering mechanism based on the values of the `PacketMetadata` fields to
+  select which `PacketOut` messages are allowed to be sent by the controller.
+
+An unset `role.config` implies "full pipeline access" (similar to the default
+role explained above). In order to support different role definition schemes,
+`role.config` is defined as an `Any` Protobuf message [@ProtoAny]. Such schemes
+are out-of-scope of this document. When partitioning of the control plane is
+desired, the P4Runtime client(s) and server need to agree on a role definition
+scheme in an out-of-band fashion.
```
teverman: Do you think it would be useful to add a rule(s) to the section below stating what the server does if it receives a role for which it does not recognize any agreed upon "out-of-band" scheme?

antoninbas: Yes I think it makes sense, I will add a sentence

antoninbas: @teverman I added the sentence. I also changed a few other parts of the spec. PTAL.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3468,7 +3471,18 @@ message WriteRequest {
...
+before processing the `updates` list:
+
+1. If `device_id` does not match any of the devices known to the P4Runtime
+   server or if `role_id` does not match any of the roles for the device, the
+   server must return a `NOT_FOUND` error.
+
+2. If the client is not the master for (`device_id`, `role_id`) according to the
+   `election_id` value, the server must return a `PERMISSION_DENIED` error.
+
+3. If the `Write` is attempted begore a `ForwardingPipelineConfig` has been set,
```
chrispsommers: s/begore/before/

PR #84, opened 2018-11-08
Support architecture-specific notifications on stream

### LINKED ISSUE #57: Add ability to extend StreamMessageRequest and StreamMessageResponse for architecture-specific messages

antoninbas:
Extend StreamMessageRequest and StreamMessageResponse with an Any
Protobuf field to embed architecture-specific messages.

Fixes #57


jafingerhut:
Thanks for invite to review.  The intent seems pretty clear, to enable new spontaneously-generated-notification type messages to be generated from a server to the client for future externs that might need it (and I guess also from client to server, too?  If so I don't see how externs would need that, but I have no problem if it the capability is added).

Even if I got that intent of the changes correct, I'm still not enough of a gRPC lawyer to know whether you got the 'how' part in message formats correct, so I am hoping another requested reviewer can double-check that.

antoninbas:
@jafingerhut while it's unlikely that this mechanism will be used to spontaneously send messages from the client to the server, it may be used for acknowledging notifications generated by the switch (see digest for example)

jafingerhut:
@antoninbas Thanks for response.  Makes sense.

PR #85, opened 2018-11-08
Better support for @defaultonly and @tableonly annotations

### LINKED ISSUE #83: Improve support for @defaultonly and @tableonly annotations in P4Info
1) Improve "@defaultonly" and "@tableonly" support in P4Info by making them Protobuf fields:

```
message ActionRef {
  uint32 id = 1;
  enum Scope {
    DEFAULT = 0,
    DEFAULT_ONLY = 1,
    TABLE_ONLY = 2
  }
  repeated string annotations = 3;
}
```

2) Remove `const_default_action_has_mutable_params` field which is meaningless for P4_16 programs

antoninbas:
These annotations are now "first-class citizens" in P4Info: an enum
field is used to indicate the scope of the action, i.e. how the action
can be used.

This commit also removes const_default_action_has_mutable_params, which
was introduced for P4_14 initially and has fallen out of use.

Fixes #83


antoninbas:
@jafingerhut this is what we discussed at the API WG meeting, please take a look

proto/p4/config/v1/p4info.proto
```diff
@@ -182,11 +182,6 @@ message Table {
   repeated ActionRef action_refs = 3;
   // 0 (default value) means that the table does not have a const default action
   uint32 const_default_action_id = 4;
-  // a table may have a const default action, whose action parameter values can
-  // be changed at runtime. However, in most cases the parameters of the default
-  // action are also bound at compile-time and cannot be changed by the runtime,
-  // which is what this boolean flag indicates.
-  bool const_default_action_has_mutable_params = 5;
```
jafingerhut: To clarify for me, do you also believe it is the case that in P4_16 a default action of a table falls into one of these two categories?

(a) declared 'const' in the source code, and neither the action name nor the action parameters may be changed by a client
(b) not declared 'const' in the source code, and the action name can be changed by the client to any action name _except_ one annotated @tableonly, and the action parameters can be assigned any values that are correct for the action name.

antoninbas: yes this is my understanding of the P4_16 spec as well

jafingerhut: Sounds good.

PR #88, opened 2018-11-12
Add link to PSA specification

antoninbas:
Link will be updated after PSA 1.1 release.


antoninbas:
@jafingerhut I will update the link to point to the "frozen" PSA 1.1 spec when it is released

jafingerhut:
Sounds good.  We have not received any comments on the "release candidate" of PSA v1.1 yet, so are still on track to release it 2018-Nov-21.

PR #89, opened 2018-11-12
Add more details for action selector programming

### LINKED ISSUE #43: Add section about psa_empty_group_action
when describing action selectors or tables with action selector implementations
in particular add this restriction for v1.0: `psa_empty_group_action` has to be const to avoid making the `TableEntry` message too complex

antoninbas:
Based on recent additions to the PSA spec, such as the
psa_empty_group_action table property.

Fixes #43

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2581,20 +2581,20 @@ following semantics.
 * `INSERT`: Add a new member entry bound to an eligible P4 action
   specification. The member id must be different from ids of already programmed
   entries for that extern, or the server must return an `ALREADY_EXISTS` error
-  code. The action specification must be provided or the server must return
+  code. The action specification must be provided, or the server must return
   `INVALID_ARGUMENT`. The total number of members should not exceed the maximum
-  specified in the P4 extern specification as a result of this insertion or the
+  specified in the P4 extern specification as a result of this insertion, or the
   server should return `RESOURCE_EXHAUSTED`.
 * `MODIFY`: Modify the action specification of an existing member entry. An
-  entry with the member id must exist (or the server must return `NOT_FOUND`)
-  and the action specification must be provided (or the server must return
-  `INVALID_ARGUMENT`).
+  entry with the member id must exist, or the server must return `NOT_FOUND`,
+  and the action specification must be provided, or the server must return
+  `INVALID_ARGUMENT`.
 * `DELETE`: Delete the member entry and deallocate the member id. If the member
   id is not valid the server must return a `NOT_FOUND` error code. The member
-  must not be part of an action profile group or the server must return
+  must not be part of an action profile group, or the server must return
   `FAILED_PRECONDITION`. If needed, the action profile group should first be
   modified to remove the member from the group. The member must not be
-  referenced in the table action of any table entry or the server must also
+  referenced in the table action of any table entry, or the server must also
   return `FAILED_PRECONDITION`.
 
 ### Action Profile Group Programming


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2574,24 +2575,27 @@ An `ActionProfileMember` entity update message has the following fields:
...
-  provided. If the member is part of an action profile group, then the bound
-  action must be of the same action type as actions bound to other members of
-  the group. See the condition on bound actions for members belonging to a
-  group.
-* `DELETE`: Delete the member entry and deallocate the member id. The member
-  should not be part of an action profile group. If needed, the action profile
-  group should first be modified to remove the member from the group. The member
-  should not be referenced in the table action of any table entry.
+  entry with the member id must exist (or the server must return `NOT_FOUND`)
+  and the action specification must be provided (or the server must return
```
teverman: Style nit - for MODIFY, the server's error responses are in parens, whereas they flow with the text in INSERT & DELETE.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2574,24 +2575,27 @@ An `ActionProfileMember` entity update message has the following fields:
...
 * `INSERT`: Add a new member entry bound to an eligible P4 action
   specification. The member id must be different from ids of already programmed
-  entries for that extern. The action specification must be provided. The total
-  number of members should not exceed the maximum specified in the P4 extern
-  specification, as a result of this insertion.
+  entries for that extern, or the server must return an `ALREADY_EXISTS` error
+  code. The action specification must be provided or the server must return
+  `INVALID_ARGUMENT`. The total number of members should not exceed the maximum
+  specified in the P4 extern specification as a result of this insertion or the
+  server should return `RESOURCE_EXHAUSTED`.
```
teverman: Style nit - there's a comma before "or the server must return" in one of the 3 INSERT error cases.   Do you want to make them consistent?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2630,52 +2634,53 @@ following semantics.
...
 - `MODIFY`: Modify the member set specification of an existing group entry. An
-  entry with the `group_id` must exist. All members specified in the group entry
-  must exist in the selector. The value of `max_size` must be identical to the
-  value used when inserting the group, otherwise an `INVALID_ARGUMENT` error is
-  returned.
-- `DELETE`: Delete the group entry and deallocate the `group_id`. The group
-  should not be referenced in the table action of any table entry.
+  entry with the `group_id` must exist, or the server must return
+  `NOT_FOUND`. All members specified in the group entry must exist in the
+  selector, or the server must return `NOT_FOUND`. The value of `max_size` must
```
teverman: A client receiving a NOT_FOUND response has no way to know whether it's because the group_id does not exist or a member does not exist. Would it make sense to differentiate them?

antoninbas: NOT_FOUND seems like the most appropriate error code in both of these situations, however we could switch to FAILED_PRECONDITION for the second case (one of the members do not exist). Is there value in being able to separate these cases in a programmatic fashion in the client, or should we just rely on the error message?

docs/v1/P4Runtime-Spec.mdk
antoninbas: Will fix (probably add a comma everywhere)

docs/v1/P4Runtime-Spec.mdk
antoninbas: Will fix (remove the parens)

docs/v1/P4Runtime-Spec.mdk
teverman: I haven't checked, but I would guess there are other request messages where the same error response indicates one of multiple possible error conditions.  OK to leave as is.

antoninbas:
Thanks for the reviews!
@jafingerhut do you want to take a look at this before I merge?

jafingerhut:
Thanks for the reminder.  As discussed in another medium, please go ahead without my reading this one yet.  Will work on making the time to do a once-over on the entire P4Runtime spec before release, and create issues for any questions or concerns I have.

PR #90, opened 2018-11-15
Add section giving atomicity requirements for individual operations

### LINKED ISSUE #54: Should the P4Runtime spec include a generic section on atomicity of individual operations?
Similar to the "Atomicity of control plane API operations" section in the PSA spec (https://github.com/p4lang/p4-spec/blob/master/p4-16/psa/PSA.mdk#L2437)?
We can also just add a link to the PSA spec section somewhere...

antoninbas:
Mostly a pointer to the PSA specification, with some concrete examples
of the guarantees provided by the @atomic annotation with respect to
control-plane operations. Future versions of the PSA specification may
benefit from including more details about how the @atomic annotation
interacts with control-plane operations.

Fixes #54

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -3575,9 +3575,13 @@ control C1 {
 ~ End P4Example
 
 If a P4Runtime client issues a *wildcard* `Read` on Register `r1`, there is no
-guarantee that `r1[1] == r1[2]` in the response, as the read for `r1[1]` and the
-read for `r1[2]` may not happen "within" the same atomic block. The server is
-also not required to read `r1[1]` and `r1[2]` sequentially and may report a more
+guarantee that `r1[1] == r1[2]` in the response, as the read for `r1[1]` may
+occur after the data plane executes the first atomic block, but before the
+second atomic block, and the read for `r1[2]` may occur after the data plane
+executes the second atomic block. In other words, the server is explicitly
+allowed to read `r1[1]` and `r1[2]` separately, while allowing the data plane to
+perform operations on the register between those two reads.The server is also
+not required to read `r1[1]` and `r1[2]` sequentially and may report a more
 recent value for `r1[2]` than for `r1[1]`. The atomicity guarantees for a
 wildcard read are the same as for the equivalent batch (as one `ReadRequest`
 message) of individual read requests.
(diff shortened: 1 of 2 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3509,6 +3509,83 @@ gRPC provides utility functions `ExtractErrorDetails()` and `SetErrorDetails()`
...
+        r1.write((Index_t)1, (Value_t)200);
+        r1.write((Index_t)2, (Value_t)200);
+    }
+  }
+}
+~ End P4Example
+
+If a P4Runtime client issues a *wildcard* `Read` on Register `r1`, there is no
+guarantee that `r1[1] == r1[2]` in the response, as the read for `r1[1]` and the
+read for `r1[2]` may not happen "within" the same atomic block. The server is
```
jafingerhut: suggested change, if it matches what you intended to mean:

Replace this: 'may not happen "within" the same atomic block.'

with this: 'may happen after the data plane executes the first atomic block, but before it executes the second atomic block.'

jafingerhut: or maybe even more precisely (I think I messed up with my previous suggestion):

replace the phrase starting with 'as the read ...' with this instead: 'as the read for `r1[1]` may occur after the data plane executes the first atomic block, but before the second atomic block, and the read for `r1[2] may occur after the data plane executes the second atomic block.'

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3509,6 +3509,83 @@ gRPC provides utility functions `ExtractErrorDetails()` and `SetErrorDetails()`
...
+also not required to read `r1[1]` and `r1[2]` sequentially and may report a more
```
jafingerhut: It isn't completely clear to me what 'sequentially' means here, or maybe there is a clearer way to say this.  Maybe instead of this sentence, the following instead?

'The server is explicitly allowed to read `r1[1]` and `r1[2]` while allowing the data plane to perform operations on the register between those two reads.'

docs/v1/P4Runtime-Spec.mdk
antoninbas: I made that change. Thanks!

docs/v1/P4Runtime-Spec.mdk
antoninbas: I added the suggested sentence because I think it's a good clarification of the previous point.

However, what I mean here specifically is that the server can issue parallel reads for `r1[1]` and `r1[2]` (thus the "not sequentially") and that based on the ordering of the actual HW read, one may get a more "recent" version for `r1[2]`. I think this is consistent with the guarantees we have for batch reading. Would you agree with that statement and if yes, do you have suggestions on how to rephrase it if it is not clear?

jafingerhut:
Looking good.  I think I understand all of the changes now, and hopefully others will, too :)

PR #91, opened 2018-11-21
Add paragraph about table size

antoninbas:
And the guarantees provided by P4_16 / P4Runtime. This is mostly a
pointer to the relevant section in the P4_16 spec.


antoninbas:
Thanks!

PR #94, opened 2018-11-29
More details on valid values for multicast and clone session ids

### LINKED ISSUE #44: Can a multicast group id / a clone session id / a replica id be 0?
Usually in P4Runtime, 0 is an invalid id. Are things different for multicast group id / clone session id / replica id? Unlike an action profile member id for example (but much like a port id?), these ids map directly to P4 metadata fields. We need to make a decision and include it in the specification.

I think these ids are different from from other P4Runtime ids (e.g. action profile member id) which never leave the realm of the control-plane. They are also different from port ids IMO, since I would expect most targets to support a linear sequence of multicast group ids (for example) in the range `[0, 2^W[` and we may not want to worry about mapping / translation like we did for ports.

antoninbas:
Based on the PSA spec.

Fixes #44

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -3117,13 +3117,18 @@ A multicast group may be inserted, modified or deleted as per the following
 semantics.
 
 * `INSERT`: Add a new multicast group entry bound to a set of egress ports and
-  replica IDs. The `multicast_group_id` is `uint32` type, must be unique across
-  all multicast group entries, and its value may not exceed the maximum allowed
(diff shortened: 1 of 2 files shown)


antoninbas:
@jafingerhut thanks for the review.
There is some information about where the mapping comes from in the spec already:

> The first argument to the @p4runtime_translation annotation is a URI that indicates to the P4Runtime server which numerical mapping - provided by the out-of-band switch configuration mechanism - to use to translate between the SDN value and the dataplane value. The second argument is the bitwidth of the SDN representation of the translated entity (32-bit in the case of ports).

Issue #102, opened 2018-12-20
Describe how to populate P4Info via @brief() and @description() annotations


chrispsommers:
See https://github.com/p4lang/p4c/issues/1170.

PR #103, opened 2018-12-20
Described how to document with @brief, @description, @pkginfo.

chrispsommers:
See https://github.com/p4lang/p4runtime/issues/102 and https://github.com/p4lang/p4c/issues/1170
I think this needs to be reality-checked against possible implementation in p4c.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -844,18 +844,40 @@ message Preamble {
 }
 ~ End Proto
 
+### Annotating P4 Entities with `Documentation`
+P4 entities may be annotated using the following annotations:
+
+~ Begin p4example
+@brief(string...)
+@description(string...)
+~ End p4example
+
+Attaching either or both of these annotations to an entity will generate a
+P4Info [Documentation Message](#sec-documentation-message), which in turn will
+appear in the [Preamble Message](#sec-preamble-message) for the entity.
+
+The P4 compiler should not emit `annotation` messages in the P4Info for these specific
+cases; instead, it should generate the `Documentation` messages as described.
+
+The following example shows documentation annotations for a `table` entity:
+
+~ Begin p4example
+@brief("Match IPv4 addresses to next-hop MAC and port")
+@description("Match IPv4 addresses to next-hop MAC and port. \
+Uses LPM match type.")
+table my_ipv4_lkup {
+  ...
+}
+~ End p4example
+
 ## `PkgInfo` Message
 
 The `PkgInfo` message contains package-level metadata which describes the
 overall P4 program itself, as opposed to P4 entities. `PkgInfo` can be extracted
-and used to facilitate "browsing" of available P4 programs from a
-library. Although all fields are technically "optional," every implementation
-should include as a minimum the name, version, doc and arch fields. The other
-fields are recommended to be included.
-
-Note, the known P4 compilers as of this writing don't emit `PkgInfo` as part of
-the P4Info output. Until compiler support is added, a utility to post-process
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -878,6 +900,48 @@ message PkgInfo {
...
+~ End p4example
+
+Multiple `@pkginfo` annotations are allowed. For compactness, multiple key-value
+pairs can appear in a single `@pkginfo` annotation, separated by commas. Each
+key should only appear once, but if it appears multiple times, then the last
+value shall be used. Declaring one or more of these annotations will generate a
+single corresponding `PkgInfo` message in the P4Info as described in
+[PkgInfoMessage](#sec-pkginfo-message). The `key`s must be from among the
+message elements inside `PkgInfo`, for example, `name`, `version`, etc. One
+exception is that the `Documentation` element of `PkgInfo` must be expressed as
```
antoninbas: shouldn't we use `@brief` and `@description` for documentation annotations? It may be more consistent.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -878,6 +900,48 @@ message PkgInfo {
   string url = 8;
 }
```
antoninbas: do you want to add a `repeated string` field for extra annotations on the package? I would be fine with that.

antoninbas:
One more thing: would you mind moving this PR from your fork to a branch in this repo (you have write access). This way, I can visualize the rendered HTML & PDF online before merging.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -878,6 +900,48 @@ message PkgInfo {
...
+
+The following example shows `@pkginfo` annotations using a mixture of single and
+multiple key-value pairs:
+
+~ Begin p4example
+@pkginfo(name="switch.p4",version="2")
+@pkginfo(brief="L2/L3 switch")
+@pkginfo(description="L2/L3 switch.\
+Built for data-center profile.")
+@pkginfo(arch="v1model")
```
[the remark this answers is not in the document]
chrispsommers: Good point. Do you think the annotation should still be allowed, and if so, which takes precedence if the compiler populates it too?

docs/v1/P4Runtime-Spec.mdk
chrispsommers: In other words, if you add those annotations to main() they populate the PkgInfo? Sounds good. In fact I was vacillating over this when I wrote it.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: I like the idea.

docs/v1/P4Runtime-Spec.mdk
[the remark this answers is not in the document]
antoninbas: I think we should not allow the annotation for this. If the programmer's intent is to include some different piece of information, a custom annotation can be used (and the "extra" repeated field I mentioned below can be used).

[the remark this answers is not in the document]
chrispsommers: @antoninbas Sounds good to me, thx.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: @antoninbas Actually field 4 is already there for annotations. I'd forgotten I put it there 10 months ago.

chrispsommers:
Closing this PR since it's from a fork and opened a new PR https://github.com/p4lang/p4runtime/pull/126 after incorporating all feedback.

PR #127, opened 2018-12-23
Fix several typos, and rephrase one whole sentence

jafingerhut:


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1730,7 +1731,7 @@ For example, the following P4~16~ objects involve complex types that need to be
 exposed in P4Runtime in order to support runtime operations on these objects.
 
 ~ Begin P4Example
-value_set<tuple<bit<16>, bit<8> > >(16) pvs_complex;
+value_set<tuple<bit<4>, bit<8> > >(16) pvs_complex;
```
chrispsommers: Why the change to 4 bits?

jafingerhut: This is really just a nit-pick change -- the field being compared against in the example is the IHL field of the IPv4 header, which is 4 bits wide in the standard.  This code snippet is short enough that it does not contain the explicit declaration of the IPv4 header type.  My OCD kicked in while reading it :-)

chrispsommers: @jafingerhut Got it, I was wondering.  And in this case,  OCD stands for "Occupational Capability Demonstration." :)

jafingerhut: Oh, and I was misremembering that it was the IPv4 IHL field, but it was another 4-bit field, the IPv4 version field.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4042,9 +4043,9 @@ the P4Info, the opaque target-dependent forwarding-pipeline configuration data
 to uniquely identify such configuration. See the [Forwarding-Pipeline
 Configuration](#sec-p4-fwd-pipe-config) section for details.
 
-A P4Runtime server running on an a typical device may not support
+A P4Runtime server running on an a non-programmable device may not support
```
antoninbas: there is a typo that is not of your making on this line, there is an extra article ("an a")

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4322,7 +4323,7 @@ caption: "P4Runtime Metadata Translation for the Portable Switch Architecture" }
 [psa-metadata-translation]: build/psa-metadata-translation.[svg,png] \
 { height: 7cm; page-align: here }
 
-Figure [#fig-psa-metadata-translation] above illustrates a motivating example,
+Figure [#fig-psa-metadata-translation] illustrates a motivating example,
```
antoninbas: lol, nice catch

docs/v1/P4Runtime-Spec.mdk

PR #144, opened 2019-01-02
Fix definition of Action message in Table Entry section

### LINKED ISSUE #140: Section 9.1.2 "Action Specification"
This sentence seems wrong: "The `Action` Protobuf message, which is used for direct tables only, has the following fields:"

because the `Action` Protobuf message is also used for tables within action profiles and action selectors, too?

Or perhaps the statement is meant as "which is used as the `oneof` choice, among those available in a `TableAction` message, only for direct tables" ?

antoninbas:
Fixes #140


jafingerhut:
Action is also used in ActionProfileMember, too, yes?  So the use of Action doesn't seem to be unique to the one-shot for indirect tables.

antoninbas:
Hmm that's correct. Then I really don't know where the sentence came from in the first place. I'll use your siuggestion from #140.

PR #146, opened 2019-01-03
Fix #133 - change description of scenarios where P4 programs are not …

### LINKED ISSUE #133: Please do not mislead P4Runtime users about the sufficiency of P4Info file for control plane developers
The following text in the current P4Runtime spec seems to me _fundamentally_ misleading:

In section: "P4 as a Behavioral Description Language"

"Furthermore, it is not necessary to have a P4 source program to begin with, since the controller does not use it. From the standpoint of controller (not pipeline) implementers, the P4 source code is just helpful documentation. Some parties may wish to keep their P4 source code private. The minimum requirement is a P4Info file which can be loaded by a controller in order to render the correct P4Runtime API. As long as the target supports the operations implied by the P4Info file, the underlying implementation is moot."

In my opinion (based on years of experience of writing control plane software sometimes correctly, often incorrectly), the P4 source code can be _crucial_ to the development of correct controller code for a complete system.

There are an _unlimited_ number of P4 programs that will all produce the _same_ P4Info file when they are compiled.

Here are just some of the ways that two P4 programs with the same P4Info file could differ:

+ The _order_ that tables are applied.  This order is _not_ represented anywhere in the P4Info file, only which tables exist.
+ The assignment statements performed in actions.  The P4Info file represents the names of actions and their parameters, but says nothing about the contents of those actions.
+ The contents of 'if' conditions, or indeed, whether there are 'if' conditions that make the application of tables or other controls conditional, or unconditional.
+ Assignment statements that occur between applying tables or calling externs.

Developers attempting to write control plane software for a P4 data plane _rely_ on knowledge of how the data plane works, in order to know what table entries to install.  Just imagine two P4 programs where program P1 applies table T1 only when condition (A && B) is true, and another program P2 applies table T1 only when condition (A && !B) is true.  Or where program P1 has an action set_nexthop that assigns the value of parameter nexthop_id to local_metadata.nexthop_id, but program P2's set_nexthop action _does not do this_.

In summary, having access to a _complete_ description of a data plane's behavior in the form of an accurate P4 program (whether the data plane is programmable or not) can make the difference between control software writers relying on inaccurate natural language documentation (e.g. in English) and writing completely incorrect control plane software, versus writing correct control plane software.  The contents of a P4Info file alone _is nowhere near enough to achieve this_.

jafingerhut:
…available

Emphasize how useful either a P4 program, or at least detailed
documentation of the data plane behavior, are to writing correct
control plane software.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -439,24 +439,32 @@ state.
 
 ## P4 as a Behavioral Description Language
 
-P4 can be considered a behavioral description of a switching device which may or
-may not execute "P4" natively. There is no requirement that a P4 compiler be
-used in the production of either the P4 device config or the P4Info. There is no
-absolute requirement that the target accept a `SetForwardingPipelineRequest` to
-change its pipeline "program", as some devices may be fixed in function.
-Furthermore, it is not necessary to have a P4 source program to begin with,
-since the controller does not use it. From the standpoint of controller (not
-pipeline) implementers, the P4 source code is just helpful documentation. Some
-parties may wish to keep their P4 source code private. The minimum requirement
-is a P4Info file which can be loaded by a controller in order to render the
-correct P4Runtime API. As long as the target supports the operations implied by
-the P4Info file, the underlying implementation is moot.
+P4 can be considered a behavioral description of a switching device
+which may or may not execute "P4" natively. There is no requirement
+that a P4 compiler be used in the production of either the P4 device
+config or the P4Info. There is no absolute requirement that the target
+accept a `SetForwardingPipelineRequest` to change its pipeline
+"program", as some devices may be fixed in function, or configured via
+means other than P4 programs. Furthermore, a controller can run
+without a P4 source program, since the P4Info file provides all of the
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -483,7 +491,7 @@ number of reasons, the most likely of which are:
    protect intellectual property or maintain security.
 
 2. The target was not implemented using P4 code to begin with, although it still
-   obeys the "contract" specified in the P4Info.
+   obeys the control plane API specified in the P4Info.
```
antoninbas: Do you want to emphasize here that in the absence of a P4 program, the datapath knowledge required to write the controller has to come from somewhere else: documentation, RFCs, ...

antoninbas:
@chrispsommers would you mind taking a look at this since you wrote the original text IIRC?

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Sure, I  will add a sentence on that with another commit, and see about changing the text wrap to 80 characters like the rest of the document.  72 is the default I have set in my Emacs config (or perhaps that comes with it by default).

PR #147, opened 2019-01-03
Fix #121 - Add a few notes on how a server identifies which client

### LINKED ISSUE #121: Can clients change their election_id across different WriteRequest messages?
12. Write RPC

Should a server check the election_id value in every WriteRequest to
see if it matches what the same client has sent before?

Should a client ever be allowed to change its election_id in a single
session?

jafingerhut:
... sent a message.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -697,7 +697,8 @@ on clients identifying themselves in every write request, by including the
 values `device_id`, `role_id`, and `election_id` in all write requests. The
 server trusts clients not to use a triple of values other than their own in
 their write requests. gRPC provides authentication methods [@gRPCAuth] that can
-be deployed to prevent untrusted clients from making changes.
+be deployed to prevent untrusted clients from creating channels, and thus from
+making changes or even reading the state of the server.
 
 ## Default Role
 


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -679,6 +679,22 @@ described as follows:
       controller is determined to be a slave for a given (`device_id`,
       `role_id`).
 
+The gRPC libraries enable a server process to cause a function to be called when
```
antoninbas: This assumes that the gRPC P4Runtime service is used in synchronous mode. Maybe this should be rephrase in a more generic fashion (i.e. "gRPC lets the server identify which client originated each stream packet etc.") or we could indicate that we are talking about synchronous mode while stating that the same principle applies to asynchronous as well.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -679,6 +679,22 @@ described as follows:
...
+unexpectedly terminated). Thus the server can easily associate all
+`StreamChannel` messages received from the same client, because they are
+processed within the context of the same function call.
+
+A P4Runtime implementation need not rely on the gRPC library providing
+information with unary RPC messages that identify which client they came from.
+Unary RPC messages include requests to write table entries in the data plane, or
+read state from the data plane, among others described later. P4Runtime relies
+on clients identifying themselves in every write request, by including its
+`election_id` value in all write requests. The server trusts clients not to use
```
antoninbas: + device_id & role_id (see https://github.com/p4lang/p4runtime/issues/149)

jafingerhut: Yes, this one I realized as you were reviewing this PR, and I committed a change that I believe addresses this comment.

jafingerhut: but do please check the updated version for accuracy.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: I have attempted to address this comment, by starting with a sentence based on your suggestion, and keeping the rest of the paragraph as an example of how, in synchronous mode, gRPC enables this capability.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -679,6 +679,25 @@ described as follows:
       controller is determined to be a slave for a given (`device_id`,
       `role_id`).
 
+gRPC enables the server to identify which client originated each message in the
+`StreamChannel` stream. For example, in synchronous mode, the gRPC libraries
```
antoninbas: can we replace "gRPC libraries" with "C++ gRPC library" here to be more accurate (streams work differently in gRPC Python for example)., and maybe add a reference to `[@gRPCStreamC]`

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -679,6 +679,25 @@ described as follows:
...
+enable a server process to cause a function to be called when a new client
+creates a `StreamChannel` stream. This function should not return until the
+stream is closed and the server has done any cleanup required when a
+`StreamChannel` is closed normally (or broken, e.g. because a client process
```
antoninbas: s/`e.g.`/`&eg;`

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -679,6 +679,25 @@ described as follows:
...
+processed within the context of the same function call.
+
+A P4Runtime implementation need not rely on the gRPC library providing
+information with unary RPC messages that identify which client they came from.
+Unary RPC messages include requests to write table entries in the data plane, or
+read state from the data plane, among others described later. P4Runtime relies
+on clients identifying themselves in every write request, by including the
+values `device_id`, `role_id`, and `election_id` in all write requests. The
+server trusts clients not to use a triple of values other than their own in
+their write requests.
```
antoninbas: Should we add here that clients are trusted because gRPC provide authentication mechanisms that can be deployed in a production implementation? We can add a reference to https://grpc.io/docs/guides/auth.html for completeness.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Done, with reference added.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Done

docs/v1/P4Runtime-Spec.mdk
jafingerhut: I added a sentence about this, and a new reference in the .bib file to the link you suggested.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -679,6 +679,26 @@ described as follows:
...
+
+A P4Runtime implementation need not rely on the gRPC library providing
+information with unary RPC messages that identify which client they came from.
+Unary RPC messages include requests to write table entries in the data plane, or
+read state from the data plane, among others described later. P4Runtime relies
+on clients identifying themselves in every write request, by including the
+values `device_id`, `role_id`, and `election_id` in all write requests. The
+server trusts clients not to use a triple of values other than their own in
+their write requests. gRPC provides authentication methods [@gRPCAuth] that can
+be deployed to prevent untrusted clients from making changes.
```
antoninbas: from making establishing a channel actually, not just making changes

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -679,6 +679,26 @@ described as follows:
...
+`StreamChannel` messages received from the same client, because they are
+processed within the context of the same function call.
+
+A P4Runtime implementation need not rely on the gRPC library providing
+information with unary RPC messages that identify which client they came from.
+Unary RPC messages include requests to write table entries in the data plane, or
+read state from the data plane, among others described later. P4Runtime relies
+on clients identifying themselves in every write request, by including the
+values `device_id`, `role_id`, and `election_id` in all write requests. The
+server trusts clients not to use a triple of values other than their own in
```
antoninbas: is "triple of values" equivalent to "triplet of values" in usage?

docs/v1/P4Runtime-Spec.mdk
jafingerhut: clarified this point in the text in the next commit coming in a few minutes

docs/v1/P4Runtime-Spec.mdk
jafingerhut: I checked my dictionary, and "triple" can be used as a noun, and "triplet" is often used for 3 of the same or similar things, which in this case the 3 values are expected to be different in general, so "triple" might be slightly better.

PR #150, opened 2019-01-04
Fix #149 - clarify that election_id must be unique per (device_id, ro…

### LINKED ISSUE #149: Is election_id allowed to be the same for different controllers?
A more full statement of the question:

Suppose a single server has multiple devices (say 3 for example, numbered with device_id values 101, 102, and 103), and each of its devices has 2 roles (numbered with role_id values 201 and 202).

Suppose we have 2 controllers, C1 and C2.

Is it allowed that controller C1 can use election_id 42 for (device_id=101, role_id=201), and at the same time controller C2 can use election_id 42 for (device_id=101, role_id=202) ?

In general, can the same election_id be used by different controllers for different (device_id, role_id) value pairs?

Right now the spec says:

"Each controller is assigned a `role_id` and an `election_id`. The `role_id`
  defines the role (or realm) that the controller is part of. The `election_id`
  is unique per role and identifies the master for a specific role. At any point
  in time, the controller with the largest `election_id` for each role is the
  master and the rest are slaves."

It is not clear whether election_id's are required to be unique for each role_id value in a single server, or whether they are merely required to be unique for each (device_id, role_id) pair in a single server.

jafingerhut:
…le_id)

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -618,17 +618,18 @@ To support multiple controllers, P4Runtime uses the same streaming channel
 (available via `StreamChannel` RPC) for session management. The workflow is
 described as follows:
 
-* Each controller is assigned a `role_id` and an `election_id`. The `role_id`
-  defines the role (or realm) that the controller is part of. The `election_id`
-  is unique per role and identifies the master for a specific role. At any point
-  in time, the controller with the largest `election_id` for each role is the
-  master and the rest are slaves. Note that the device does not assign a
+* For each device, every controller is assigned a `role_id` and an
+  `election_id`. The `role_id` defines the role (or realm) that the controller
+  is part of. The `election_id` is unique per `(device_id, role_id)` pair, and
+  identifies the master for a specific device and role. At any point in time,
+  the controller with the largest `election_id` for each device/role is the
+  master and the rest are slaves. Note that the server does not assign a
   `role_id` and `election_id` to any controller. It is up to an arbitration
-  mechanism outside of the device to decide on the controller roles and the
+  mechanism outside of the server to decide on the controller roles and the
   master and slave controllers for each role. The P4Runtime server running on
   the device only keeps track of the `role_id` and `election_id` of the
-  controllers to determine which connected controller is master at any point of
-  time.
+  controllers (independently for each device) to determine which connected
+  controller is master at any point of time.
 
 * To start a controller session, a controller first opens a bidirectional stream
   channel to the server via the `StreamChannel` RPC for each device. This is the


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -618,17 +618,18 @@ To support multiple controllers, P4Runtime uses the same streaming channel
...
-  master and the rest are slaves. Note that the device does not assign a
+* For each device, every controller is assigned a `role_id` and an
+  `election_id`. The `role_id` defines the role (or realm) that the controller
+  is part of. The `election_id` is unique per `(device_id, role_id)` pair, and
+  identifies the master for a specific device and role. At any point in time,
+  the controller with the largest `election_id` for each device/role is the
+  master and the rest are slaves. Note that the server does not assign a
   `role_id` and `election_id` to any controller. It is up to an arbitration
-  mechanism outside of the device to decide on the controller roles and the
+  mechanism outside of the server to decide on the controller roles and the
```
antoninbas: +1 for that change

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -618,17 +618,18 @@ To support multiple controllers, P4Runtime uses the same streaming channel
 (available via `StreamChannel` RPC) for session management. The workflow is
 described as follows:
 
-* Each controller is assigned a `role_id` and an `election_id`. The `role_id`
-  defines the role (or realm) that the controller is part of. The `election_id`
-  is unique per role and identifies the master for a specific role. At any point
-  in time, the controller with the largest `election_id` for each role is the
-  master and the rest are slaves. Note that the device does not assign a
+* For each device, every controller is assigned a `role_id` and an
```
antoninbas: I'm not sure this adds clarity. I'd rather stick to one controller => one device id & role id (see my comment there: https://github.com/p4lang/p4runtime/issues/149#issuecomment-451687969)

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -618,17 +618,18 @@ To support multiple controllers, P4Runtime uses the same streaming channel
...
+  the controller with the largest `election_id` for each device/role is the
+  master and the rest are slaves. Note that the server does not assign a
   `role_id` and `election_id` to any controller. It is up to an arbitration
-  mechanism outside of the device to decide on the controller roles and the
+  mechanism outside of the server to decide on the controller roles and the
   master and slave controllers for each role. The P4Runtime server running on
   the device only keeps track of the `role_id` and `election_id` of the
-  controllers to determine which connected controller is master at any point of
-  time.
+  controllers (independently for each device) to determine which connected
```
antoninbas: +1

jafingerhut:
I have attempted to put the approved parts of this PR in https://github.com/p4lang/p4runtime/pull/155 and leave out the controversial parts.  Closing this PR in favor of that one.

PR #151, opened 2019-01-04
Change representation of signed integer values to use sign extension

jafingerhut:


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1649,53 +1649,85 @@ fits the encoded integer value. This representation achieves three goals:
...
+bit to represent, regardless of whether it is signed or unsigned.
+
+The shortest possible binary string for an integer $V$ that needs $A$
+bits to represent it is computed as:
 ~ Begin CPP
-if (value != 0)
-  encoded_string_size = (high_bit + 8) / 8;
-else
-  encoded_string_size = 1;
+encoded_string_size = floor((A + 7) / 8)
```
jafingerhut: From teverman: Would it be clearer to rename encoded_string_size to minimum_string_size? Just a thought.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1649,53 +1649,85 @@ fits the encoded integer value. This representation achieves three goals:
...
+
+Upon receiving a binary string, the P4Runtime receiver (whether the
+server or the client) does not impose any restrictions on the length
+of the string itself. Instead, the receiver verifies that the value
+encoded by the string fits within the expected type (signed or
+unsigned) and P4Info-specified bitwidth for the P4 object value.
+
+For received values expected to fit within a `bit<W>` type, the value
+is in range if, after removing all most significant 0 bits, the value
+remaining is `W` bits or less.
```
jafingerhut: From teverman: I think it's more accurate to specify the value width here, i.e "the remaining value's width is W bits or less".

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1649,53 +1649,85 @@ fits the encoded integer value. This representation achieves three goals:
...
+
+For received values expected to fit within an `int<W>` type, the value
+is in range if, after "undoing sign extension", the value remaining is
+`W` bits or less. To undo sign extension, start by eliminating the
```
jafingerhut: From teverman: Add width per previous comment.

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1649,53 +1649,85 @@ fits the encoded integer value. This representation achieves three goals:
...
 * It helps facilitate non-disruptive P4 program updates.
 
-The shortest possible binary string for an integer `value` with `high_bit`
-as the most-significant non-zero bit in the value is computed as:
+In particular, the kinds of P4 program updates that this
+representation facilitates are those where a P4Runtime server and
+client can continue to transmit P4Runtime messages between them when
+one has a P4Info file for version A of a P4 program, at the same time
+that the other has a P4Info file for version B of a P4 program, and
+those P4 programs differ in the bit widths of some values of type
```
antoninbas: I think the rest of the spec uses the "bitwidth" syntax

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1649,53 +1649,85 @@ fits the encoded integer value. This representation achieves three goals:
...
+`bit<W>` and/or `int<W>`.
+
+Note that this representation does _not_ make it possible to
+seamlessly change the type of a value from signed to unsigned, or vice
+versa.  If you attempt to do so, this mechanism can quietly change
+negative signed values to positive unsigned values, or vice versa.  It
+also limits the magnitude of the values transmitted to those that fit
+within the smaller of the bit widths supported by either end of the
```
antoninbas: same syntax comment as above

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1649,53 +1649,85 @@ fits the encoded integer value. This representation achieves three goals:
...
+encoded by the string fits within the expected type (signed or
+unsigned) and P4Info-specified bitwidth for the P4 object value.
+
+For received values expected to fit within a `bit<W>` type, the value
+is in range if, after removing all most significant 0 bits, the
+remaining value's width is `W` bits or less.
+
+For received values expected to fit within an `int<W>` type, the value
+is in range if, after "undoing sign extension", the remaining value's
+width is `W` bits or less. To undo sign extension, start by
```
antoninbas: I may be wrong, but isn't it `W-1` bits here?
For example, if I expect `int<2>` (only 1 and -1 can be represented) and the client sends the 1-byte value `00000011b`, then after "undoing sign extension" I get a value (3) that fits in 2 bits but cannot be represented with `int<2>`.

jafingerhut: The type `int<W>` can always represent `2^W` different values, e.g. `int<2>` can represent any value in the range `[-2, 1]`.

A P4Runtime client/server, if it started with a value that it thought was type `int<2>` represented as `11b`, would _never_ send it as the byte `00000011b` using the proposed mechanism.  It would sign extend the 2-bit value with 1 bits, and send it as `11111111b`.  Both `11b` and `11111111b` represent the numeric value -1 in 2's complement representation.

The only way with the proposed changes that one of the client/server would send `00000011b` as the encoding for a value of type `int<W>` is if the numeric value was +3, and the type would thus have to be `int<3>` or larger at the sender.   If the receiver had type `int<2>` or `int<1>`, i would "un sign extend" the value down to `011b`, but would there have to stop because the 2 most significant bits differ, and reducing it more would change the sign of the value.  It would then detect that the value requires a width of W=3 bits, and since the receiver's type was `int<2>` or `int<1>` in this example scenario, the receiver would declare the value out of range.

jafingerhut: I can add example scenarios like this, either directly in this section explaining this mechanism, or in an appendix, if you think having such examples would clarify things for readers.  I know that many are not as intimately familiar with the gory details of 2's complement notation as I forced myself to become :)

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Done in the new branch I created in this repo: https://github.com/p4lang/p4runtime/tree/issue111-change-signed-int-representation

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Done in the new branch I created in this repo: https://github.com/p4lang/p4runtime/tree/issue111-change-signed-int-representation

jafingerhut:
I created and committed a new branch in this repository: Done in the new branch I created in this repo: https://github.com/p4lang/p4runtime/tree/issue111-change-signed-int-representation

Is there some way I should be able to create  PR from this branch?  I don't see a way to do that in Github's web interface yet.

docs/v1/P4Runtime-Spec.mdk
antoninbas: Yes I did forget that P4 used 2's complement for a second there :)
I was also thinking of a sender with `int<3>` and a receiver with `int<2>`, but didn't "unsign extend" properly. My only suggestion would be to maybe use binary string when talking about the bit pattern sent by the client and value for the numerical value. Apart from that I think it's clear.

jafingerhut: I have made a few changes from "value" to "bit string" in the version pushed to this branch: https://github.com/p4lang/p4runtime/tree/issue111-change-signed-int-representation

antoninbas:
I have reviewed the generated outputs and they look good
One minor nit: the rest of the spec uses "bitstring" consistently, not "bit string"
I think that the only reason why you don't see an option to create a PR is because the branch is one commit behind master. If you rebase master into your branch, you should see the option when you visit: https://github.com/p4lang/p4runtime/tree/issue111-change-signed-int-representation

jafingerhut:
I have replaced a couple of occurrences of "bit string" with "bitstring" for consistency with the rest of the document, rebased the changes on top of the latest master, and created this PR: https://github.com/p4lang/p4runtime/pull/152

antoninbas:
I'm closing this as it was superseded by #152

PR #153, opened 2019-01-07
Define canonical representation for P4Data compound types

### LINKED ISSUE #141: P4HeaderStack messages and the size of the P4 program's corresponding header stack variable
It seems reasonable that everything is fine if a P4HeaderStack message has the same number of headers in it as the maximum number of headers declared in the P4 program -- the first header in the message should correspond to element 0, the next element 1, etc.

If the message has more than the P4 program's maximum size for the header stack, is that an error?

If the message has less than the P4 program's maximum size for the header stack, is that an error, or should the message's header stack value be appended with enough invalid headers at the end of the header stack (i.e. the largest index values)?

The P4Info file seems to have no explicit representation of header stacks, and no way for the control plane to know what the maximum allowed size of a particular header stack is.

Same issues/questions exist for stacks with elements that are header_union.

antoninbas:
By introducing additional constraints, thus ensuring read-write
symmetry.

Fixes #141

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1960,10 +1960,9 @@ Just like its P4Info counterpart - `P4DataTypeSpec` -, `P4Data` uses a Protobuf
 We define a canonical representation for `P4Data` messages - therefore
 guaranteeing read-write symmetry - by introducing the following requirements:
 
- * The order of `members` in `P4StructLike`, the order of `bitstrings` in
-   `P4Header`, and the order of `entries` in `P4HeaderStack` and
-   `P4HeaderUnionStack` must match the order in the corresponding p4info.proto
-   type specification and hence the order in the corresponding P4~16~ type
+ * The order of `members` in `P4StructLike` and the order of `bitstrings` in
+   `P4Header` must match the order in the corresponding p4info.proto type
+   specification and hence the order in the corresponding P4~16~ type
    declaration.
 
  * An invalid header is represented by a `P4Header` message where the `is_valid`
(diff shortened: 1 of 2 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1957,10 +1957,27 @@ case (P4~16~ `bit<W>` type).
...
-`P4HeaderUnionStack` must match the order in the corresponding p4info.proto type
-specification and hence the order in the corresponding P4~16~ type declaration.
+We define a canonical representation for `P4Data` messages - therefore
+guaranteeing read-write symmetry - by introducing the following requirements:
+
+ * The order of `members` in `P4StructLike`, the order of `bitstrings` in
+   `P4Header`, and the order of `entries` in `P4HeaderStack` and
+   `P4HeaderUnionStack` must match the order in the corresponding p4info.proto
+   type specification and hence the order in the corresponding P4~16~ type
+   declaration.
```
jafingerhut: Actually, one quick thought.  Do we want to explicitly say that for stacks, that the order of `entries` is from element 0 of the stack, then additional elements in ascending order of index?  I know that is the "obvious" thing to do, but couldn't hurt to say explicitly.

antoninbas:
Note that I ultimately decided to go with the "simpler" stack representation (always provide all the entries even the invalid ones at the tail) because the size of an invalid header / header union should be very small (I would imagine at most 4 bytes). I know that I can define a crazy case with a 100,000-entry register, each entry storing a stack of size 100, in which case the size of a wild-card `ReadResponse` message on the entire register would be quite huge. But we should probably just optimize for meaningful cases.

docs/v1/P4Runtime-Spec.mdk
antoninbas: Sure, I'll update the PR.

PR #155, opened 2019-01-08
Small improvements in description of master-slave arbitration

jafingerhut:


jafingerhut:
This is an attempt to keep the useful parts of PR https://github.com/p4lang/p4runtime/pull/150 but leave out the controversial parts until they can be discussed and agreed upon separately from this PR.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -624,21 +624,35 @@ reasons:
...
-  master and slave controllers for each role. The P4Runtime server running on
-  the device only keeps track of the `role_id` and `election_id` of the
-  controllers to determine which connected controller is master at any point of
-  time.
+To support multiple controllers, P4Runtime uses the streaming channel (available
+via `StreamChannel` RPC) for session management. The workflow is described as
+follows:
+
+* Each controller instance (e.g. a controller process) can participate in one or
+  more roles. For each `(device_id, role_id)`, the controller receives an
```
stefanheule:
```suggestion
  more roles. For each (`device_id`, `role_id`), the controller receives an
```
For consistency with the rest of the document.

jafingerhut: Done.  While checking the PDF output, I also replaced this sentence:

"It is up to an arbitration mechanism outside of the server to decide on the controller roles, and the master and slave controllers for each role."

with this one, which seems more precise, given that P4Runtime servers _do_ participate in deciding which clients are masters and slaves:

"It is up to an arbitration mechanism outside of the server to decide on the controller roles, and the `election_id` values used for each `StreamChannel`."

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -624,21 +624,35 @@ reasons:
...
-  the device only keeps track of the `role_id` and `election_id` of the
-  controllers to determine which connected controller is master at any point of
-  time.
+To support multiple controllers, P4Runtime uses the streaming channel (available
+via `StreamChannel` RPC) for session management. The workflow is described as
+follows:
+
+* Each controller instance (e.g. a controller process) can participate in one or
+  more roles. For each (`device_id`, `role_id`), the controller receives an
+  `election_id`, which may be the same numeric value across multiple roles. For
```
antoninbas: could we replace "which may be the same numeric value across multiple roles" with a proposal made by Stefan in the Github issue:

> this election_id can be the same for different roles and/or devices, as long as the tuple (role_id, device_id, election_id) is unique

Based on feedback from Waqar, it seems that the most common case is "same election id advertised to all devices in a given role" and I don't think we should emphasize the other - less frequent - use case ("same election id across roles").

PR #156, opened 2019-01-08
Improve P4Runtime Value Set support

### LINKED ISSUE #134: Consider adding ValueSet programming example for value_set with 2+ fields
e.g. for a P4_16 parser state with a statement like `transition select ({hdr.ipv4.version, hdr.ipv4.fragment_offset, hdr.ipv4.protocol}) { ... }`

Or if configuring such a value_set is not supported by P4Runtime API v1.0, state that.

### LINKED ISSUE #135: Should ValueSet in p4info.proto have match_fields like Table does, instead of bitwidth?
In p4runtime.proto, ValueSetEntry has a repeated match field of type FieldMatch, just like the TableEntry message does.

That seems reasonable to me.  What seems wrong is that in the p4info.proto, message Table has a repeated match_fields of type MatchField, but message ValueSet has bitwidth.

antoninbas:
The objectives of this change is to:
 * cover more cases supported by the P4_16 spec
 * provide more intuitive runtime programming
 * be more future-proof to reduce the potential changes in future
   P4Runtime versions

In particular, we now support structs with bit<W> members as the type
parameter for value_set. The members of the struct may have different
match types.

Fixes #135
Fixes #134


antoninbas:
@jafingerhut please take a look, if it passes your review I'll add more people :)

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1434,13 +1434,117 @@ The `ValueSet` message defines the following fields:
...
+   program. If the Value Set is supported, the `match` field will include one
+   `MatchField` message for each field in the struct, with the following fields:
+
+ * `id`: must be unique with respect to the other `match` entries. If the P4Info
+   message was generated from a P4 compiler, we recommend that the IDs be
+   assigned incrementally, starting from 1, in the same order as the fields in
+   the P4 struct declaration.
+ * `name`: set to the name of the corresponding struct field.
+ * `annotations`: set to the list of P4 annotations associated with the struct
+   field.
```
jafingerhut: Maybe say something like this here?   "except for the `@match` annotation, if present -- see the `match` field below."

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1434,13 +1434,117 @@ The `ValueSet` message defines the following fields:
...
+value_sets {
+  preamble {
+    id: 0x03000001
+    name: "pvs"
+  }
+  match {
+    id: 1
+    bitwidth: 8
+    match_type: EXACT
+  }
```
jafingerhut: Each of the three `match` messages should have a `name` field.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1434,13 +1434,117 @@ The `ValueSet` message defines the following fields:
...
+~ Begin Prototext
+value_sets {
+  preamble {
+    id: 0x03000001
+    name: "pvs"
+  }
+  match {
+    id: 1
+    bitwidth: 8
+    match_type: EXACT
```
jafingerhut: Shouldn't the `match_type` message be nested inside of a `match` message, that is itself inside of the `match` message of line 1472?

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Same comment about the nesting of `match_type` and `other_match_type` inside of `match` messages that are themselves nested inside of the given `match` messages?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1434,13 +1434,117 @@ The `ValueSet` message defines the following fields:
...
+  match {
+    id: 3
+    bitwidth: 32
+    other_match_type: "custom"
+  }
+  size: 4
+}
+~ End Prototext
+
+Although not mentionned in the P4 specification, P4Runtime also supports the
```
jafingerhut: typo: "mentionned" -> "mentioned"

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3496,19 +3594,44 @@ state parse_l2 {
...
+~ Begin Prototext
+value_sets {
+  preamble {
+    id: 0x03000001
+    name: "trill_types"
+  }
+  match {
+    id: 1
+    bitwidth: <ETH_TYPE_BITWIDTH>
+    match_type: EXACT
```
jafingerhut: Same comment as above on `match_type` being inside of another `match` message

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4972,5 +5110,85 @@ primarily for the purpose of adding features for the P4Runtime API.
...
+~ Begin Prototext
+value_sets {
+  preamble {
+    id: 0x03000001
+    name: "pvs"
+  }
+  match {
+    id: 1
+    bitwidth: 8
+    match_type: EXACT
```
jafingerhut: Need `name` fields

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk
antoninbas: Good catch for the missing names, I added them.

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk
antoninbas: The other `match` message here is a oneof. It doesn't actually create a level of nesting, which is why the field numbering doesn't get a new scope. It is not exactly like a C union. What it means is that when serializing the message, at most one of the fields will be present.

antoninbas:
@jafingerhut Thanks for the comments. Do you see any logical issue with the new approach?

jafingerhut:
@antoninbas I see no logical problems with any of this.  Only the small issues I've already raised.  Everything else looks good, and I will approve it.  Definitely good to review with others, too, of course.

antoninbas:
Thanks. I'll move this to a new PR (I mistakenly created this one from my fork) and invite others to comment.

antoninbas:
Replaced by https://github.com/p4lang/p4runtime/pull/157

PR #157, opened 2019-01-08
Improve P4Runtime Value Set support

### LINKED ISSUE #134: Consider adding ValueSet programming example for value_set with 2+ fields

### LINKED ISSUE #135: Should ValueSet in p4info.proto have match_fields like Table does, instead of bitwidth?

antoninbas:
The objectives of this change is to:
 * cover more cases supported by the P4_16 spec
 * provide more intuitive runtime programming
 * be more future-proof to reduce the potential changes in future
   P4Runtime versions

In particular, we now support structs with bit<W> members as the type
parameter for value_set. The members of the struct may have different
match types.

Fixes #135
Fixes #134


jafingerhut:
While I could approve this, it seems best to me if at least one other person besides Antonin and I review this.  If you want me to approve it so it can be merged in without someone else's review, let me know and I can do that, as I have read it in detail and it looks good to me.

antoninbas:
@jafingerhut I'm working on scheduling a P4 API WG meeting next Wednesday at the usual time slot to discuss the other outstanding issues
My plan is to bring this up at the meeting and not to merge this PR until after the meeting

PR #158, opened 2019-01-09
Match the p4c implementation for duplicate keys in PkgInfo

antoninbas:
In this case, I think both solutions (use the last occurrence or reject
the program) are valid. However, the current p4c complier rejects the P4
program in case of duplicate keys. Since it is "hard" to change the
implementation (the error happens when instantiating the IR, not when
generating the P4Info), I suggest we just match the current
implementation...


antoninbas:
Merging this since the corresponding P4_16 spec change has been merged (https://github.com/p4lang/p4-spec/pull/720)

PR #159, opened 2019-01-16
Support numeric translation for counter / meter / register indices

### LINKED ISSUE #131: Another possible note for somewhere in Section 17.1 "PSA Metadata Translation" ?
If the following is explicitly planned _not_ to be supported in P4Runtime API version 1.0, it seems maybe worth mentioning somewhere in the spec:

+ Using a value with numeric runtime translation (e.g. PortId_t) as an index to a register, indexed counter, or indexed meter.

If it is planned to be supported, that also seems worth mentioning, but at least right now there doesn't seem to be anydata in the P4Info file that could indicate the type of the index.

antoninbas:
This is done in a similar fashion to other kinds of translation (e.g. for match
fields). It is convenient to support translation for indices as it is not
unusual for indirect counters or registers to be indexed by the ig / eg port.

Fixes #131

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1294,6 +1294,12 @@ by this counter array. Conversely, the `DirectCounter` message contains a
 `direct_table_id` field that carries the `unit32` identifier of the table to
 which this direct counter is attached.
 
+For indexed counters, the `Counter` message contains also an `index_type_name`
+field, which indicates whether the index has a [user-defined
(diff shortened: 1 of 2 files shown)


jafingerhut:
Looks reasonable to me.  Note that you will want to say "no, we won't go that far in supporting arbitrary things" at some point.  For example, I could ask about an index that is calculated as the equation `(((bit<16>) ingress_port_id) << 2) + ((bit<16> pkt_type_code)`.  I am _not_ suggesting that P4Runtime should support such things.

PR #161, opened 2019-01-16
Security concerns for the P4Runtime handshake

hashkash:
* State that the device configuration needs to be securely transferred to the controller
* Mention that the controller should authenticate the switch device_id during the handshake
* could->should for gRPC authentication

Looks like Andy Fingerhunt already mentioned the need for gRPC authentication, which is great, however, I've changed could -> should to make it stronger.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -675,10 +675,13 @@ described as follows:
   controllers with the highest `election_id` as master.
 
 * The mechanism via which the controller receives the P4Runtime server details
-  is implementation specific and beyond the scope of this specification.
-  Nevertheless, if the server information is transferred for example, from the P4
-  build machine via the network, it is recommended to use TLS or similar
-  encryption and authentication mechanisms to prevent eavesdropping attacks.
+  which includes the `device_id`, `ip` and `port`, is implementation specific
+  and beyond the scope of this specification. Similarly, the mechanism via which
+  the P4Runtime server receives its switch config (which includes the
+  `device_id`, `ip` and `port`) is beyond the scope of this specification.
+  Nevertheless, if the server details or switch config are transferred via the
+  network, it is recommended to use TLS or similar encryption and authentication
+  mechanisms to prevent eavesdropping attacks.
 
 * After the controller sends a `StreamMessageRequest` message to the P4Runtime
   server, the server sends a `StreamMessageResponse` message back to the
@@ -715,12 +718,6 @@ server trusts clients not to use a triple of values other than their own in
 their write requests. gRPC provides authentication methods [@gRPCAuth] that
 should be deployed to prevent untrusted clients from creating channels, and thus
 from making changes or even reading the state of the server.
-
-To avoid misconfiguration and abusing identifiers even after authenticating the
-gRPC channel, the controller should authenticate the P4Runtime server's
-`device_id`. Furthermore, the P4Runtime server should be protected from
-unauthorized users by restricting communication between the controller and
-switches only to the ports and direction relevant to P4Runtime channels.
 
 ## Default Role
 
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -674,6 +674,12 @@ described as follows:
   broken, in which case the P4Runtime server quickly sets one of the slave
   controllers with the highest `election_id` as master.
 
+* The mechanism via which the controller receives the P4Runtime server details
```
antoninbas: by P4Runtime server details / information, do you just mean the Forwarding Pipeline Configuration, or do you include some extra information as well?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -706,9 +712,15 @@ read state from the data plane, among others described later. P4Runtime relies
...
-their write requests. gRPC provides authentication methods [@gRPCAuth] that can
-be deployed to prevent untrusted clients from creating channels, and thus from
-making changes or even reading the state of the server.
+their write requests. gRPC provides authentication methods [@gRPCAuth] that
+should be deployed to prevent untrusted clients from creating channels, and thus
+from making changes or even reading the state of the server.
+
+To avoid misconfiguration and abusing identifiers even after authenticating the
+gRPC channel, the controller should authenticate the P4Runtime server's
+`device_id`. Furthermore, the P4Runtime server should be protected from
```
antoninbas: could you clarify what you mean by "authenticate the P4Runtime server's `device_id`" and why it is important?

docs/v1/P4Runtime-Spec.mdk
hashkash: From my experience with ONOS and P4Runtime, I'm referring to the target specific P4 device config (netcfg.json) file that is sent to the controller. The json includes the `device_id, ip, deviceKeyId, port`.

I understand the Forwarding Pipeline Configuration is sent to the switch only after the switch responds to the controller's `StreamMessageRequest`.

antoninbas: Then it is worth clarifying. Also the text mentions the P4 build machine. By definition, this is the machine that produces the Forwarding Pipeline Config and sends it to the controller. I don't know what `deviceKeyId` is, I don't think that is a P4Runtime concept.

docs/v1/P4Runtime-Spec.mdk
antoninbas: If the controller issues both the `StreamMessageRequest` and the `SetForwardingPipelineConfigRequest`, how can the device ids be different?
The controller is in charge of building the `SetForwardingPipelineConfigRequest` message from the `ForwardingPipelineConfig` message sent by the "P4 build machine".

docs/v1/P4Runtime-Spec.mdk
hashkash: Sure, I will clarify that. Regarding the `deviceKeyId`, your're right, it's not a part of the P4Runtime.

docs/v1/P4Runtime-Spec.mdk
hashkash: I agree with you that the`StreamMessageRequest` and the `SetForwardingPipelineConfigRequest` will have the same device ids as they are from the controller.
However, my suggestion is to have the controller ask the switch for it's `device_id`, and then check if `device_id` the switch sends is the one the controller expects. Does that make sense?

antoninbas: Not sure I follow. The switch doesn't just choose its device id, it is assigned by the controller or a separate entity.
 Basically I would imagine the controller always has access to this information for every device from the get go:
(device_id, switch IP , switch port, pub key)
One the secure gRPC channel is created, summing that somehow there is a device_id mismatch, the controller would get an error status: INVALID_ARGUMENT, "Unknown device_id"
You cannot ask the switch for the device id BTW, because several devices may be on the switch, with a single P4Runtime service.
Maybe the issue is that I'm not sure about what point you are trying to make here, besides "the Forwarding Pipeline Config should be delivered to the switch in a secure fashion". Since the release is supposed to happen very soon (it was supposed to happen today), maybe we should have a quick call to sync-up, or only keep the section for which we are on the same page, for this release of P4Runtime.

PR #163, opened 2019-01-16
Do not require order to be preserved for repeated fields

### LINKED ISSUE #110: Question about read-write symmetry and repeated fields, and/or sequences
Is it ever important that read-write symmetry should preserve the
_order_ of multiple values for a repeated field?  If not, is it then
OK for read-write symmetry to have different orders there?  That would
definitely appear "syntactically different" but "semantically-same".
If it is intended to allow that, it might be good to mention that as
explicitly allowed and expected.

For example, when reading all entries in a table, is it expected that
they are returned in some predictable order, mandated across all
P4Runtime implementations?

### LINKED ISSUE #119: Questions on action profile member and group update messages
9.2.1. Action Profile Member Programming

If a target implementation restricts an action selector group to
contain only actions with the same action_id, is it an error to do a
MODIFY ActionProfileMember message that changes the action_id of a
member action that is currently part of such a group with 2 or more
members?

Is it OK to make such a change when it is part of a group with exactly
1 member (i.e the member being modified)?

Is it allowed to add a member to more than one group at the same time?
Why or why not?  Either way, I don't think it is mentioned in the spec
right now.

9.2.2. Action Profile Group Programming

Is the list of members optional when deleting a group?  If it must be
provided, why?  If it need not be provided, it should be explicitly
stated in the docs.

antoninbas:
In our definition of read-write symmetry.

This commit also includes some clarifications regarding action selector
programming.

Fixes #110
Fixes #119

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1644,14 +1644,25 @@ observed_value = server.read(p4_entity)
 assert(intended_value == observed_value)
 ~ End Pseudo
 
-To ensure read-write symmetry, the rest of this document tries to offer canonical
-representations for various data types, but this principle should be thought of
-where it falls short. Ensuring this will allow client software to recover
-programmatically from failures that can affect the switch stack software,
-communication channel, or the client replicas. If `Read` RPC returns a
+To ensure read-write symmetry, the rest of this document tries to offer
+canonical representations for various data types, but this principle should be
+thought of where it falls short. Ensuring this will allow client software to
+recover programmatically from failures that can affect the switch stack
+software, communication channel, or the client replicas. If `Read` RPC returns a
 semantically-same but syntactically-different response then the client would
 have to canonicalize the read values to check its internal state, which only
 pushes the protocol's complexities to the client implementations.
+
+In order to avoid placing too much burden on the P4Runtime server
+implementation, we do not in general mandate that the order of values in a
+Protobuf repeated field be preserved. For example, the server is not required to
+preserve the order of the `match` fields in a `TableEntry` message. If there is
+a specific case for which the order is significant and / or needs to be
+preserved, it will be explicitly stated in this document. The
+`MessageDifferencer` class [@ProtoMessageDifferencer] included in the Protobuf
+C++ API supports comparing messages while treating repeated fields as sets, so
+that different orderings of the same elements are considered equal. This method
+of comparing Protobuf messages may come at a cost in performance.
 
 ## Zero as Reserved Value
 
@@ -2845,7 +2856,8 @@ following semantics.
   `FAILED_PRECONDITION`. If needed, the action profile group should first be
   modified to remove the member from the group. The member must not be
   referenced in the table action of any table entry, or the server must also
-  return `FAILED_PRECONDITION`.
+  return `FAILED_PRECONDITION`. `member_id` is the only field which is
+  considered when performing a `DELETE` and every other field will be ignored.
 
 ### Action Profile Group Programming
 
@@ -2900,7 +2912,13 @@ following semantics.
 - `DELETE`: Delete the group entry and deallocate the `group_id`. The group must
   not be referenced in the table action of any table entry, or the server must
   return a `FAILED_PRECONDITION` error code. If the `group_id` is invalid, the
-  server must return `NOT_FOUND`.
+  server must return `NOT_FOUND`. `group_id` is the only field which is
+  considered when performing a `DELETE` and every other field will be ignored.
+
+When setting the group membership with `INSERT` or `MODIFY`, the `members`
+repeated field must not include duplicates, &ie; members with the same
+`member_id`. The `weight` field is used instead to logically "repeat" the member
+inside the group.
 
 ### One Shot Action Selector Programming { #sec-oneshot }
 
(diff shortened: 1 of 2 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3013,6 +3031,13 @@ updates](#sec-batching-and-ordering-of-updates). Members need to be inserted
 first, then the group needs to be created, and finally the match entry can be
 inserted. Therefore, 3 distinct `WriteRequest` batches are required.
 
+It is possible to include several `ActionProfileAction` messages with the same
+exact `action` specification in one `ActionProfileActionSet` message. However,
+the P4Runtime client is encouraged not to do so, as the same can be achieved by
+using the `weight` field. Note that to preserve read-write symmetry, the server
+must not coalesce multiple `ActionProfileAction` messages with the same `action`
+specification into one.
+
```
stefanheule: Is there are reason we allow duplicates here but not for the members of action profile groups?  Both have a weight that can be used to "repeat" elements, so I'm not sure why we are making a distinction.  Would it make sense to also disallow duplicates here, as we don't expect anyone to use them, there is a clear alternative (weight), and it simplifies targets to not having to deal with repeats?

antoninbas: As far as I'm concerned, the burden on the P4Runtime server would be to have to compare the action specifications to make sure they are all distinct. By allowing repeats, we actually simplify the target IMO, since it can blindly configure the HW (but maybe that depends on the HW).

Note that for traditional action selector programming, we allow 2 distinct members (with distinct member ids) to have the same action specification, which is similar to what is allowed here for one-shot selectors. What we disallow is duplicate member ids in an action profile group definition, which doesn't really have an equivalent for one-shot programming.

stefanheule: Makes sense, thanks for the explanation.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2900,7 +2912,13 @@ following semantics.
...
   not be referenced in the table action of any table entry, or the server must
   return a `FAILED_PRECONDITION` error code. If the `group_id` is invalid, the
-  server must return `NOT_FOUND`.
+  server must return `NOT_FOUND`. `group_id` is the only field which is
+  considered when performing a `DELETE` and every other field will be ignored.
+
+When setting the group membership with `INSERT` or `MODIFY`, the `members`
+repeated field must not include duplicates, &ie; members with the same
+`member_id`. The `weight` field is used instead to logically "repeat" the member
+inside the group.
```
jafingerhut: Would it be worth explicitly saying something like "It is explicitly allowed for the same member id to be added to multiple groups at the same time. If, as a result, an implementation 'stores' the action id and parameters in the target in multiple locations, the server must update all of those locations when a request to modify such a member is made."

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1644,15 +1644,26 @@ observed_value = server.read(p4_entity)
...
+thought of where it falls short. Ensuring this will allow client software to
+recover programmatically from failures that can affect the switch stack
+software, communication channel, or the client replicas. If `Read` RPC returns a
 semantically-same but syntactically-different response then the client would
 have to canonicalize the read values to check its internal state, which only
 pushes the protocol's complexities to the client implementations.
 
+In order to avoid placing too much burden on the P4Runtime server
+implementation, we do not in general mandate that the order of values in a
+Protobuf repeated field be preserved. For example, the server is not required to
```
jafingerhut: Do we want to qualify this with something like "in a repeated field that is anywhere inside of an `Entity` message"?  I believe this comment doesn't really apply to repeated fields used outside of `Entity` messages.

jafingerhut:
Added 2 comments, and opened a couple of pretty focused small issues related to action selectors that I thought of while reviewing this.

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk
antoninbas: Could you clarify what you mean by that? What kind of repeated field outside of `Entity` do you have in mind (that can be both used in a WriteRequest and a ReadResponse)?

jafingerhut: Maybe it is already qualified somewhere near here, but there are several repeated fields defined in p4info.proto and p4runtime.proto that can never appear inside of an `Entity` message, and thus also cannot appear inside of a WriteRequest or ReadResponse message, and this text isn't saying anything about whether order is significant or insignificant for those, is it?

antoninbas: Given that this section as a whole only talks about Write & Read, and therefore entities, I'm not sure where the confusion is coming from, but I don't mind adding that clarification. Are you referring to the updates repeated field used to send batches for example? Because this field never appears in a ReadResponse, and there is no real symmetry to talk about.

jafingerhut: Re-reading the beginning of the section I am commenting on here, I think it is fine to commit this as is now, without further clarification.  It does seem to properly qualify which repeated fields it could possibly be talking about to write and read requests.  Thanks.

jafingerhut:
I approve of the changes as they are now, and would mark them approved on Github if I could see how to do it.  It currently shows me as having some unresolved comments, even though I have attempted to resolve any that I can find.

PR #165, opened 2019-01-18
Add to the spec that idle timeout for default entries is not supported

antoninbas:
I don't see a use case for this and this would potentially make
P4Runtime implementation trickier.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2744,6 +2744,12 @@ value they can support, rather than reject the TableEntry write with an error
 code. Similarly, each target should do its best to provide reasonably-accurate
 values for `time_since_last_hit`.
 
+P4Runtime does not support idle timeout for default entries. When the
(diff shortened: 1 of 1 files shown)


antoninbas:
This seems to be consistent with the PSA spec:

> PSA does not require a timeout value for a default action entry. The reason for not making this mandatory in the specification is that tthe default action may not have an explicit table entry to represent it, and also there are no known compelling use cases for a controller knowing when no misses have occurred for a particular table for a long time. The default action entry will not be aged out.

I suggest that in P4Runtime we return an error if the client attempts to set a TTL value for a default entry.

antoninbas:
Actually the precedent here seems to be to also make support optional in P4Runtime (see action selector groups for which members use a different action function). I don't know if the 2 features are really comparable, because the PSA explicitly states that there is no compelling use case for default entry idle timeout... @jafingerhut any thoughts?

jafingerhut:
Given what PSA says, it seems reasonable to me in P4Runtime to make it an error to attempt to set this.

PR #171, opened 2019-01-20
Changes for nested struct

hesingh:
Allow a bit-vector nested struct inside a P4 header.
Also, allow a P4 header to include a boolean.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1883,18 +1883,21 @@ specification.
 | `void`         | error                 | error        | error           |
 | `error`        | error                 | error        | allowed         |
 | `match_kind`   | error                 | error        | error           |
-| `bool`         | error                 | error        | allowed         |
+| `bool`         | allowed               | error        | allowed         |
 | `enum`         | allowed[^enum_header] | error        | allowed         |
 | `header`       | error                 | allowed      | allowed         |
 | header stack   | error                 | error        | allowed         |
 | `header_union` | error                 | error        | allowed         |
-| `struct`       | error                 | error        | allowed         |
+| `struct`       | allowed[*st_header]   | error        | allowed         |
 | `tuple`        | error                 | error        | allowed         |
 |----------------|-----------------------|--------------|-----------------|
 ~
 
 [^enum_header]: an `enum` type used as a field in a `header` must specify a
     underlying type and representation for `enum` elements.
+[*st_header]: a `struct` or nested 'struct' type that has the same properties, 
+              used as a field in a `header` must contain only 
+              `bit<W>`, `int<W>`, a Serializable Enum, or a `bool`.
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
Hemant, this topic was discussed at a P4 API working group meeting, and there are P4Runtime implementations that are far enough along, with a restriction of table key fields and action parameters being restricted to only type `bit<W>`, that a decision was made to limit the P4Runtime v1.0 specification to that.

hesingh:
Understood, Andy - thanks.

I added the changes in this PR for whenever the spec can be updated in future - no rush.  Likewise, I am developing the implementation and when its done, the community can decided how to add support to the repo(s).

hesingh:
These PI spec changes are planned for after the release of PI P4Runtime v1.0.0  which works with P4-16, v1.1.0.  PI P4Runtime v1.0.0 has a plan to release towards end of January 2019.

I would like to summarize the changes:

In the notes below, a struct is allowed to be a nested bit-vector struct.

1. Support struct as parameter to Action.
2. Support struct for a Table Key.
3. Support serialization of nested struct/header.

I will update an existing issue (https://github.com/p4lang/p4c/issues/1670) with implementation updates.

antoninbas:
I suggest you close this PR and open a Github issue instead, that we can tag with "post-v1.0"
I haven't figured out how we will maintain different versions on the spec in parallel in this repo yet, so it's possible that such a change will not go into master directly. 
Note that your patch currently only updates the specification document, it doesn't make any of the corresponding changes to the Protobuf files.

hesingh:
> I suggest you close this PR and open a Github issue instead, that we can tag with "post-v1.0"
> I haven't figured out how we will maintain different versions on the spec in parallel in this repo yet, so it's possible that such a change will not go into master directly.
> Note that your patch currently only updates the specification document, it doesn't make any of the corresponding changes to the Protobuf files.

Ok, will do.  In a private repo, I have made changes to p4c/control-plane/p4runtime/proto/p4/config/v1/.  I am not sure if you meant a proto directory in the spec.  thanks.

antoninbas:
There is no "p4c/control-plane/p4runtime/proto/p4/config/v1"
p4lang/p4runtime is imported as a submodule of p4c, so everything under "p4c/control-plane/p4runtime" comes from this repo.

hesingh:
> There is no "p4c/control-plane/p4runtime/proto/p4/config/v1"
> p4lang/p4runtime is imported as a submodule of p4c, so everything under "p4c/control-plane/p4runtime" comes from this repo.

Understood.  However, when I added a new P4Ids message to p4RuntimeSerialization.cpp file, I had to go to the path I stated to add the new image to 'message P4Ids' in p4info.proto.  Then my p4c build passed.  I think there is something about the p4runtime directory I don't understand.  thanks.

hesingh:
New issue is:

https://github.com/p4lang/p4c/issues/1691

PR #172, opened 2019-01-23
Include more information about operations allowed on constant tables

### LINKED ISSUE #162: Add a quick sentence in the spec about reading entries from const tables
(tables with only static entries)

antoninbas:
Fixes #162

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2632,6 +2632,27 @@ indirect tables - tables with an ActionProfile or ActionSelector
 `implementation` property - to a constant `NoAction` action entry, with the hope
 that it would simplify the implementation of the P4Runtime service.
 
+### Constant Tables
+
+Constant tables are defined as tables whose match entries are immutable. They
+are identified by the `is_const_table` flag in P4Info. The only write updates
+which are allowed for constant tables are `MODIFY` operations on the default
+action, assuming the default action itself is not constant. If the P4Runtime
+client attempts to perform any other kind of write update on a constant table,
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2404,6 +2404,22 @@ nonsensical values (such as an invalid action field). In the case of a *keyless*
 table (the table has an empty match key), the server must reject all attempts to
 `INSERT` a match entry and return an `INVALID_ARGUMENT` error.
 
+If the P4Runtime client attempts to perform any kind of write update on a
```
jafingerhut: Minor nit: It is still legal to perform a write update on a table with `is_const_table = true` if it is a MODIFY of its default action, and the default action is not declared `const`.  Maybe some qualification of this statement could indicate that whether modifying the default action is legal is independent of the value of `is_const_table` ?

jafingerhut: Or at least that is my understanding of current implementations in BMv2 and the way the specs are written, that this combination is possible.

antoninbas: You're correct. The P4_16 spec actually includes an example table declaration with immutable entries but a non-const default action. I'll update my PR.

antoninbas: I updated the paragraph and moved it to its own sub-section. PTAL.

antoninbas:
> LGTM. There is one occurrence of P4_16 that I think Madoko markup format is something like P416

Fixed the P4_16 mention

PR #175, opened 2019-01-23
Clarify meaning of max_size and max_group_size

### LINKED ISSUE #168: Should server auto-replace action selector group max_size of 0 with a non-0 value?
There are two places I have found where it says this in the section "Action Profile Group Programming":

"If the `max size` is not known at group creation-time, the client may
  leave this field unset (default value 0), in which case the static
  `max_group_size` value will be used and the group will be able to include up
  to `max_group_size` weighted member entries."

and later when describing the `INSERT` operation:

"If the client does
  not set `max_size`, the default value (0) implies that the statically-defined
  maximum size should be used for this group."

It seems like doing this could break read-write symmetry, similar to issue https://github.com/p4lang/p4runtime/issues/166

If so, and we want to avoid that, perhaps a `max_size` that is 0 should be either left at 0, or an error?  An error seems reasonable, unless there is a known use case for creating a group that can never have members added to it.

antoninbas:
Fixes #168

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -3093,7 +3093,7 @@ in the P4Info message:
 
 * If `max_group_size` is greater than 0, then `max_size` must be greater than 0,
   and less than or equal to `max_group_size`. We assume that the target can
-  support selector groups for which the sum of all member weigths is up to
+  support selector groups for which the sum of all member weights is up to
   `max_group_size`, or the P4Runtime server would have rejected the Forwarding
   Pipeline Config. If `max_size` is greater than `max_group_size`, the server
   must return `INVALID_ARGUMENT`.
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1253,7 +1253,7 @@ The `ActionProfile` message includes the following fields:
   instance of a PSA Action Selector extern.
 
 * `size`, an `int64` representing the maximum number of member entries that the
-  Action Profile can hold, or, in the case of an Action Profile, the maximum sum
+  Action Profile can hold, or, in the case of an Action Selector, the maximum sum
   of all member weights across all selector groups.
 
 * `max_group_size`, an `int32` representing the maximum sum of all member
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1215,7 +1215,7 @@ The `Action` message defines the following fields:
       [user-defined type](#sec-user-defined-types); this is useful for
       [translation](#sec-psa-metadata-translation).
 
-### `ActionProfile`
+### `ActionProfile` { #sec-p4info-action-profile}
 
 `ActionProfile` messages are used to specify all available instances of Action
 Profile and Action Selector PSA externs.
@@ -1257,7 +1257,10 @@ The `ActionProfile` message includes the following fields:
   groups for an Action Selector.
 
 * `max_group_size`, an `int32` representing the maximum number of *weighted*
-  member entries in any given Selector group, or 0 for an Action Profile.
+  member entries in any given Selector group, or 0 for an Action Profile. PSA
+  programs can use the `@max_group_size` annotation to provide this value for
+  Action Selectors. If the annotation is omitted, the P4Info field will default
+  to 0.
 
 ### `Counter` & `DirectCounter`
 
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3086,11 +3081,34 @@ repeated field must not include duplicates, &ie; members with the same
...
+#### Rules on Setting `max_size` { #sec-max-size-rules}
+
+The valid values for `max_size` depend on the static `max_group_size` included
+in the P4Info message:
+
+* If `max_group_size` is greater than 0, then `max_size` must be greater than 0,
+  and less than or equal to `max_group_size`. We assume that the target can
+  support up to `max_group_size` *weighted* member entries, or the P4Runtime
+  server would have rejected the Forwarding Pipeline Config. If `max_size` is
+  less than `max_group_size`, the server must return `INVALID_ARGUMENT`.
```
jafingerhut: Should "less than `max_group_size`" be "greater than `max_group_size`" here?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1257,7 +1257,10 @@ The `ActionProfile` message includes the following fields:
   groups for an Action Selector.
 
 * `max_group_size`, an `int32` representing the maximum number of *weighted*
-  member entries in any given Selector group, or 0 for an Action Profile.
+  member entries in any given Selector group, or 0 for an Action Profile. PSA
```
jafingerhut: I may be confused here, but it seems that here it is saying that max_group_size is the maximum _number_ of member entries, whereas later it is saying that max_group_size is the maximum total weight of all member entries.

Is it sometimes one, sometimes another?  In which cases?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3109,7 +3127,10 @@ One shots are programmed by choosing the `ActionProfileActionSet` message as the
 
 * `weight` specifying the probability of the action's selection at runtime. 0 is
   not a valid `weight` value and the server must return `INVALID_ARGUMENT` if
-  the client attempts to use it.
+  the client attempts to use it. The sum of all weights across all
+  `ActionProfileAction` messages for that `ActionProfileActionSet` message must
```
jafingerhut: Here is the "later" I referred to in an earlier comment, where it says sum of all weights, rather than number of entries.

docs/v1/P4Runtime-Spec.mdk
antoninbas: Fixed

docs/v1/P4Runtime-Spec.mdk
antoninbas: It is always meant to be the sum of member weights, but I guess "weighted member entries" doesn't make this clear, so I made a bunch of edits to be more consistent.

antoninbas: PTAL.

stefanheule: This might be a stupid question, but why is it the sum of weights rather than the count?  I assumed the hardware stores each member + weight, so the limit would be regardless of the value of the weight.

jafingerhut: Not a stupid question at all.  I do not claim knowledge of all hardware implementations of weighted cost multipath load balancing, but at least one common way is to implement equal cost multipath load balancing in hardware, with one hardware entry for each equal cost member, and then use the weights to determine how many times to duplicate each weighted entry in the hardware table, so that you achieve the relative frequencies specified by the weights.  Thus the hardware table size can end up containing "sum of weights" number of entries rather than "sum of number of entries".

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3086,11 +3081,35 @@ repeated field must not include duplicates, &ie; members with the same
...
+those locations when a request to modify such a member is made.
+
+#### Rules on Setting `max_size` { #sec-max-size-rules}
+
+The valid values for `max_size` depend on the static `max_group_size` included
+in the P4Info message:
+
+* If `max_group_size` is greater than 0, then `max_size` must be greater than 0,
+  and less than or equal to `max_group_size`. We assume that the target can
+  support selector groups for which the sum of all member weigths is up to
```
jafingerhut: typo: "weigths" -> "weights"

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1252,12 +1252,15 @@ The `ActionProfile` message includes the following fields:
 * `with_selector`, a boolean flag indicating whether this message describes an
   instance of a PSA Action Selector extern.
 
-* `size`, an `int64` representing the maximum number of *weighted* member
-  entries that this Action Profile or Selector can hold - across all selector
-  groups for an Action Selector.
+* `size`, an `int64` representing the maximum number of member entries that the
+  Action Profile can hold, or, in the case of an Action Profile, the maximum sum
```
ccascone:
```suggestion
  Action Profile can hold, or, in the case of an Action Selector, the maximum sum
```

docs/v1/P4Runtime-Spec.mdk
stefanheule: Got it, thanks for the explanation.

PR #180, opened 2019-02-01
Support nested bit-vector struct inside header or struct for p4runtime API generation

hesingh:
It was decided to also use ```P4DataTypeSpec``` for ```P4HeaderTypeSpec``` so that a struct can be supported inside a header.  The p4c frontend restricts what kind of fields can exist in a struct if the struct is inside a header.  Thus the protobuf changes to ```P4HeaderTypeSpec``` should be good. 

I think nested struct support in p4runtime should be a new feature and thus I would like to submit these changes to in v1.1.0.

Thanks much to Antonin for holding my hands regarding p4runtime and PI changes!

### DIFF
--- a/proto/p4/config/v1/p4info.proto
+++ b/proto/p4/config/v1/p4info.proto
@@ -85,7 +85,8 @@ message P4Ids {
     TABLE = 0x02;
     VALUE_SET = 0x03;
     CONTROLLER_HEADER = 0x04;
-
+    HEADER_OR_STRUCT = 0x05;  // used to test specific struct or header other than
+                              // controller_header for nested struct support.
     // PSA externs
     PSA_EXTERNS_START = 0x10;
     ACTION_PROFILE = 0x11;
--- a/proto/p4/config/v1/p4types.proto
+++ b/proto/p4/config/v1/p4types.proto
@@ -31,16 +31,20 @@ package p4.config.v1;
  * | void               | error     | error        | error           |
  * | error              | error     | error        | allowed         |
  * | match_kind         | error     | error        | error           |
- * | bool               | error     | error        | allowed         |
+ * | bool               | allowed   | error        | allowed         |
  * | enum               | allowed*  | error        | allowed         |
  * | header             | error     | allowed      | allowed         |
  * | header stack       | error     | error        | allowed         |
  * | header_union       | error     | error        | allowed         |
- * | struct             | error     | error        | allowed         |
+ * | struct             | allowed** | error        | allowed         |
  * | tuple              | error     | error        | allowed         |
  * |--------------------|-----------|--------------|-----------------|
  *
  * *if serializable
+ *
+ * **struct_header: a `struct` or nested 'struct' type that has the same properties, 
+ *   used as a field in a `header` must contain only 
+ *   bit<W>, int<W>, a Serializable Enum, or a bool.
  */
 
 // These P4 types (struct, header_type, header_union and enum) are guaranteed to
@@ -129,7 +133,7 @@ message P4StructTypeSpec {
 message P4HeaderTypeSpec {
   message Member {
     string name = 1;
-    P4BitstringLikeTypeSpec type_spec = 2;
+    P4DataTypeSpec type_spec = 2;
   }
   repeated Member members = 1;
   repeated string annotations = 2;
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1883,18 +1883,21 @@ specification.
 | `void`         | error                 | error        | error           |
 | `error`        | error                 | error        | allowed         |
 | `match_kind`   | error                 | error        | error           |
-| `bool`         | error                 | error        | allowed         |
+| `bool`         | allowed               | error        | allowed         |
 | `enum`         | allowed[^enum_header] | error        | allowed         |
 | `header`       | error                 | allowed      | allowed         |
 | header stack   | error                 | error        | allowed         |
 | `header_union` | error                 | error        | allowed         |
-| `struct`       | error                 | error        | allowed         |
+| `struct`       | allowed[*st_header]   | error        | allowed         |
 | `tuple`        | error                 | error        | allowed         |
 |----------------|-----------------------|--------------|-----------------|
 ~
 
 [^enum_header]: an `enum` type used as a field in a `header` must specify a
     underlying type and representation for `enum` elements.
+[*st_header]: a `struct` or nested 'struct' type that has the same properties, 
+              used as a field in a `header` must contain only 
+              `bit<W>`, `int<W>`, a Serializable Enum, or a `bool`.
 
 For example, the following P4~16~ objects involve complex types that need to be
 exposed in P4Runtime in order to support runtime operations on these objects.
--- a/proto/p4/v1/p4data.proto
+++ b/proto/p4/v1/p4data.proto
@@ -46,7 +46,7 @@ message P4Header {
   // If the header is invalid (is_valid is "false"), then the repeated
   // bitstrings must be empty.
   bool is_valid = 1;
-  repeated bytes bitstrings = 2;
+  repeated P4Data members = 2;
 }
 
 message P4HeaderUnion {


proto/p4/config/v1/p4info.proto
```diff
@@ -85,7 +85,8 @@ message P4Ids {
     TABLE = 0x02;
     VALUE_SET = 0x03;
     CONTROLLER_HEADER = 0x04;
-
+    HEADER_OR_STRUCT = 0x05;  // used to test specific struct or header other than
```
antoninbas: I don't know what this is for. If this is just for testing your code, it should not be merged into p4info.

proto/p4/config/v1/p4types.proto
```diff
@@ -129,7 +133,7 @@ message P4StructTypeSpec {
 message P4HeaderTypeSpec {
   message Member {
     string name = 1;
-    P4BitstringLikeTypeSpec type_spec = 2;
+    P4DataTypeSpec type_spec = 2;
```
antoninbas: this change (and the one to p4data.proto) break backward-compatibility and therefore cannot be included in a minor release such as 1.1.0 (see https://s3-us-west-2.amazonaws.com/p4runtime/docs/v1.0.0/P4Runtime-Spec.html#sec-p4runtime-versioning).
maybe you could look for a way to add support for this which does not break backward-compatibility, but if it makes things too messy it may not be accepted, unless there is a very strong driving use-case for the change.

proto/p4/config/v1/p4info.proto
hesingh: Sorry, I forgot to add a note that this change could be used for a struct for special handling.  Maybe, even a header that is not used by packet_in/packet_out.  I am using this entry in ```control-plane/p4RuntimeSerializer.cpp``` as follows:

```id = symbols.getId(P4RuntimeSymbolType::HEADER_OR_STRUCT(), name);```

Then the id is used to setup the Preamble.

```setPreamble(header->mutable_preamble(), id, name, name, annotations);```

If an alternative can be suggested how to get the id without using the new HEADER_OR_STRUCT, I can remove this change.

proto/p4/config/v1/p4types.proto
hesingh: I don't see any other way out.  

With these protobuf changes 'make check' passes for all p4c tests.  The same changes, when incorporated in p4lang/PI also pass 'make check'.   Anything else to test for breaking any backward compatibility?  

The community asked for nested bit-vector struct inside a header which has been checked in to p4c.  So, it's a natural extension of the work to change p4c p4runtime API generation.  Using these protobuf changes and changing p4c/control-plane/ code to support nested structs, the API generation ran fine and 'make check' for p4c and p4lang/PI also passed.  

Thus, some release juxtaposition to get this one in ASAP would help immensely.

proto/p4/config/v1/p4info.proto
antoninbas: You'll have to give a concrete example of when this is required by a runtime application. For now it is completely at odds with the other entities defined in the P4Info message.

proto/p4/config/v1/p4types.proto
hesingh: ok, understood about backward compatibility.

jafingerhut: One way to maintain backward compatibility in the face of such a change is _never_ remove a field, only add new fields.  If you want to change the type of an existing field, the new type should allow all of the previous legal values, plus new ones.  No previous legal value should ever be disallowed in the new type.

I don't know if this is a _good_ way to do it, but you could keep type_spec, and add a type_spec_extended, and in the text indicate that at most one of type_spec and type_spec_extended should be set in any message.  There are probably cleaner ways using protobuf messages to do something similar here.

antoninbas: @jafingerhut 

> Move fields into or out of a oneof: You may lose some of your information (some fields will be cleared) after the message is serialized and parsed. However, you can safely move a single field into a new oneof and may be able to move multiple fields if it is known that only one is ever set.

(https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility-issues)

However, I wonder if it is actually worth it to change the API at all. Ultimately headers are meant to be "flat" and I see the recent changes to allow structs in headers as just some syntactic convenience. I'm unsure whether there is any value / compelling use case in preserving the exact "structure" of the header?

hesingh: > One way to maintain backward compatibility in the face of such a change is _never_ remove a field, only add new fields. If you want to change the type of an existing field, the new type should allow all of the previous legal values, plus new ones. No previous legal value should ever be disallowed in the new type.
> 
> I don't know if this is a _good_ way to do it, but you could keep type_spec, and add a type_spec_extended, and in the text indicate that at most one of type_spec and type_spec_extended should be set in any message. There are probably cleaner ways using protobuf messages to do something similar here.

Excellent suggestion.  I will look into it and test and one option.

hesingh: > p4lang/p4runtime is used as a git submodule by both p4c and PI. If you have your own forks of these projects, you can set the p4runtime submodule ref point to whatever you want. However, if you want to point to a different remote (i.e. point to your own fork and not to a sha in p4lang/p4runtime), I believe you will need to change the remote for the submodule in your forks (https://stackoverflow.com/questions/913701/how-to-change-the-remote-repository-for-a-git-submodule).
> 
> Note that there is a less extreme way of supporting structs IMO. You can just ensure that all your headers are flattened when generating the P4Info in p4c. This was our original idea but we rolled back that change because of issues with the FlattenHeaders pass and its numerous dependencies. It is not ideal because the P4Info doesn't _strictly_ reflect the P4 program, but from an API perspective I would say it is very acceptable & sufficient. After all P4 headers are meant to be serializable / deserializable on the wire, and on the wire headers have a flat structure...

Thanks for the details on submodule for forks!

Both flatten passes are midend passes.  Thus, at least one pass would have to move to being a frontend pass because just after frontend.run(), p4runtime API is generated.  Each of frontend vs. midend have dependencies for specific passes which would need looking into which is a rat-hole we have hit in the pass by randomly adding FlattenHeaders to p4RuntimeSerializer.cpp.

jafingerhut: I think Antonin's comments may apply purely to the contents P4Runtime API messages, not to the internals of a P4 compiler, which can be discussed semi-independently.

hesingh: > @jafingerhut
> 
> > Move fields into or out of a oneof: You may lose some of your information (some fields will be cleared) after the message is serialized and parsed. However, you can safely move a single field into a new oneof and may be able to move multiple fields if it is known that only one is ever set.
> 
> (https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility-issues)
> 
> However, I wonder if it is actually worth it to change the API at all. Ultimately headers are meant to be "flat" and I see the recent changes to allow structs in headers as just some syntactic convenience. I'm unsure whether there is any value / compelling use case in preserving the exact "structure" of the header?

The link than Antonin sent out is when adding/removing oneof fields.  The P4HeaderTypeSpec does not even use oneof.   It has only one field and if a second field is added, then oneof will be used.

antoninbas: Yes Andy is correct. Your use-case justifies the use of nested structs in P4, but not necessarily in P4Runtime messages.

My link talks about creating new oneofs while maintaining backward-compatibility.

hesingh: > Yes Andy is correct. Your use-case justifies the use of nested structs in P4, but not necessarily in P4Runtime messages.
> 
> My link talks about creating new oneofs while maintaining backward-compatibility.

Duh, very sorry, I missed "adding" oneof in your link.   If p4runtime uses flat headers, doesn't the frontend flattens headers first?  We will have to move the midend FlattenHeaders pass as a frontend pass.  After all, this issue was filed which would need flattening by the frontend p4c.

https://github.com/p4lang/p4c/issues/1670

hesingh: @antoninbas Ah, got it.  We can leverage some code from the FlattenHeaders midend pass. 
So, the PI server also sends digest messages to the client in flattened form and does not use nested struct/header.

hesingh:
Closing this PR until we are ready to accept v2 changes.

Issue #125, opened 2018-12-22
Is there a "minimum message size" that a server must support for requests?


jafingerhut:
That might be a misleading way to ask the question in the summary line, but here are some specific examples:

Suppose a client tries to send 17 billion entity updates in a single write request batch.  What should a server unable to store that entire message do?

Is there some "minimum batch size" that all server implementations should be able to support?

Similarly for any other message with a 'repeated' field in a client->server message (perhaps we can ignore repeated things for action parameters or table key fields, since those will typically be pretty small for successfully compiled P4 programs, but maybe not, since a client can create erroneous long messages, e.g. with the same field id millions of times).

antoninbas:
I'll just provide some context here, but this an issue that is common to a lot of gRPC services.
By default gRPC limits message sizes to 4MB. In my case this is too small, because my `SetForwardingPipelineConfig` messages are usually larger than that, so I routinely increase the "max receive message size" at the server (https://github.com/p4lang/PI/blob/master/proto/server/pi_server.cpp#L565).

One important thing about this configuration parameter is it has to be set when building the server, and cannot be changed during its lifetime. It also impacts the entire gRPC server, not a specific service (e.g. P4Runtime). In practice I'm not sure how this would play out. It seems that when deploying a NOS which integrates a gRPC server, this should be available as a configuration parameter, potentially platform-dependent. The controller may be able to query that value, but has doesn't have any ability to modify the value. For write batches, the controller may be able to organize its batches so that they don't exceed the max size (which will probably always be >= 4MB, because there is no reason to modify the default size except to increase it...). For `SetForwardingPipelineConfig`, there is no possibility to split the message. However, because the config parameter can be platform-aware, we can ensure that it can accommodate all possible `SetForwardingPipelineConfig` messages. For example, I can come up with an upper-bound for all `SetForwardingPipelineConfig` messages for Tofino, and all Tofino-based platforms can set a size greater to this bound when starting the NOS (e.g. Stratum). I actually exposed this configuration parameter as a command-line argument in Stratum, and it can be leveraged by any deployment mechanism we choose for Stratum.

PR #181, opened 2019-02-14
Appendix with implementation advice for avoiding common pitfalls

stefanheule:
Large batches in the `Write` RPC can cause gRPC to error with `RESOURCE_EXHAUSTED`, because the default maximum size for metadata is very low.  This new section shows how to avoid that.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -5387,14 +5387,28 @@ This section contains practical advice for using P4Runtime.
 
 ### gRPC Metadata Maximum Size
 
-In gRPC, the status of a RPC request is sent as metadata, whose size is limited by the `grpc.max_metadata_size` gRPC channel argument.  By default, this limit is 8KB, which can be a problem for the `Write` P4Runtime RPC.  The `Write` RPC returns an individual error for every item in a batch (see Section [#sec-write-rpc]), which can quickly result in a status size over 8KB.  In that case, the gRPC server would not send the status, but instead send a `RESOURCE_EXHAUSTED` error, without any of the individual errors.
+In gRPC, the status of a RPC request is sent as metadata, whose size is limited
+by the `grpc.max_metadata_size` gRPC channel argument.  By default, this limit
+is 8KB, which can be a problem for the `Write` P4Runtime RPC.  The `Write` RPC
+returns an individual error for every item in a batch (see Section
+[#sec-write-rpc]), which can quickly result in a status size over 8KB.  In that
+case, the gRPC server would not send the status, but instead send a
+`RESOURCE_EXHAUSTED` error, without any of the individual errors.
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
Thanks for proposing adding a section like this.  I do not have enough detailed knowledge of gRPC to review these changes, but the question I asked in this issue https://github.com/p4lang/p4runtime/issues/125 and Antonin's response, while not exactly the same topic as your changes discuss, seem closely related (max metadata size vs. maximum message size).  If there is anything worth adding to what you wrote here on the maximum message size, and why increasing it might be useful for enabling larger binary blobs to be sent, that sounds useful.

stefanheule:
Thanks Andy for pointing that out.  I agree that the maximum message size is also something we should point out, maybe we can do that in a follow-up PR.  The metadata size is probably a bigger problem, because the limit is so small, so it seems quite common to run into this.

antoninbas:
@jafingerhut That is a good point. I actually brought that up with Stefan earlier and was planning on having a follow-up PR to his.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5381,4 +5381,21 @@ entity {
 ~ End Prototext
 
 
+## Guidelines for Implementations
+
+This section contains practical advice for using P4Runtime.
+
+### gRPC Metadata Maximum Size
+
+In gRPC, the status of a RPC request is sent as metadata, whose size is limited by the `grpc.max_metadata_size` gRPC channel argument.  By default, this limit is 8KB, which can be a problem for the `Write` P4Runtime RPC.  The `Write` RPC returns an individual error for every item in a batch (see Section [#sec-write-rpc]), which can quickly result in a status size over 8KB.  In that case, the gRPC server would not send the status, but instead send a `RESOURCE_EXHAUSTED` error, without any of the individual errors.
```
antoninbas: Could you wrap lines to 80 characters?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5381,4 +5381,21 @@ entity {
...
+
+To fix this problem, one can set the `grpc.max_metadata_size` option on the client channel.  This allows the client to receive more than 8KB of metadata, based on the new limit.  Note that the gRPC server does not have to change it's limit, as only the receiving side's limit is relevant.  The exact limit that is required depends on the maximum batch size, the length of error messages inside every `p4.Error`, as well as any other metadata that is being sent over the gRPC channel.  As a rule of thumb, it might make sense to allow for at least `8192 + MAX_UPDATES_PER_WRITE * 100` bytes of metadata.
+
+For example, in C++ one can create a client channel as follows:
+~ Begin CPP
+::grpc::ChannelArguments arguments;
+arguments.SetInt(GRPC_ARG_MAX_METADATA_SIZE, 8192 + MAX_UPDATES_PER_WRITE * 100);
```
antoninbas: could you include something like `const int MAX_UPDATES_PER_WRITE = 100000;` in the example, just in case someone thinks `MAX_UPDATES_PER_WRITE` is a standard gRPC or P4Runtime constant...

docs/v1/P4Runtime-Spec.mdk
stefanheule: Yes, of course.  I meant to, but forgot.

docs/v1/P4Runtime-Spec.mdk
stefanheule: Done.  I put 100 as the number, I don't think most people expect to use 10000 updates per batch.

PR #186, opened 2019-02-25
Explicitly list additional types supported as table search key fields

jafingerhut:
and action parameters.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -5246,8 +5246,13 @@ properties, but we may include on in future versions of the API.
 # Known-limitations of P4Runtime v1.0.0
 
 * `FieldMatch`, action `Param`, and controller packet metadata fields only
-  support unsigned bitstrings, i.e. values of type `bit<W>` (not the more
-  general `P4Data`).
+  support unsigned bitstrings, i.e. values of one of the following types (not
+  the more general `P4Data`):
+  * `bit<W>`
+  * an `enum` with underlying type `bit<W>`
+  * a `type` or `typedef` with an underlying type that is one of the above (or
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5246,8 +5246,13 @@ properties, but we may include on in future versions of the API.
 # Known-limitations of P4Runtime v1.0.0
 
 * `FieldMatch`, action `Param`, and controller packet metadata fields only
-  support unsigned bitstrings, i.e. values of type `bit<W>` (not the more
-  general `P4Data`).
+  support unsigned bitstrings, i.e. values of one of the following types (not
+  the more general `P4Data`):
+  * `bit<W>`
+  * an `enum` with underlying type `bit<W>`
```
antoninbas: can we emphasize that this is a "serializable" enum, and not the "default" enum?

jafingerhut: The word "serializable" appears nowhere in the P4_16 language spec today.  There are multiple occurrences in the P4_16 language spec of an enum having an underlying type `bit<W>`.

Perhaps the P4_16 language spec could/should be changed to introduce that name for this kind of `enum`, but at least right now using the word "serializable enum", while it may be common verbally between us, isn't in any spec that I know of.

jafingerhut:
I did check the formatting locally on my machine, visually inspecting the PDF file at least.  Sorry, I keep forgetting about the other way to submit PRs to this repo.

docs/v1/P4Runtime-Spec.mdk
antoninbas: my bad, this comes as a surprise to me

Issue #187, opened 2019-02-26
Should `Metadata` message in P4Info contain a `type_name` field?


jafingerhut:
The `MatchField` messages, and the `Param` message inside of an `Action` message, both have `int32 bitwidth` and `P4NamedType type_name` fields, where the `type_name` field is used if the P4_16 program defines a custom `type` for a table search key field (`MatchField`) or action parameter (`Param`).

The `Metadata` message inside of a `ControllerPacketMetadata` message has a `int32 bitwidth` field, but no `P4NamedType type_name` field.  I believe the intent is that user-defined metadata fields in Packet-In and Packet-Out messages could also have custom `type` definitions, e.g. `PortId_t`.

Should we add a `type_name` field inside of the `Metadata` messages for this purpose?

antoninbas:
That's correct. Without this field, there can be no translation for metadata fields in Packet In / Out messages. We should add it for the 1.0.1 release.

jafingerhut:
Created a PR for it already, in case your answer was "yes".

jafingerhut:
Closing issue, since relevant PR was merged.

PR #188, opened 2019-02-26
Add type_name field to P4Info Metadata message

jafingerhut:
Add a description of this field to the text of the section describing
the contents of ControllerPacketMetadata messages.

Also add a description of the type_name field in the ValueSet section.
It was already re-using the MatchField messages used by table search
key fields, and already had this type_name field in it, but it seems
best to explicitly mention that it can be included in MatchField
messages for ValueSet objects, too.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1398,7 +1398,10 @@ message contains the following fields:
       example below).
     * `annotations`, a repeated field of strings, each one representing a P4
       annotation associated to this metadata.
-    * `bitwidth`, an `int32` representing the size in bit of this metadata.
+    * `bitwidth`, an `int32` representing the size in bits of this metadata.
+    * `type_name`, which indicates whether the metadata field has a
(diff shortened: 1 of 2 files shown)


jafingerhut:
This time I remembered to create a fork in this repository, so you should be able to see preview of generated output if you want, I hope.  I checked PDF output on my machine anyway, and it looked reasonable.

These changes might not be what we want, of course.  They seemed reasonably small and quick to write, in case it is true that we want to add this type_name field based on issue: https://github.com/p4lang/p4runtime/issues/187

jafingerhut:
Understood.  No rush on merging it in on my part.  I only noticed this because of doing some nit-pick level checking of details in some notes I am writing up on how I believe P4Info files should be generated for arbitrary P4_16 programs that have user-defined `type` and/or `typedef` definitions for these types, with or without `p4runtime_translation` annotations on them.  Will send you a link to that when I think it is ready to review.

antoninbas:
I think this has been opened for long enough :)

PR #193, opened 2019-03-11
Add "Guide for Generating P4Info message" document

jafingerhut:
This is intended to be a supplement to the P4Runtime v1.0
specification, with details and examples for things that might not be
clear from reading the specification.

### DIFF
--- a/docs/v1/guidance-for-generating-p4info.md
+++ b/docs/v1/guidance-for-generating-p4info.md
@@ -1,5 +1,43 @@
 # Guidance for generating P4Info messages
 
+
+## Recommendations for URIs in `p4runtime_translation` annotations
+
+For the use of P4Runtime major version 1 with the PSA architecture,
+all URIs for `p4runtime_translation` annotations defined in the psa.p4
+include file will be of this form:
+
+    p4.org/psa/v1/<type_name>
+
+No other organization should use URIs beginning with `p4.org`.  They
+are reserved for use by the `p4.org` organization.
+
+It is recommended that another organization that wishes to use the
+`p4runtime_translation` annotation for types defined in an
+architecture they create, other than PSA, will be of this form:
+
+    <domain_name>/<proprietary_arch>/<type_name>
+
+For example, `mycompany.com/our_arch/PortId`.  The organization should
+document its rules for such URIs.
+
+For P4_16 `type` definitions for which P4Runtime translation is
+desired, where the `type` is not defined within an architecture, but
+in the user's P4 code, we recommend using a URI of this form:
+
+    <domain_name>/<arch>/<P4_program>/<type_name>
+
+Where here the `domain_name` is that of the organization or individual
+who developed or owns the P4 program, and the `<arch>` element is
+optional.  Again, the organization or individual should form and
+document a policy for how they form such URIs.  For example, they may
+create libraries of P4 code containing widely used types and
+translations for them, and wish to include an element in the URI
+representing the name of the library, its major version, etc.
+
+
+## Handling P4_16 `type` and the `p4runtime_translation` annotation
+
 The P4Runtime v1.0 specification restricts the types that it supports
 for the following kinds of things:
 
@@ -11,8 +49,8 @@ for the following kinds of things:
   defined in the P4Info message in a `Param` message
 + metadata fields in a header sent from data plane to controller, or
   from controller to the data plane, defined in the P4Info message in
-  a `Metadata` message, if a recently proposed [p4runtime PR
-  #188](https://github.com/p4lang/p4runtime/pull/188) is merged in.
+  a `Metadata` message (the `type_name` field was added to `Metadata`
+  messages in P4Runtime version 1.0.1).
 
 Later in this section, we will use the term "constrained value" for
 brevity, instead of repeating all of the kinds of objects listed
@@ -21,7 +59,8 @@ of the following types, but currently no others:
 
 + `bit<W>`
 + an `enum` with an underlying type of `bit<W>`, also called a
-  serializable `enum`
+  serializable `enum` (TBD whether all of the pieces needed to make
+  this work are actually supported for P4Runtime 1.0)
 + a `typedef` or `type` name that, when "followed back" to the lowest
   base type, is one of the above.  (As of the P4_16 language
   specification version 1.1, it is not required to support a `type`
@@ -29,8 +68,8 @@ of the following types, but currently no others:
   [p4runtime issue
   #192](https://github.com/p4lang/p4runtime/issues/192).)
 
-P4Info `MatchField`, `Param`, and probably soon also `Metadata`
-messages may optionally contain the following two fields:
+P4Info `MatchField`, `Param`, and `Metadata` messages may optionally
+contain the following two fields:
 
 + `int32 bitwidth`
 + `P4NamedType type_name`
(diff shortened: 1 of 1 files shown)


docs/v1/guidance-for-generating-p4info.md
```diff
@@ -0,0 +1,363 @@
...
+
++ table search key fields, defined in the P4Info message in a
+  `MatchField` message
++ fields of a `ValueSet`, also defined in the P4Info message in a
+  `MatchField` message
++ parameters specified by the control plane for a table action,
+  defined in the P4Info message in a `Param` message
++ metadata fields in a header sent from data plane to controller, or
+  from controller to the data plane, defined in the P4Info message in
+  a `Metadata` message, if a recently proposed [p4runtime PR
```
antoninbas: you may want to update this now that the PR has been merged, although it may be good to mention that it is not available in P4Runtime v1.0.0 (but will be in v1.0.1).

docs/v1/guidance-for-generating-p4info.md
```diff
@@ -0,0 +1,363 @@
...
++ an `enum` with an underlying type of `bit<W>`, also called a
+  serializable `enum`
++ a `typedef` or `type` name that, when "followed back" to the lowest
+  base type, is one of the above.  (As of the P4_16 language
+  specification version 1.1, it is not required to support a `type`
+  definition with a serializable `enum` as its base type.  See
+  [p4runtime issue
+  #192](https://github.com/p4lang/p4runtime/issues/192).)
+
+P4Info `MatchField`, `Param`, and probably soon also `Metadata`
```
antoninbas: same, may want to update this statement

docs/v1/guidance-for-generating-p4info.md
```diff
@@ -0,0 +1,363 @@
...
+Otherwise, `bitwidth` should be equal to `W` where `bit<W>` or `enum
+bit<W>` is the last element of `type_list(x)`.
+
+Note that all type names that are present as the value of a
+`type_name` field anywhere in a P4Info message must contain an entry
+describing it in the `type_info` field of the P4Info message.  Thus
+this `bitwidth` value could be considered redundant information, since
+it can be derived from the description of the type contained within
+the `type_info` field (see [this
+comment](https://github.com/p4lang/p4runtime/issues/189#issuecomment-471240515)).
```
antoninbas: I never replied to this comment. I agree that it is redundant information, but I believe it is a good thing to have here. I don't think we want some match fields / params to have a "valid" `bitwidth` field and others to have an "unset" `bitwidth` field. Keeping things uniform gives "cleaner" implementations, and probably makes the P4Info message easier to read IMO.

docs/v1/guidance-for-generating-p4info.md
```diff
@@ -0,0 +1,363 @@
...
+The basic design described here tries to keep things fairly
+straightforward to explain and understand, if a P4_16 program does so.
+
+If a constrained value `f` has a `type` name as the first element of
+the list `type_list(f)`, then that type name is used for `f`,
+regardless of whether it has a `p4runtime_translation` annotation.
+
+In the absence of such an annotation on that `type`, no P4runtime
+translation is done for that type, _even if a later type in
+`type_list(x)` does have such an annotation_.
```
antoninbas: Doesn't this text apply to the previous example (example 2)?

docs/v1/guidance-for-generating-p4info.md
```diff
@@ -0,0 +1,363 @@
...
+There is never anything put into a P4Info message because of `typedef`
+definitions in a P4 program.
+
+Note that the bit width of 10 appears in the P4Info message for any
+`type`s "built on top of" a `bit<10>`, _unless that type has its own
+`p4runtime_translation` annotation_.
+
+```
+typedef bit<10> T1uint_t;
+@p4runtime_translation("mycompany.com/psa/v1/T1_t", 32)
```
antoninbas: Nit-picking: I would expect the URI to be `"p4.org/psa/v1/<type_name>"` for types defined in psa.p4 (`"company.com/<proprietary_arch>/<type_name>"` for types defined in a proprietary architecture). For types defined in a P4 program, I would expect something like `"company2.com/[<arch>/]<P4_program>/<type_name>"`, with company2 being the device user and company being the vendor. I know I put in the spec that the arch name would ideally always be part of the URI, but what I had in mind were types introduced by the architecture specifically. For types defined by a user in their program, the architecture name is not as important as something identifying the particular P4 program the type is defined in. If a company defines a type in a separate p4 file (that they include in all their P4 programs), they could go with something like `"company2.com/std/<type_name>"`.

docs/v1/guidance-for-generating-p4info.md
jafingerhut: I will change the text to this: "defined in the P4Info message in
  a `Metadata` message (added in P4Runtime version 1.0.1)"

docs/v1/guidance-for-generating-p4info.md
jafingerhut: Removed "probably soon also"

docs/v1/guidance-for-generating-p4info.md
jafingerhut: I have replaced the text "(see [this comment](https...))" with:

"However, it is considered worth keeping this
small amount of redundancy for simplicity of interpreting the messages
that contain these `bitwidth` fields, without having to find the value
elsewhere in the P4Info message."

docs/v1/guidance-for-generating-p4info.md
jafingerhut: Nit-pick away!  I have added a section near the beginning of this document attempting to capture your recommendations here, and stating them as recommendations.  Feel free to review and comment on it after I have committed it.

jafingerhut: I have replaced all example URIs of hte form "mycompany.com/psa/v1/<type_name>" with "mycompany.com/myco_p4lib/v1/<type_name>".

jafingerhut: I meant "Nit-pick away!" in the sense of "go ahead and do it!", not "go away".

jafingerhut:
Regarding this comment:

"Would it make sense to have the basic common case as an example?

@p4runtime_translation("p4.org/psa/v1/PortId_t", 32)
type bit<9> PortId_t;
PortId_t port;"

I agree such a case is a good simple one to have, and I can add one.  For reasons described below, I will make it a mycompany.com private example, not a p4.org/psa/v1 one.

It may be likely to be the common case in a user's program, but it currently is _not_ the case in the psa.p4 include file.  Mihai Budiu was the one who recommended first creating the `typedef` definition of a target-specific bit width for a type like PortId_t, and then after that do the `type` definition for PortId_t based on that one, because it means that the bit width is only represented once in the include file, not duplicated in two places.  Unless we find a reason to change that in psa.p4, that pair of `typedef` followed by `type` based on the `typedef`, will be the common case in that include file.

docs/v1/guidance-for-generating-p4info.md
jafingerhut: It does.  I moved that text to the previous example, which has been renumbered example 3 to make room for a new simpler example 2, an example that is the simple one you suggested.

jafingerhut:
I believe all of your review comments have been addressed with the latest commit.  I did add a new example 2, and renumbered what was examples 2 thru 5 as 3 thru 6.

PR #198, opened 2019-04-19
Fix master arbitration

konne-google:


konne-google:
Please take a look: @jafingerhut @kuujo @stefanheule @wmohsin @ccascone

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -686,10 +686,11 @@ follows:
   for each pair of (`device_id`, `role_id`) at any point of time.
 
 * The streaming channel between the controller and the server defines the
-  liveness of the controller session. The controller is considered "offline" or
-  "dead" as soon as its corresponding stream channel to the switch is
-  broken, in which case the P4Runtime server quickly sets one of the slave
-  controllers with the highest `election_id` as master.
+  liveness of the controller session. The controller is considered “offline” or
+  “dead” as soon as its corresponding stream channel to the switch is broken, in
+  which case the P4Runtime server will be without a master, until one of the
```
jafingerhut: My only comment here is on the wording of "until one of the slave controllers".  The intended meaning is that any client that sends a successful master arbitration update message with a greater election_id than that of the previous master, yes?  As this is worded, it sounds like you may intend to restrict that to one of the _existing_ clients at the time the previous master died, but I don't think you intended any such restriction.

Perhaps something like the following?

Replace "until one of the slave controllers increases its election_id to a value higher than"

with: "until a client sends a successful `MasterArbitrationUpdate` message with an `election_id` value higher than"

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4391,6 +4388,10 @@ any of the devices known to the P4Runtime server, the server must return a
 `NOT_FOUND` error. The `entities` repeated field is a list of P4 entities, each
 acting as a query filter to be applied to P4 entity containers on the server.
 
+Since `ReadRequest`s do not mutate any state on the switch, they do not
+require an `election_id`, and they do not require the pressence of an open
```
jafingerhut: typo: "pressence" -> "presence"

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4705,8 +4710,18 @@ controller populates the `MasterArbitrationUpdate` field in this message using
 its `role_id` and `election_id` and the `device_id` of the device, as explained
 in detail in the [Master-Slave Arbitration and Controller
 Replication](#sec-master-slave-arbitration-and-controller-replication)
-section. For any given (`device_id`, `role_id`), the controller with the highest
-`election_id` is the master and the rest are slaves.
+section. For any given (device_id, role_id), the P4Runtime server keeps track
+of the highest election_id that it has ever received. If a controller's
+election_id is equal to the highest election_id that the server has ever
+received, that controller is the master. All other controller are slaves. Note
```
jafingerhut: Suggest replacing "All other controller are" with "All other controllers are"

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -686,10 +686,11 @@ follows:
...
 
 * The streaming channel between the controller and the server defines the
-  liveness of the controller session. The controller is considered "offline" or
-  "dead" as soon as its corresponding stream channel to the switch is
-  broken, in which case the P4Runtime server quickly sets one of the slave
-  controllers with the highest `election_id` as master.
+  liveness of the controller session. The controller is considered “offline” or
+  “dead” as soon as its corresponding stream channel to the switch is broken, in
+  which case the P4Runtime server will be without a master, until one of the
+  slave controllers increases its election_id to a value higher than anything
```
jafingerhut: Question that I am not sure about: Is it important that the new controller must have an election_id _strictly greater_ than any election_id value that the server has previously considered the master?

Or is it acceptable that it be _at least_ the previous master's election_id?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4705,8 +4710,18 @@ controller populates the `MasterArbitrationUpdate` field in this message using
 its `role_id` and `election_id` and the `device_id` of the device, as explained
 in detail in the [Master-Slave Arbitration and Controller
 Replication](#sec-master-slave-arbitration-and-controller-replication)
-section. For any given (`device_id`, `role_id`), the controller with the highest
-`election_id` is the master and the rest are slaves.
+section. For any given (device_id, role_id), the P4Runtime server keeps track
+of the highest election_id that it has ever received. If a controller's
+election_id is equal to the highest election_id that the server has ever
```
jafingerhut: Is there an assumption that no two clients will ever have the same election_id for the same (`device_id`, `role_id`) pair?

If so, is this a good place to state that explicitly?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -828,14 +829,10 @@ the server must return an `INVALID_ARGUMENT` error.
...
 
-"Mastership change" refers to either one of these cases:
-
-1. A new `MasterArbitrationUpdate` is received from an already connected
-   controller for a given (`device_id`, `role_id`), which changes the controller
-   mastership status (the controller becomes master or slave).
-
-2. A streaming channel for a given master controller breaks, forcing a new
-   master to be elected.
+"Mastership change" refers to the case where a new `MasterArbitrationUpdate` is
```
wmohsin: I feel we should disambiguate various cases that lead to mastership change:

1. Mastership is gained when a connected instance (other than the currently selected master) advertises a new higher election_id.
2. Mastership is updated when current master advertises a new highest election_id.
3. Mastership is lost when either (i) streaming channel for the current master breaks or (ii) current master advertises a new election_id lower than (but not used by) the election_ids advertised by other controller connections (for this role_id).

Open question: Should the protocol explicitly disallow 3(ii), i.e. require election_id to be monotonically increasing?

docs/v1/P4Runtime-Spec.mdk
ccascone:
```suggestion
  slave controllers increases its `election_id` to a value higher than anything
```

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4695,6 +4696,10 @@ message PacketMetadata {
 
 ## Master Arbitration Update
 
+P4Runtime's master arbitration mechanism ensures the following invariant:
+the device will not allow a write from an older master, if it has already
+accepted a write from a newer master.
+
```
ccascone: Should this be changed to "if it has already accepted a `MasterArbitrationUpdate` message from a newer master."?  Otherwise, it seems the mastership change is enforced only after the new master sends a `WriteRequest`, thus allowing an old master to modify state after a mastership change,  as defined in the rest the spec, has already occurred.

docs/v1/P4Runtime-Spec.mdk
ccascone: I vote for disallowing 3(ii)  as it can produce corner-cases, e.g. the switch now considering master a client that does not expect to be master

docs/v1/P4Runtime-Spec.mdk
konne-google: Done. I change "higher" to "greater or equal"

docs/v1/P4Runtime-Spec.mdk
konne-google: Done

docs/v1/P4Runtime-Spec.mdk
konne-google: Done.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: I thought "or equal" was perhaps maybe wrong, and a new master must always have a strictly greater electon_id than the old master?

docs/v1/P4Runtime-Spec.mdk
konne-google: It's important that the election_id can be equal. Consider the case where the master disconnects, and then reconnects immediately. It would still have the same election id, and should be elected master again.

docs/v1/P4Runtime-Spec.mdk
konne-google: Done.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: OK, that sounds like a detail good to include in any formal analysis done on this, as it sounds racy to me.

docs/v1/P4Runtime-Spec.mdk
konne-google: I think the "or equal" is important. I've commented on this issue in a later comment.

docs/v1/P4Runtime-Spec.mdk
konne-google: Done.

docs/v1/P4Runtime-Spec.mdk
konne-google: Great point, done!

docs/v1/P4Runtime-Spec.mdk
konne-google: It seems fine to allow 3 (ii). e.g. if a master wants to downgrade itself to read only.

konne-google: 3 is not a "mastership change", as no new master will be chosen.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -622,8 +622,9 @@ reasons:
    assigned to each role, is **out-of-scope** of this document.
 
 2. Redundancy and fault tolerance: Supporting multiple controllers allows having
-   one or more standby slave controllers, which take over controlling the
-   devices in case the master controller goes offline.
+   one or more standby slave controllers. These can already have a connection
+   open, and can read the switch state while being a slave, which can help them
+   become master more quickly.
```
jafingerhut: Is it actually true that a new master can become up to date with the current state more quickly given today's P4Runtime protocol, unless it has some way of knowing which of the data it read before it became master is still valid, and which is stale?

i.e. if P4Runtime had a per-table "content version id" number that was monotonically increasing whenever any of its contents were changed, and this value could be read by a client without reading all of the table state, and it came with a read of an entire table's state, then a new master could quickly read that value for all tables, compare it against what it has now, and then only read the contents of tables that it did not have the latest version of.

With today's protocol, it seems that no matter how many reads a slave may have performed before, it must assume that any or all of it could change by the time it becomes the master, because of writes issued by the previous master since the last time the slave did a read.

konne-google: I agree with this point. I'm not sure what we should do about it. It seems like a major change, but it might make sense to remove the notion of a standby controller.

stefanheule: Would it be sufficient to remove "Redundancy and fault tolerance" as a goal?  It really seems like we are falling short of that right now.

antoninbas: I remember a conversation with @wmohsin where he mentioned the potential cost of establishing a connection from a new controller to the switch (cost to discover a route between controller and switch). Maybe that alone justifies having standby controllers?
There is also what we discussed earlier today. Notifying standby controllers that the master has disconnected (assuming the master cannot itself notify the consensus layer for some reason) may enable faster recovery.

antoninbas: In any case, we should definitely remove "and can read the switch state while being a slave", but as I wrote above there may be some value in already having an open connection

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -686,10 +687,11 @@ follows:
   for each pair of (`device_id`, `role_id`) at any point of time.
 
 * The streaming channel between the controller and the server defines the
-  liveness of the controller session. The controller is considered "offline" or
-  "dead" as soon as its corresponding stream channel to the switch is
-  broken, in which case the P4Runtime server quickly sets one of the slave
-  controllers with the highest `election_id` as master.
+  liveness of the controller session. The controller is considered “offline” or
+  “dead” as soon as its corresponding stream channel to the switch is broken, in
```
antoninbas: could we use straight quotes for "offline" and "dead", just like in the rest of the doc?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -715,6 +717,9 @@ follows:
     * Non-OK (with `status.code` set to `google.rpc.ALREADY_EXISTS`) when the
       controller is determined to be a slave for a given (`device_id`,
       `role_id`).
+  An OK message is only sent after all previous write requests have been
```
antoninbas: Carmelo and Jordan have confirmed that it would be useful to them to have a google.rpc.NOT_FOUND` Non-OK status sent to all the stand-bys when the master disconnects, so I would recommend adding this as part of the PR given the very small implementation cost
It may also provide some justification / value for the stand-by connections...

jafingerhut:
I think this comment has not been addressed yet?  https://github.com/p4lang/p4runtime/pull/198#discussion_r280521092

jafingerhut:
From 2019-Jun-19 P4 API working group meeting, suggestions of some additions to make (unless they are already in the current PR and I have missed it somehow):

+ It should be implied by everything else already said, but perhaps explicitly mention that once a master becomes no longer the master, if any further write requests are received from the old master, they are rejected.

+ Perhaps more subtly, the new master should have a way to know when any WriteRequest messages already stored on the P4Runtime server are complete.  Some of these perhaps already have started, but more dangerously some perhaps may not have yet begun their processing.  According to the concurrency allowances for processing WriteRequest and ReadRequest messages, if the new master was allowed to send a ReadRequest or WriteRequest _before_ those old WriteRequest messages were complete, the new ReadRequest or WriteRequest message might pass up the not-yet-begun WriteRequest messages, and the not-yet-begun ones could modify the state after the new master had some operations performed.

One way to avoid that issue is to never notify a new master that it is the new master, until after any stored WriteRequest messages are complete in their processing.  Then there are no such messages for the new master's messages to pass up.

jafingerhut:
Following up on my previous comment https://github.com/p4lang/p4runtime/pull/198#issuecomment-503670018

A way to avoid the problem of WriteRequests from a new master never "passing up" WriteRequests from the old master is to specify that all WriteRequest messages must be processed in increasing order of election id.

That still does not prevent a new master's ReadRequest messages from passing up WriteRequest messages currently on the server from an old master, which is a problem for a client trying to get up to date with the latest state -- it could read stale state.  I do not see any way to prevent that locally on the server other than the one proposed in the comment above (delaying notifying a new master until after all existing WriteRequest messages have completed processing), _or_ somehow requiring that ReadRequest messages can never pass up WriteRequest messages received earlier by the server.

docs/v1/P4Runtime-Spec.mdk
antoninbas: @stefanheule could you take care if this when you update the PR?

docs/v1/P4Runtime-Spec.mdk
antoninbas: @stefanheule I think it would be good to address this as you update the PR

stefanheule:
> From 2019-Jun-19 P4 API working group meeting, suggestions of some additions to make (unless they are already in the current PR and I have missed it somehow):
> 
> * It should be implied by everything else already said, but perhaps explicitly mention that once a master becomes no longer the master, if any further write requests are received from the old master, they are rejected.
> * Perhaps more subtly, the new master should have a way to know when any WriteRequest messages already stored on the P4Runtime server are complete.  Some of these perhaps already have started, but more dangerously some perhaps may not have yet begun their processing.  According to the concurrency allowances for processing WriteRequest and ReadRequest messages, if the new master was allowed to send a ReadRequest or WriteRequest _before_ those old WriteRequest messages were complete, the new ReadRequest or WriteRequest message might pass up the not-yet-begun WriteRequest messages, and the not-yet-begun ones could modify the state after the new master had some operations performed.
> 
> One way to avoid that issue is to never notify a new master that it is the new master, until after any stored WriteRequest messages are complete in their processing. Then there are no such messages for the new master's messages to pass up.

Addressed both of these points in the updated PR.  For the second issue I chose to delay the mastership notification until all previous writes are done processing.  I believe it is the right thing to require fully sequential change to mastership (ie. everything old finishes, and only then the new master actually becomes master).  This is much simpler, and I don't believe we would need the increased performance of doing some things in parallel, as mastership changes are not happening frequently.

stefanheule:
I have addressed all the comments we had both here and from the last working group meeting.  Unfortunately the previous text was confusing, and didn't address a number of cases, so I tried to improve it a bunch. Here is a (hopefully complete) list of changes I made:
- Explicitly require a election_id on MasterArbitrationUpdate messages.
- Added notifications when a master loses its connection.
- Delay notifications when there are writes from the old master still in flight.
- Properly document how one can change the `role.config` (it was rather vague before), and send a advisory notification when it gets updated.
- Remove the ability to change roles. That seems rather tricky to get right (the current specification is certainly not precise enough to capture this), and I don't believe anybody is using this at the moment.  If a controller would like to change role, they should tear down the connection and start a new one (just like with the device).

Please take a look and let me know what you think.  Since the chances are significant and some key people are OOO next week, I don't expect this to merge soon.

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk
stefanheule: 3i and ii are both allowed and explicitly documented in the updated PR.  @ccascone had concerns that the switch might start considering a client to be master that doesn't expect to be master, but that can't happen.  If a master downgrades (or loses connection), then the switch will be without master.  This is explicitly documented in the spec now.  Let me know if there are still concerns.

chrispsommers:
"Delay notifications when there are writes from the old master still in flight."

Do error conditions on those writes affect this? Should we say this is the way it's handled regardless of the outcome of such in-flight writes? Maybe it's obvious.

stefanheule:
> "Delay notifications when there are writes from the old master still in flight."
> 
> Do error conditions on those writes affect this? Should we say this is the way it's handled regardless of the outcome of such in-flight writes? Maybe it's obvious.

I think it would probably be clear that errors won't affect the rest of the scheme, but it's good to explicit.  I added a sentence about this (and also made one other thing about in-flight write requests more clear).

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4755,8 +4835,19 @@ controller populates the `MasterArbitrationUpdate` field in this message using
...
-`election_id` is the master and the rest are slaves.
+section. For any given `(device_id, role_id)`, the P4Runtime server keeps track
+of the highest `election_id` that it has ever received. If a controller's
+`election_id` is equal to the highest `election_id` that the server has ever
+received, that controller is the master. All other controllers are slaves. Note
+that it is possible that all controllers are slaves, and that there is no
+master. There can be at most one master, because for any given
+`(device_id, role_id)`, each connected controller has a unique `election_id`.
+
+To ensure that the invariant holds across reboots, the device needs to persist
```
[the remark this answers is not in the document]
stefanheule: Yeah this was not right, if it is persisted across restarts then there is no way to reset the election ID. I clarified that all switches have "full restarts" (where the forwarding state is cleared etc), and then the election_id is also reset (and NOT persisted).  Some targets may also support in-service software upgrade (ISSU), ie. restart while forwarding traffic.  In that case, the election_id must be persisted.

Does that satisfy your "nice-to-have" wish, since ISSU is optional?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4797,6 +4888,10 @@ also populates the `status` field in the `MasterArbitrationUpdate` as follows:
 * Non-OK (with `status.code` set to `google.rpc.ALREADY_EXISTS`) when the
   controller is determined to be a slave for a given (`device_id`, `role_id`).
 
+The sender need not specify an `election_id`. If the `election_id` is not
```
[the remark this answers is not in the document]
stefanheule: Done

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -769,10 +759,13 @@ are out-of-scope of this document. When partitioning of the control plane is
 desired, the P4Runtime client(s) and server need to agree on a role definition
 scheme in an out-of-band fashion.
 
-## Rules for Handling `MasterArbitrationUpdate` Messages Received from Controllers
+It is the job of the P4Runtime server to remember the `role.config` for every
```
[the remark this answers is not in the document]
stefanheule: Yes, that is what we are saying.  We were already saying this before, just much less explicitly, but I think it was the only way to interpret v1 of our spec.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -782,67 +775,146 @@ scheme in an out-of-band fashion.
...
+           the [following section](#sec-mastership-notification).
+
+        3. The server will finish processing any `Write` requests that have
+           already started. If there are errors, they are reported as usual to
+           the previous master. If the previous master has already disconnected,
+           any possible errors are dropped and not reported.
+
+        4. The server now accepts the current controller as the new master, thus
+           accepting `Write` requests from this controller. The server updates
+           the highest election ID (i.e., `election_id_past`) it has seen for
```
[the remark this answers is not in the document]
stefanheule: Fixed.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -769,10 +759,13 @@ are out-of-scope of this document. When partitioning of the control plane is
...
-## Rules for Handling `MasterArbitrationUpdate` Messages Received from Controllers
+It is the job of the P4Runtime server to remember the `role.config` for every
+`device_id` and `role_id` pair.
+
+## Rules for Handling `MasterArbitrationUpdate` Messages Received from Controllers { #sec-mastership-updates }
 
-1. If the `MasterArbitrationUpdate` message is received for the first time (for
-   a newly connected controller):
+1. If the `MasterArbitrationUpdate` message is received for the first time on
+   this particular channel (i.e., for a newly connected controller):
```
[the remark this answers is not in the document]
stefanheule: Fixed.  Thanks for the review, since I added some text for the restarts, please take another look.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4745,6 +4830,10 @@ message PacketMetadata {
 
 ## Master Arbitration Update
 
+P4Runtime's master arbitration mechanism ensures the following invariant: the
```
konne-google: This doesn't make a lot of sense. Can we replace it with something like this:

P4Runtime's master arbitration mechanism ensures that only the current master can modify state on the switch, and that the election id is monotonically increasing. For example, the switch must finish all previous write operations before changing masters, and must only accept write requests from the current master.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -687,9 +697,13 @@ follows:
 
 * The streaming channel between the controller and the server defines the
   liveness of the controller session. The controller is considered "offline" or
-  "dead" as soon as its corresponding stream channel to the switch is
-  broken, in which case the P4Runtime server quickly sets one of the slave
-  controllers with the highest `election_id` as master.
+  "dead" as soon as its corresponding stream channel to the switch is broken, in
+  which case the P4Runtime server will be without a master, until a client sends
```
wmohsin: nit: "in which case the P4Runtime server will be without a master"...
This will only be true if the master channel is broken -- the preceding sentence fragment doesn't clarify that.
How about:
The controller is considered "offline" or "dead" as soon as its stream channel to the switch is broken. When a master channel gets broken: (i) an advisory message is sent to all other controllers for that `device_id` and `role_id`, as described in the [following section](#sec-mastership-notification) (ii) the P4Runtime server will be without a master, until a client sends a successful `MasterArbitrationUpdate` (as per the rules in a  [later section](#sec-mastership-updates)).

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -511,6 +511,15 @@ In this situation, P4Info is selectively packaged into role-based subsets to
 allow some controllers access to just the functionality required. For example, a
 controller may only need read access to statistics counters and nothing more.
 
+## P4Runtime State Across Restarts { #sec-restarts }
+
+All targets support full restarts, where all forwarding state is reset and the
+P4Runtime server starts with a clean state.  Some targets may also support
+In-Service Software Upgrade (ISSU), where the software on the target can be
+restarted while traffic is being forwarded. In this case, the P4Runtime Server
```
wmohsin: This may sounds a little too restrictive. Perhaps something like: "In this case, the P4Runtime server may have the ability to access information from memory before the upgrade".

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -658,17 +669,16 @@ follows:
   different (`device_id`, `role_id`) pairs; it allows it.
 
 * To start a controller session, a controller first opens a bidirectional stream
```
wmohsin: I feel we agreed that is also needed if the controller wishes to mutate state (via one of the P4RT RPCs) down the road, and a stream-channel is not required for just reading state. Could you please disambiguate this?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4797,6 +4898,16 @@ also populates the `status` field in the `MasterArbitrationUpdate` as follows:
 * Non-OK (with `status.code` set to `google.rpc.ALREADY_EXISTS`) when the
   controller is determined to be a slave for a given (`device_id`, `role_id`).
 
+The sender need not specify an `election_id`. If the `election_id` is not
+specified, the sender's `election_id` is considered lower than any
+`election_id`, and the sender will thus never become master. This way, if a
+controller is informed by the consensus layer that it is not supposed to be
```
wmohsin: 1. Does this section better belong up with the section that describes the details of master-arbitration between controller and P4RT-server?
2. I suggest we omit "consensus layer" (we haven't defined it) and simply say something like: "This way, if a controller that knows it is supposed to be the master may choose not to ..."

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -782,67 +784,146 @@ scheme in an out-of-band fashion.
        (`device_id`, `role_id`), the P4Runtime server shall terminate the stream
        by returning an `INVALID_ARGUMENT` error.
 
-    3. If the max number of clients for the given (`device_id`, `role_id`)
+    3. If `role.config` does not match the "out-of-band" scheme previously
+       agreed upon, the server must return an `INVALID_ARGUMENT` error.
+
+    4. If the max number of clients for the given (`device_id`, `role_id`)
```
ccascone: I think it's more correct to talk about streams here, as there can be clients without an open stream, for example when performing a Read RPC.
```suggestion
    4. If the number of open streams for the given (`device_id`, `role_id`) 
```

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -782,67 +784,146 @@ scheme in an out-of-band fashion.
...
+           the advisory message for every update, even if the master sets the
+           same `role.config` as it has before. See the
+           [following section](#sec-mastership-notification) for the format of
+           the advisory message.
+
+        2. If the controller is a slave, this is a no-op and the `role.config`
+           is ignored. No response is sent to any controller.
+
+    6. Otherwise, the server updates the `election_id` it has stored for this
+       controllers. This change might cause a change in mastership (this
```
ccascone:
```suggestion
       controller. This change might cause a change in mastership (this
```

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -782,67 +784,146 @@ scheme in an out-of-band fashion.
...
+
+If the `MasterArbitrationUpdate` is accepted by either of the two steps above
+(cases 1.5. and 2.6. above), then the server determines if there are changes in
+mastership. Let `election_id_past` be the highest election ID the server has
+ever seen for the given `device_id` and `role_id` (including the one of the
+current master if there is one).
+
+1. If `election_id` is greater than or equal to `election_id_past`, then the
+   controller becomes master. The server updates the role configuration to
+   `role.config` for the given `role.id`.  Furthermore:
```
ccascone:
```suggestion
   `role.config` for the given `role.id`. Furthermore:
```

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -782,67 +784,146 @@ scheme in an out-of-band fashion.
...
+
+        4. The server now accepts the current controller as the new master, thus
+           accepting `Write` requests from this controller. The server updates
+           the highest election ID (&ie; `election_id_past`) it has seen for
+           this `device_id` and `role_id` to `election_id`.
+
+        5. The server notifies the new master by sending the advisory message
+           described in the [following section](#sec-mastership-notification).
+
+2. Otherwise, the controller becomes a slave: If the controller was previously a
```
ccascone:
```suggestion
2. Otherwise, the controller becomes a slave. If the controller was previously a
```

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -782,67 +784,146 @@ scheme in an out-of-band fashion.
...
+`role.config` is updated by the master, all controllers for that
+(`device_id`, `role_id`) are informed of this by sending a
+`StreamMessageResponse`. The `MasterArbitrationUpdate` is populated as follows:
+
+* `device_id` and `role.id` as given.
+
+* `role.config` is set to the role configuration the server received most
+  recently in a `MasterArbitrationUpdate` from a master.
+
+* If there currently is a master, then `election_id` is set to the `election_id`
```
ccascone: I believe this point could be simplified by removing the case "If there currently is a master...". Indeed, election_id should always be set to the highest election ID that the server has seen, which includes the case where a master exists.

Also, I would enforce the case when there has not been any master at all by removing the parentheses. Maybe you can create a new bullet point:

If there has not been any master at all, the `election_id` is left unset.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4797,6 +4898,16 @@ also populates the `status` field in the `MasterArbitrationUpdate` as follows:
 * Non-OK (with `status.code` set to `google.rpc.ALREADY_EXISTS`) when the
   controller is determined to be a slave for a given (`device_id`, `role_id`).
 
+The sender need not specify an `election_id`. If the `election_id` is not
+specified, the sender's `election_id` is considered lower than any
+`election_id`, and the sender will thus never become master. This way, if a
+controller is informed by the consensus layer that it is not supposed to be
+master, it may choose not to advertise an election_id, in order to avoid master
```
ccascone:
```suggestion
master, it may choose not to advertise an `election_id`, in order to avoid master
```

docs/v1/P4Runtime-Spec.mdk
stefanheule: 1. After reading over the document again, this actually seems like a good place.  Let me know if you disagree.
2. Done.

Thanks for the review!

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -622,8 +631,10 @@ reasons:
    assigned to each role, is **out-of-scope** of this document.
 
 2. Redundancy and fault tolerance: Supporting multiple controllers allows having
-   one or more standby slave controllers, which take over controlling the
-   devices in case the master controller goes offline.
+   one or more standby slave controllers. These can already have a connection
+   open, which can help them become master more quickly, especially in the case
+   where the management traffic is in-band and connection setup might be more
```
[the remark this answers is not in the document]

docs/v1/P4Runtime-Spec.mdk
stefanheule: Fixed by another commit.

docs/v1/P4Runtime-Spec.mdk

Thank you for your review, I applied all changes.

stefanheule:
Thank you everyone, I have addressed all comments, but feel free to add more if there is more feedback.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -511,6 +511,14 @@ In this situation, P4Info is selectively packaged into role-based subsets to
 allow some controllers access to just the functionality required. For example, a
 controller may only need read access to statistics counters and nothing more.
 
+## P4Runtime State Across Restarts { #sec-restarts }
+
+All targets support full restarts, where all forwarding state is reset and the
+P4Runtime server starts with a clean state.  Some targets may also support
+In-Service Software Upgrade (ISSU), where the software on the target can be
+restarted while traffic is being forwarded. In this case, the P4Runtime server
+may have the ability to access information from memory before the upgrade.
```
antoninbas: sorry for the late comment, but "information from memory before the upgrade" sounds very vague to me. Shouldn't we clarify? Do you mean forwarding state, server state (such as the highest election id), or both?

stefanheule: Waqar wanted to keep this relatively open, so that we don't restrict what exactly a server does.  The only requirement we have is the election ID needs to be preserved across ISSU, which we state clearly in the later section:

> This invariant must be maintained across in-service software upgrades, and the
> P4Runtime server must remember the highest `election_id` after such a restart.
> However, across a [full restart](#sec-restarts), the `election_id` must be
> reset. In fact, a full restart is the only way to reset the `election_id`.

Does that make sense?  I didn't want to duplicate the requirements in multiple places, and the later section seemed most appropriate for the actual requirement, but it didn't seem appropriate to introduce the notion of ISSU there.

antoninbas: I don't have a strong objection. We can clarify later if needed.

PR #207, opened 2019-06-17
Add note about support of values of type `bool` for table key fields …

jafingerhut:
…etc.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -5249,12 +5249,18 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
 `p4.v1.TableEntry` message based on the value of architecture-specific table
 properties, but we may include on in future versions of the API.
 
-# Known-limitations of P4Runtime v1.0.0
+# Known limitations of P4Runtime v1.0.0
 
 * `FieldMatch`, action `Param`, and controller packet metadata fields only
   support unsigned bitstrings, i.e. values of one of the following types (not
   the more general `P4Data`):
   * `bit<W>`
+  * `bool`.  Note that as far as the `P4Info` message contents and
+    thus controller software is concerned, such fields of type `bool`
+    will be indistinguishable from those that have been declared with
+    type `bit<1>`.  P4Runtime server software will automatically
+    perform any conversion needed between the type `bit<1>` values in
+    P4Runtime messages and the data plane representation.
   * an `enum` with underlying type `bit<W>`
   * a `type` or `typedef` with an underlying type that is one of the above (or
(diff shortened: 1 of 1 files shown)


jafingerhut:
@antoninbas mentioned adding something to the P4Runtime spec similar to this.  Reference: https://github.com/p4lang/p4c/pull/1969#issuecomment-502873598

It needs review, of course, to ensure I have captured the idea correctly.

jafingerhut:
Just a slight warning about possible future consequences of this proposed change:

If the goal in some future P4Runtime version is to enable table search key fields, action parameters, etc. to have more general P4_16 types than the few listed in this section now, e.g. to use P4Data in the way that digest parameters do, then defining `bool` to auto-convert to `bit<1>` like this today could mean that we need to preserve that behavior in the future.  For example, we might have to keep an exception that says "if you want a bool type for action parameters, table search key fields, etc., you must _not_ use P4Data for that, but instead treat it the same as if it were a `bit<1>` type in the P4 program".

antoninbas:
@jafingerhut moving to `P4Data` for match fields and action parameters will be a non backward-compatible change anyway, and that won't happen for a while. Until then boolean value will be treated as `bit<1>` wherever `P4Data` is not used. In places where `P4Data` is already used (digests, registers), the `P4Data` oneof already has a case for boolean values (https://github.com/p4lang/p4runtime/blob/master/proto/p4/v1/p4data.proto#L23).
If, and when, we move to `P4Data` for match fields and action parameters in a future P4Runtime version, then the spec will be updated to say that boolean values are no longer treated as `bit<1>` values. I originally thought that maybe we will have an intermediate version of P4Runtime, which would be backward-compatible, and let you use either `P4Data` or `bytes` for match fields and action parameters. We will see if it is ever needed. So far I'm not sure there has been a use case which requires `P4Data` support for match fields and action parameters. It would definitely be nice to have it and would make the interface more uniform. But it would also make the implementation more complex, and so far there has not been a compelling reason to add this support (either in the form of a backward-compatible version using a `oneof`, or as a non backward-compatible P4Runtime v2.0 release).

jafingerhut:
Regarding this part of the previous comment: "So far I'm not sure there has been a use case which requires P4Data support for match fields and action parameters."

There is the admittedly probably not terribly common scenario of wanting to match on values of type `error`, e.g. in order to convert them to values of type `bit<W>` so they can be put into headers, e.g. to-CPU headers: https://github.com/p4lang/p4-spec/blob/master/p4-16/psa/examples/psa-example-parser-error-handling.p4#L361-L382

An alternate way to handle that use case is to change the P4_16 language spec so that `error` is changed to a serializable enum, perhaps with a target-specific bit width.

PR #208, opened 2019-06-18
Document that p4runtime_translation annotations need only be supported

### LINKED ISSUE #195: Complete list of places where we expect to support p4runtime_translation annotation?
I was hoping to ensure that there is some mention somewhere of the complete list of places in a P4_16 program where a `p4runtime_translation` annotation is significant, i.e. actually used by the implementation.

Certainly the primary use case is for a `type` definition.

Would it be worth specifying that such an annotation is explicitly ignored by P4Runtime v1.0 if it appears anywhere else in a P4_16 program?  For example, someone might be tempted to try putting such an annotation on a `typedef` definition, or on a particular table search key field, or action parameter.  Making it clear that either these are definitely supported, and affect the P4Info message created, or explicitly documenting that such annotations are ignored, seems like a good idea to me.

I do not currently have any strong preferences along the question of "we should support those other places" or "we should explicitly _not_ support them".  I have at least a mild preference for not supporting it anywhere other than a `type` definition, for slightly simplifying the implementation.

jafingerhut:
... on type declarations in a P4 program.

Fixes #195

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2287,6 +2287,11 @@ server and the *bitwidth* of the bitstring type exposed to the control plane. It
 is recommended that the URI includes at least the P4 architecture name and the
 type name.
 
+A `@p4runtime_translation` annotation need not have any effect if it is used to
+annotate anything in a P4 program other than a `type` declaration. It is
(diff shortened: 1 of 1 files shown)


jafingerhut:
Intended to address this issue https://github.com/p4lang/p4runtime/issues/195

Issue #199, opened 2019-04-25
Explicitly document that servers must never concurrently perform ReadRequest and WriteRequest?


jafingerhut:
**This is for reporting issues with the specification only. If you are running into a bug with a specific `P4Runtime` implementation, please open an issue with the appropriate repository (e.g. https://github.com/p4lang/behavioral-model or https://github.com/p4lang/PI for bmv2).**

Section 12.1 explicitly allows a server to implement multiple WriteRequest operations concurrently.

Section 13.3.1 seems to imply to me that a server is allowed to implement multiple ReadRequest operations concurrently.  Corrections welcome if I am misinterpreting that.

In a recent P4 API WG meeting, it was mentioned that reads cannot happen concurrently.  If so, updating Section 13.3.1 seems like a good idea.  However, I can't think of any good reason to disallow a server from processing multiple concurrent ReadRequest operations.

It was also mentioned that a server is not allowed to concurrently execute a WriteRequest and a ReadRequest.  If that is in the spec already, I would appreciate someone pointing it out to me.  If it isn't there already, it sounds like a good addition.

antoninbas:
For the multiple concurrent Reads, the confusion comes from me. A server is allowed to process multiple Reads concurrently. The minutes for the meeting were incorrect, so I updated them.

Section 13.3 describes how a server implementation must process a ReadRequest: https://s3-us-west-2.amazonaws.com/p4runtime/docs/v1.0.0/P4Runtime-Spec.html#sec-batch-processing. The first step is "Lock state (preventing new writes)".

antoninbas:
Note that the p4lang/PI implementation currently serializes ReadRequests (while WriteRequests are processed concurrently). It's just a shortcut I took.

jafingerhut:
Thanks for clarifications and pointer to existing text in the spec.

So "Lock state (preventing new writes)" also means "and lock will not succeed until any unfinished writes are complete"?

antoninbas:
That's how I understand it. And I believe that is definitely what Waqar meant.
The read will be delayed until all ongoing writes complete, and new writes should not preempt the read but be delayed until the read completes.
Feel free to submit a clarification if you think the document would benefit from it.

jafingerhut:
So is the required observable behavior something like the following?

(REQ1) Every `ReadRequest` message must return results as if all other `WriteRequest` messages were completely finished, or not yet started?

[...]

PR #209, opened 2019-06-18
Text describing the parallelism allowed in a P4Runtime server

jafingerhut:
... when serving ReadRequest and WriteRequest messages.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -4508,11 +4508,11 @@ For example, imagine a client that wanted to use `WriteRequest`
 messages with large batches of insert, modify, and/or delete
 operations on an IP route table, in order to achieve higher throughput
 of updates to this table.  Such a client might also wish to send
-`WriteRequest` messages with small numbers of updates to an
-`ActionSelector` object that controlled which links were in which
-LAGs, and have those small requests start processing even if there is
-a large `WriteRequest` batch currently being processed, but it will
-not complete for a significant amount of time.
+`WriteRequest` messages with only a few updates to an `ActionSelector`
+object that controlled which links were in which LAGs, and have those
+small requests start processing even if there is a large
+`WriteRequest` batch currently being processed, but it will not
+complete for a significant amount of time.
 
 The restrictions on which a client may rely are:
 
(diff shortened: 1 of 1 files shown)


jafingerhut:
This new proposed text is intended to address this issue: https://github.com/p4lang/p4runtime/issues/199

I was not sure of the best place to add such text.  Right now this PR has it added in a new subsection at the end of the "Read RPC" section.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4496,6 +4496,51 @@ multi-reader lock (also known as multiple-readers/single-writer lock).
...
+  processing, or `W` has not yet begun processing.
+
+For example, if a P4Runtime server maintained, independently for each
+device it managed, a separate multi-reader single-writer lock for each
+stateful object in the P4 program, and before starting the processing
+of a `WriteRequest` message it acquired a write lock for each stateful
+object affected by the `WriteRequest`, and before starting the
+processing of a `ReadRequest` message it acquired a read lock for each
+stateful object accessed by the `ReadRequest`, such an implementation
+should meet all of the requirements above.
```
stefanheule: Why does it say "should meet" instead of "meets"?  Is there anything else to be considered?

jafingerhut: The only reason it says that is my often-cautious way of writing when I'm not yet sure of something.  Changing it to "meets" sounds like a good idea.  We can correct it later if I messed it up :)

stefanheule: No worries, thank for updating and writing all of this in the first place.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4496,7 +4496,7 @@ multi-reader lock (also known as multiple-readers/single-writer lock).
 Conversely (&eg; in a single-threaded architecture), it may choose to serialize
 `Read` RPC processing.
 
-## Parallelism of read and write requests
+## Parallelism of Read and Write requests
```
antoninbas: Actually that's not specific to "read" and "write", we capitalize every significant word in titles: https://github.com/p4lang/p4runtime/tree/master/docs#what-to-capitalize

jafingerhut: Doh!  If I had read your comment more carefully, I would have noticed that.  Fixed now.

PR #217, opened 2019-07-30
Add error field to StreamMessageResponse oneof

### LINKED ISSUE #215: What happens if the server receives a PacketOut message from a non-master controller?
How should the server react to such a situation? The spec just says that the controller has to be a master:

> In fact, before a controller becomes able to do Packet I/O or program any forwarding entry (via Write RPC), it needs to start a controller session and become a “master”.

I see the following possibilities, apart from ignoring the message, of course:

- Do nothing (Stratum current state)
- Log an error on the server
- Close the RPC channel with PERMISSION_DENIED (@ccascone's and my preference)

I prefer option 3 because it would match the behavior for non-master Write RPC calls:

> If the client is not the master for (device_id, role_id) according to the election_id value, the server must return a PERMISSION_DENIED error. 

It would also enforce stricter error checking and encourage defensive coding of controllers. A correctly implemented controller _knows_ that it can't send packets if it's not master. A faulty one will lose its connection, which will hopefully be noticed and corrected by the operator. Silently ignoring a permission violation seems not wise to me.

antoninbas:
This field may be used by the server to report errors generated by
invalid / unauthorized StreamMessageRequests.

Fixes #215

### DIFF
--- a/proto/p4/v1/p4runtime.proto
+++ b/proto/p4/v1/p4runtime.proto
@@ -137,7 +137,7 @@ message ExternEntry {
   // range [0x81, 0xfe].
   uint32 extern_type_id = 1;
   uint32 extern_id = 2;  // id of the instance
-  google.protobuf.Any entry = 3;
+  .google.protobuf.Any entry = 3;
 }
 
 message TableEntry {
@@ -227,7 +227,7 @@ message FieldMatch {
     Range range = 6;
     // Architecture-specific match value; it corresponds to the other_match_type
     // in the P4Info MatchField message.
-    google.protobuf.Any other = 100;
+    .google.protobuf.Any other = 100;
   }
 }
 
@@ -472,7 +472,7 @@ message StreamMessageRequest {
     MasterArbitrationUpdate arbitration = 1;
     PacketOut packet = 2;
     DigestListAck digest_ack = 3;
-    google.protobuf.Any other = 4;
+    .google.protobuf.Any other = 4;
   }
 }
 
@@ -500,7 +500,10 @@ message StreamMessageResponse {
     PacketIn packet = 2;
     DigestList digest = 3;
     IdleTimeoutNotification idle_timeout_notification = 4;
-    google.protobuf.Any other = 5;
+    .google.protobuf.Any other = 5;
+    // Used by the server to asynchronously report errors which occur when
+    // processing StreamMessageRequest messages.
+    StreamError error = 6;
   }
 }
 
@@ -560,7 +563,7 @@ message Role {
   // (default case), it implies all P4 objects and control behaviors are in
   // scope, i.e. full pipeline access. The format of this message is
   // out-of-scope of P4Runtime.
-  google.protobuf.Any config = 2;
+  .google.protobuf.Any config = 2;
 }
 
 message IdleTimeoutNotification {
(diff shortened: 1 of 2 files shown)


proto/p4/v1/p4runtime.proto
```diff
@@ -500,7 +500,10 @@ message StreamMessageResponse {
     PacketIn packet = 2;
     DigestList digest = 3;
     IdleTimeoutNotification idle_timeout_notification = 4;
-    google.protobuf.Any other = 5;
+    .google.protobuf.Any other = 5;
+    // Used by the server to asynchronously report errors which occur when
+    // processing StreamMessageRequest messages.
+    .google.rpc.Status error_status = 6;
   }
```
stefanheule: I think we should have a separate message, maybe called "PacketOutError" that contains a gooogle.rpc.Status, instead of directly having the status here.  I think that would be much clearer.  Right now the comment/message reads as if this error can be used for any error wrt. StreamMessageRequest, but that is not the case.  E.g., MasterArbitrationUpdate have their own error handling.

antoninbas: My assumption was that it could be used for other types of notification (e.g. `DigestListAck`), and I gave an example in the spec, but I agree that there is some redundancy for master arbitration.

stefanheule:
I think there are at least three options for this (as Antonin said in his email):

1. Drop this PR.
2. Accept as is.
3. Add a new field called packet_out_error of type PacketOutError to StreamMessageResponse (not a oneof though, as Antonin said, maybe that would be a forth proposal).  This error will only be used for PacketOut and nothing else, and PacketOutError for now only contains a Status.

I'm okay with 1 and 3, I think 2 would not be good:

- If we want to add any additional information to the error for packet out, we can only do this in an unstructured way (details of the Status).
- The proposal says this will also be used for other errors in the future potentially, in which case we have no way of distinguishing packet-out errors from those future errors (other than in an unstructured way in details).
- We can't extend the packet-out error with more information in the future in a clean way.  If there is a PacketOutError message, however, we can just add more fields if that becomes desirable, in a backwards compatible way.
- This proposal limits how any future error reporting for other features of the stream channel looks like.  In particular, if another kind of error reporting wants to always include more information, we are again bound to only doing that in an unstructured way in the details field.

One objection to 3 (over 2) might be added complexity because of the one additional message type.  I would actually argue 3 is much less complex overall, because it clearly specifies error handling for packet-out errors and nothing else. 2 on the other hand will limit any future error reporting we might want.  Thinking about all possible futures is really hard.

We should definitely also wait and see what @pudelkoM and Carmelo think.

antoninbas:
@ccascone @pudelkoM
Could we have your input on this, since the WG is unsure which way to go?
Please refer to https://docs.google.com/document/d/1qzeAnKDDOIgBzq3sLS1C1CacunTtrV5x5T5xPXUFRGM/edit#heading=h.7b84013o7tmu for some context on the discussion.

ccascone:
Sorry for the late reply, I was on vacation.

Based on our original concern, i.e., provide ways to the client to detect that a packet-out message was rejected, this PR works as-is. Indeed, I agree that stream error reporting should be provided for debugging purposes only. For this reason, I don't think we would need to carry any other additional information besides an error code and message (already included in `.google.rpc.Status`), not even the original packet-out message that caused the error, as packet-out messages should be considered best-effort/ephemeral (the resulting data plane packet could be dropped along the way for many other reasons).

However, I understand @stefanheule concerns. Perhaps defining a field for each possible error in the `StreamMessageResponse`'s `oneof` would add too much complexity, and there could be cases where we want just to convey an error code and message. For this reason, I would define a generic `StreamErrorStatus` message that looks like `.google.rpc.Status` but provides a more structured way to convey additional information. For example that looks like this:

```proto
message StreamMessageResponse {
  oneof update {
    MasterArbitrationUpdate arbitration = 1;
    PacketIn packet = 2;
    DigestList digest = 3;
    IdleTimeoutNotification idle_timeout_notification = 4;
    .google.protobuf.Any other = 5;
    StreamErrorStatus error_status = 6;
  }
}

message StreamErrorStatus {
  // The status code, which should be an enum value of
  // google.rpc.Code. Expected to be non-OK status code.
  int32 code = 1;
  // Optional. An explanation of the error.
  string message = 2;
  // Optional. Used by the server to convey additional
  // information about the error.
  oneof details {
  	PacketOutError packet_out = 1;
  	DigestListAckError digest_list_ack = 2;
  	StreamOtherError other = 3;
  	// etc..
  }
}

message PacketOutError {
	// The packet out message that caused the error.
	PacketOut packet_out = 1;
}

message DigestListAckError {
	// The packet out message that caused the error.
	DigestListAck digest_list_ack = 1;
}

message StreamOtherError {
	// The "other" message that caused the error.
	google.protobuf.Any other = 1;
}
```

PR #226, opened 2019-08-02
Support wildcard reads for multicast groups and clone sessions

antoninbas:
By reserving a special id value (0xffffffff) which is assigned to a enum
member in p4runtime.proto.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -3715,6 +3715,22 @@ semantics.
   operation. Any packets with their `multicast_group` metadata in the data plane
   set to the deleted `multicast_group_id` will be dropped.
 
+When reading a multicast group, only `multicast_group_id` is considered. All
+other fields in `MulticastGroupEntry`` are ignored. To perform a *wildcard* read
+on all configured multicast group entries, the client must set the
+`multicast_group_id` field to `MULTICAST_GROUP_ID_WILDCARD`, which is defined in
+p4runtime.proto as an enum member with the value `0xffffffff`:
+
+~ Begin Proto
+enum MulticastGroupIdReserved {
+  // 0xffffffff
+  MULTICAST_GROUP_ID_WILDCARD = -1;
+}
+~ End Proto
+
+This reserved value can never appear in a `Write` RPC. If it does, the server
+must return an `INVALID_ARGUMENT` error.
+
 ### `CloneSessionEntry`
 
 PSA supports cloning of packets in both the ingress and egress pipeline. Ingress
@@ -3808,6 +3824,22 @@ semantics:
   `clone_session_id`. Other fields need not be provided for this operation. Any
   packet with their `clone_session_id` metadata in the data plane set to the
   deleted `session_id` will no longer be cloned.
+
+When reading a clone session, only `session_id` is considered. All other fields
+in `CloneSessionEntry` are ignored. To perform a *wildcard* read on all
+configured clone session entries, the client must set the `session_id` field to
+`CLONE_SESSION_ID_WILDCARD`, which is defined in p4runtime.proto as an enum
+member with the value `0xffffffff`:
+
+~ Begin Proto
+enum CloneSessionIdReserved {
+  // 0xffffffff
+  CLONE_SESSION_ID_WILDCARD = -1;
+}
+~ End Proto
+
+This reserved value can never appear in a `Write` RPC. If it does, the server
+must return an `INVALID_ARGUMENT` error.
 
 ## `ValueSetEntry`
 
--- a/proto/p4/v1/p4runtime.proto
+++ b/proto/p4/v1/p4runtime.proto
@@ -394,8 +394,17 @@ message Replica {
 // each replica's egress input metadata will be set to the respective values
 // programmed in the multicast group entry.
 message MulticastGroupEntry {
+  // Set multicast_group_id to MULTICAST_GROUP_ID_WILDCARD for wildcard reads
   uint32 multicast_group_id = 1;
   repeated Replica replicas = 2;
+
+  // We define a special reserved value for the multicast_group_id field. This
+  // value can be used to do a wildcard Read on all configured multicast
+  // groups. This value cannot be used in Write RPCs.
+  enum MulticastGroupIdReserved {
+    // 0xffffffff
+    MULTICAST_GROUP_ID_WILDCARD = -1;
+  }
 }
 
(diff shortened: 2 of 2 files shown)


proto/p4/v1/p4runtime.proto
```diff
@@ -394,8 +394,19 @@ message Replica {
...
+  // Set multicast_group_id to MULTICAST_GROUP_ID_WILDCARD for wildcard reads
   uint32 multicast_group_id = 1;
   repeated Replica replicas = 2;
+
+  // We define a special reserved value for the multicast_group_id field. This
+  // value can be used to do a wildcard Read on all configured multicast
+  // groups. This value cannot be used in Write RPCs.
+  enum MulticastGroupIdReserved {
+    MULTICAST_GROUP_ID_INVALID = 0;
+    MULTICAST_GROUP_ID_FIRST = 1;
```
stefanheule: What do we need `MULTICAST_GROUP_ID_FIRST` for?  In particular it doesn't seem to be reserved, unlike the name suggests.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3715,6 +3715,22 @@ semantics.
...
+p4runtime.proto as an enum member with the value `0xffffffff`:
+
+~ Begin Proto
+enum MulticastGroupIdReserved {
+  // 0xffffffff
+  MULTICAST_GROUP_ID_WILDCARD = -1;
+}
+~ End Proto
+
+This reserved value can never appear in a `Write` RPC. If it does, the server
```
stefanheule: I can not appear as a multicast_group_id only, or it can't appear at all?  Same below.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3715,6 +3715,22 @@ semantics.
   operation. Any packets with their `multicast_group` metadata in the data plane
   set to the deleted `multicast_group_id` will be dropped.
 
+When reading a multicast group, only `multicast_group_id` is considered. All
+other fields in `MulticastGroupEntry`` are ignored. To perform a *wildcard* read
```
stefanheule: Typo: double `

proto/p4/v1/p4runtime.proto
```diff
@@ -394,8 +394,19 @@ message Replica {
 // each replica's egress input metadata will be set to the respective values
 // programmed in the multicast group entry.
 message MulticastGroupEntry {
+  // Set multicast_group_id to MULTICAST_GROUP_ID_WILDCARD for wildcard reads
   uint32 multicast_group_id = 1;
   repeated Replica replicas = 2;
+
+  // We define a special reserved value for the multicast_group_id field. This
```
stefanheule: This comment is only about `MULTICAST_GROUP_ID_WILDCARD`, maybe move it there.

proto/p4/v1/p4runtime.proto
```diff
@@ -412,10 +423,20 @@ message MulticastGroupEntry {
...
   uint32 class_of_service = 3;
   int32 packet_length_bytes = 4;
+
+  // We define a special reserved value for the session_id field. This value can
+  // be used to do a wildcard read on all configured clone sessions. This value
+  // cannot be used in Write RPCs.
+  enum CloneSessionIdReserved {
+    CLONE_SESSION_ID_FIRST = 0;
+    // 0xffffffff
+    CLONE_SESSION_ID_WILDCARD = -1;
```
stefanheule: Maybe we should just spell this out as 0xffffffff, especially since negative numbers aren't actually valid IDs (they are uint32).

antoninbas: Not possible in Protobuf IIRC, it will be rejected by protoc

docs/v1/P4Runtime-Spec.mdk
antoninbas: I don't think there is a distinction in this case, but I updated my PR to clarify.

docs/v1/P4Runtime-Spec.mdk

proto/p4/v1/p4runtime.proto
antoninbas: Done

proto/p4/v1/p4runtime.proto
antoninbas: I added it for symmetry with clone sessions, for which it is required since the enum must start with a value of 0.

Agreed on the naming. I replaced "Reserved" with "Values" in both cases.

konne-google:
Sorry, to come so late to the party. I feel like adding 0xffffffff as the special wildcard value is inconsistent with the rest of the read API, where 0 (aka unset) is used as wildcard. Using unset as wildcard, also seems more intuitive.

Another question is, could we just disallow multicast groups and clone sessions with id 0?

jafingerhut:
Konstantin, a little bit of history/context on these proposed changes:

There should be no problem disallowing multicast group ids equal to 0, at least for PSA and several other P4 switch architectures I have seen.

The only reason that clone session id 0 is considered valid in PSA v1.1 is the concern that perhaps clone session ids would be fairly few in number in some implementations, and thus 'burning' 0 might seem wasteful.  I have no objection to making it invalid, if enough folks think that is OK.

Given the current PSA spec, it seemed the more cautious approach to enable clone session id values of 0 to be used in P4Runtime as valid, and then from there, why not make wildcard multicast group reads consistent with that.

antoninbas:
@konne-google I agree that it is inconsistent and honestly we should have done a complete review of all entities and their wildcard read support before releasing P4Runtime 1.0. Then we would have been able to wrap them in Protobuf messages (like we did for counter / meter index), which is still my preferred solution for P4Runtime 2.0.

The name "id" may not actually be ideal for clone sessions and multicast groups, because unlike most ids in P4Runtime, they don't designate specific P4 objects (which is also the case for AP group / member ids), and they are meant to be used as is as datapath values / signals (unlike AP group / member ids which are just used to introduce indirection in the entry programming).

Personally I would be fine with disallowing 0 for clone sessions (it is already disallowed for multicast groups). I complained about that discrepancy before and there was already a discussion about it: https://github.com/p4lang/p4runtime/issues/44. But that requires changing the PSA spec theoretically and waiting for a new PSA release. We can also disallow it specifically at the P4Runtime level, but is it a good idea to introduce (another?) discrepancy with PSA? My primary concern is to actually be able to do support wildcard reads for the PRE as soon as possible and I am fine with several alternatives.

@jafingerhut Maybe you have an opinion on this specific issue?

PR #236, opened 2019-08-28
Support wildcard reads for multicast groups and clone sessions

antoninbas:
For consistency with the rest of the P4Runtime API, a wildcard Read is
achieved by "setting" the `multicast_group_id` (resp. `session_id`)
field to its default value, which is 0 for sclar fields.

This means that we introduce a small difference with the PSA spec, which
explicitly allows 0 for clone session ids.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -3844,8 +3844,8 @@ configured in the target. This means that 0 can never be used as a `session_id`
 value when inserting a clone session, whether or not numeric translation is
 enabled for clone session ids. If translation is *not* enabled, we effectively
 "lose" one clone session, assuming the target supports 0 as mandated by the PSA
-specification. Assuming this is an issue (&eg; because the target supports a
-very limited number of clone sessions), one could enable translation on
+specification. If this is an issue (&eg; because the target supports a very
+limited number of clone sessions), one can enable translation on
 `CloneSessionId_t` and map any non-zero SDN session id to the data plane clone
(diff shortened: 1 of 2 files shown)


antoninbas:
Friendly ping, I need an approval to merge this

stefanheule:
Thanks for making the changes, Antonin!

chrispsommers:
Nit: I noticed "e.g'" has been globally replaced with "&eg;" which renders as &eg; (HTML equals-or-greater entity). Am I the only one who noticed this?

antoninbas:
@chrispsommers "e.g." has been replaced by "`&eg;`", but the Madoko renderer takes care of generating the appropriate HTML / LaTex code (http://madoko.org/reference.html#sec-entity).  I reviewed https://s3-us-west-2.amazonaws.com/p4runtime/travis/support-wildcard-reads-for-pre-entries-2/P4Runtime-Spec.html and I didn't see any issue (no equals-or-greater sign instead of e.g.). Could you take a look at the document and see if you see any issue (and point me to the location)?

BTW, using "`&eg;`" instead of "e.g." seems to be the best practice when using Madoko. There is probably no benefit for the generated HTML but maybe the LaTex is more "robust" (to line breaks, ...)?

chrispsommers:
@antoninbas  You're right, Madoko renders "&eg;" correctly in the HTML. The funny thini is GitHub renders unprotected entity as the symbol.

antoninbas:
Closed by mistake I believe

antoninbas:
Thanks for the reviews!

Issue #234, opened 2019-08-21
Add AnnoPropertyMap containing annotation kv-pairs


chrispsommers:
I've found it useful to annotate P4 code with custom annotations of the form:
```
@my_anno1(a=b,c=d)
table my_table {...}
```

Where the values `b` and `d`  contain arbitrary text. I take the P4info and parse the raw annotations into a usable form. Ultimately there may be several such annotations for one entity, with multiple k-v pairs embedded in each. It would be useful to do this parsing in p4c and produce something easy to digest.

I propose something like the following. Define two new `message`s (names are arbitrary):
```
// A dictionary of strings describing something's properties
message PropertyMap {
  map <string,string> property = 1;   // key-value pairs
}

// A map of property maps (nested maps)
message AnnoPropertyMap {
  map <string,PropertyMap> propertyMap = 1;   // key-map pairs; key = annotation's tag
}
```

Then make ` AnnoPropertyMap anno_prop_map=<x>` a sibling to `annotations `in every place it appears.
Then with P4 code such as this:
```
@my_anno1(a=b,c=d)
@my_anno1(e=f)
@my_anno2(w=x,y=z)
table my_table {...}
```
you'd get raw annotations (already in P4info) like this:
```
annotations: "@my_anno1(a = b , c = d)"
annotations: "@my_anno1(e = f)"
annotations: "@my_anno2(w = x , y = z)"
```
Note: `p4c` already parses them as k-v pairs then re-serializes them with spaces around the = sign.)

The enhancement would produce a sibling to the annotations something like this (expressed as python dicts):
```
"anno_prop_map": {
   "my_anno1": {
    "a": "b", 
    "c": "d", 
    "e": "f"
  },
"my_anno2": {
    "y": "z", 
    "w": "x"
  }
}
```
The result is a set of nested maps; the outer one is keyed by the annotation tag and the inner ones by the key of each k-v pair.

Note `@my_anno1()` appeared twice in the p4 code and the k-v pairs got merged,  consistent with how p4c already merges them.

antoninbas:
This may not need to wait until v2.0 if the proposal is to *add* a new Protobuf field in addition to the existing `annotations` field, since it would not break backward-compatibility. We can discuss it at the next API WG meeting (09/28).

chrispsommers:
It is a new field, it should be totally backward-compatible. Thanks.

antoninbas:
[...]

PR #238, opened 2019-09-09
Chris/add anno property maps

chrispsommers:
#234 - Add annotation maps. I added to `p4info.proto` and updated the P4Runtime spec with an example. I generated the example P4Info by using the proposed enhanced `p4info.proto`, reading in actual `xxx.p4info.txt` from current `p4c`, parsing the P4info using python, adding parsed annotations to the P4info "DOM" (with generated `p4info_pb2.py` libs from the new `p4info.proto`) then emitting it.

If acceptable, hopefully `p4c` can be enhanced to do the *adding parsed annotations to the P4info "DOM"* task

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1016,6 +1016,104 @@ table my_ipv4_lkup {
   ...
 }
 ~ End p4example
+### Custom Annotations in P4Info
+P4 allows flexible annotations[@P4Annotations]. Users may add their own annotations as long as they obey the language syntax. To facilitate consumption of user-defined annotations, P4Info has an `annotation_map` for any element which can accept such annotations, &ie; the `PkgInfo`, `Preamble`, `MatchField`, `ActionRef`, `Action` and `ControllerPacketMetadata` messages. The `annotation_map` contains a suitable parsed representation for the different types of annotation bodies, namely: `expression_list` and `kv_map`. Empty annotation bodies are signified by an empty `value` field  in the `annotation_map` entry. 
+
+Note that `AnnotationBody` allows for simultaneous `expression_list` and `kv_map` entries even though the P6-16 specification grammar does not at this writing.
+
+~ Begin Proto
+// A parsed representation of the annotation body.
+// Ex. empty annotation:           @my_anno1()
+// Ex. expression list annotation: @my_anno2(1+1, "Hello", 1 << 3)
+// Ex. KV-pair annotation:         @my_anno3(a=b,c=d)
+// Ex. Cumulative kv-pair          @my_anno3(e=f)  // adds to my_anno3's properties
+message AnnotationBody {
+  repeated string expressions = 1;
+  map <string,string> properties = 2;
+}
+~ End Proto
+
+Given the following P4 source fragment:
+~ Begin P4Example
+@my_anno1()
+@my_anno2("Some random content: for i in range(10)")
+@my_anno3(1+1, "Hello", 1 << 3)
+@my_anno4(a=b,c=d)
+@my_anno4(e=f)
+@my_anno5(w=x, "expression one", y=z, expression two)
+table test1 {
+    key = { hdrs.data.f1 : ternary; }
+    actions = {
+        setb1;
+        noop;
+    }
+    default_action = noop;
+}
+~ End P4Example
+The corresponding P4Info `Preamble` fragment would be as follows. Note the appearance of the raw `annotations` fields which are deprecated in favor of the `annotation_map` fields.
+~ Begin Prototext
+  preamble {
+    id: 33619894
+    name: "ingress.test1"
+    alias: "test1"
+    annotations: "@my_anno1"
+    annotations: "@my_anno2(\"Some random content: for i in range(10)\")"
+    annotations: "@my_anno3(1 + 1 , \"Hello\" , 1 << 3)"
+    annotations: "@my_anno4(a = b , c = d)"
+    annotations: "@my_anno4(e = f)"
+    annotations: "@my_anno5(w = x , \"expression one\" , y = z , expression two)"
+    annotation_map {
+      key: "my_anno1"
+      value {
+      }
+    }
+    annotation_map {
+      key: "my_anno2"
+      value {
+        expressions: "\"Some random content: for i in range"
+      }
+    }
+    annotation_map {
+      key: "my_anno3"
+      value {
+        expressions: "\"Hello\""
+        expressions: "1 << 3"
+        expressions: "1 + 1"
+      }
+    }
+    annotation_map {
+      key: "my_anno4"
+      value {
+        properties {
+          key: "a"
+          value: "b"
+        }
+        properties {
+          key: "c"
+          value: "d"
+        }
+        properties {
+          key: "e"
+          value: "f"
+        }
+      }
+    }
+    annotation_map {
+      key: "my_anno5"
+      value {
+        expressions: "\"expression one\""
+        expressions: "expression two"
+        properties {
+          key: "w"
+          value: "x"
+        }
+        properties {
+          key: "y"
+          value: "z"
+        }
+      }
+    }
+~ End Prototext
 
 ## `PkgInfo` Message
 
@@ -1107,7 +1205,6 @@ Built for data-center profile.")
 PSA_Switch(IgPipeline, PacketReplicationEngine(), EgPipeline,
            BufferingQueueingEngine()) main;
 ~ End p4example
-
 ## ID Allocation for P4Info Objects { #sec-id-allocation}
 
 P4Info objects receive a unique ID, which is used to identify the object in
@@ -5472,79 +5569,6 @@ the purpose of adding features for the P4Runtime API.
 | `@p4runtime_translation` | See sections [#sec-user-defined-types], [#sec-translation-of-port-numbers] |
 +--------------------------+---------------------------------------+
 ~
-### Custom Annotations in P4Info
-P4 allows flexible annotations. Users may add their own annotations as long as they obey the language syntax.
-To facilitate consumption of user-defined annotations, P4Info has an `anno_map` for any element which can accept
-such annotations, e.g. the `PkgInfo`, `Preamble`, `MatchField`, `ActionRef`, `Action` and
-`ControllerPacketMetadata` messages. The `anno_map` contains a suitable parsed representation for the different
-types of annotation bodies, namely: empty, free-form string, and a key-value-map.
-
-Given the following P4 source fragment:
-
-~ Begin P4Example
-    @my_anno1()
-    @my_anno2("Some random content: for i in range(10)")
-    @my_anno3(a=b,c=d)
-    @my_anno3(e=f)
-    table test1 {
-        key = { hdrs.data.f1 : ternary; }
-        actions = {
-            setb1;
-            noop;
-        }
-        default_action = noop;
-    }
-~ End P4Example
-
-The corresponding P4Info `Preamble` fragment would be as follows.
-Note the lingering presence of the raw `annotations` fields, which are deprecated in favor of the newer
-`anno_map` fields and slated to be removed in some future release.
-~ Begin Prototext
-  preamble {
-    id: 33619894
-    name: "ingress.test1"
-    alias: "test1"
-    annotations: "@my_anno1"
-    annotations: "@my_anno2(\"Some random content: for i in range(10)\")"
-    annotations: "@my_anno3(a = b , c = d)"
-    annotations: "@my_anno3(e = f)"
-    anno_map {
-      key: "my_anno1"
-      value {
-        empty {
-        }
-      }
-    }
-    anno_map {
-      key: "my_anno2"
-      value {
-        string {
-          value: "\"Some random content: for i in range"
-        }
-      }
-    }
-    anno_map {
-      key: "my_anno3"
-      value {
-        kv_map {
-          property {
-            key: "a"
-            value: "b"
-          }
-          property {
-            key: "c"
-            value: "d"
-          }
-          property {
-            key: "e"
-            value: "f"
-          }
-        }
-      }
-    }
-  }
-~ End Prototext
-
 ## A More Complex Value Set Example { #sec-value-set-example}
 
 This section includes a more complex Value Set example, with multiple matches of
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -5472,6 +5472,78 @@ the purpose of adding features for the P4Runtime API.
 | `@p4runtime_translation` | See sections [#sec-user-defined-types], [#sec-translation-of-port-numbers] |
 +--------------------------+---------------------------------------+
 ~
+### Custom Annotations in P4Info
+P4 allows flexible annotations. Users may add their own annotations as long as they obey the language syntax.
+To facilitate consumption of user-defined annotations, P4Info has an `anno_map` for any element which can accept
+such annotations, e.g. the `PkgInfo`, `Preamble`, `MatchField`, `ActionRef`, `Action` and
+`ControllerPacketMetadata` messages. The `anno_map` contains a suitable parsed representation for the different
+types of annotation bodies, namely: empty, free-form string, and a key-value-map.
+
+Given the following P4 source fragment:
+
+~ Begin P4Example
+    @my_anno1()
+    @my_anno2("Some random content: for i in range(10)")
+    @my_anno3(a=b,c=d)
+    @my_anno3(e=f)
+    table test1 {
+        key = { hdrs.data.f1 : ternary; }
+        actions = {
+            setb1;
+            noop;
+        }
+        default_action = noop;
+    }
+~ End P4Example
+
+The corresponding P4Info `Preamble` fragment would be as follows.
+Note the lingering presence of the raw `annotations` fields, which are deprecated in favor of the newer
+`anno_map` fields and slated to be removed in some future release.
+~ Begin Prototext
+  preamble {
+    id: 33619894
+    name: "ingress.test1"
+    alias: "test1"
+    annotations: "@my_anno1"
+    annotations: "@my_anno2(\"Some random content: for i in range(10)\")"
+    annotations: "@my_anno3(a = b , c = d)"
+    annotations: "@my_anno3(e = f)"
+    anno_map {
+      key: "my_anno1"
+      value {
+        empty {
+        }
+      }
+    }
+    anno_map {
+      key: "my_anno2"
+      value {
+        string {
+          value: "\"Some random content: for i in range"
+        }
+      }
+    }
+    anno_map {
+      key: "my_anno3"
+      value {
+        kv_map {
+          property {
+            key: "a"
+            value: "b"
+          }
+          property {
+            key: "c"
+            value: "d"
+          }
+          property {
+            key: "e"
+            value: "f"
+          }
+        }
+      }
+    }
+  }
+~ End Prototext
 
 ## A More Complex Value Set Example { #sec-value-set-example}
 
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1033,7 +1033,8 @@ comma-separated expressions and/or key-value ("property") pairs. Empty
 annotation bodies are signified by zero `AnnotationElement`s in the `elements`
 field of `ParsedAnnotation`.
 
-The `AnnotationElement` and `ParsedAnnotation` messages are defined in `p4types.proto`
+The `AnnotationElement` and `ParsedAnnotation` messages are defined in
+`p4types.proto`
 
 ~ Begin Proto
 // One parsed element within an annotation's body.
@@ -1052,8 +1053,11 @@ message ParsedAnnotation {
 
 The following invariants hold:
 
-1. For any P4 entity with annotations, there are no two parsed_annotations that have the same name.
-2. Within a `ParsedAnnotation`, there are no two `AnnotationElement`s that have the same key, except for the empty key, which can appear any number of times.
+1. For any P4 entity with annotations, there are no two parsed_annotations that
+have the same name.
+
+2. Within a `ParsedAnnotation`, there are no two `AnnotationElement`s that have
+the same key, except for the empty key, which can appear any number of times.
 
 
 Given the following P4 source fragment:
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1020,47 +1020,60 @@ table my_ipv4_lkup {
 ~ End p4example
 
 ### Custom Annotations in P4Info
+
 P4 allows flexible annotations[@P4Annotations]. Users may add their own
 annotations as long as they obey the language syntax. To facilitate consumption
-of user-defined annotations, P4Info has an `AnnotationMap` for any element which
-can accept such annotations, &ie; the `PkgInfo`, `Preamble`, `MatchField`,
-`ActionRef`, `Action` and `ControllerPacketMetadata` messages, as well as the
-type specification messages in `p4types.proto`. The `AnnotationMap` contains a
-suitable parsed representation for the different types of annotation bodies,
-namely: `expressions` and `properties`. Empty annotation bodies are signified by
-an empty `value` field  in the `AnnotationMap` entry. 
+of user-defined annotations, P4Info has a `repeated ParsedAnnotation` for any
+element which can accept such annotations, &ie; the `PkgInfo`, `Preamble`,
+`MatchField`, `ActionRef`, `Action` and `ControllerPacketMetadata` messages, as
+well as the type specification messages in `p4types.proto`. The
+`ParsedAnnotation` contains a parsed representation which accomodates both
+unstructured annotation content, as well as common structured use-cases:
+comma-separated expressions and/or key-value ("property") pairs. Empty
+annotation bodies are signified by zero `AnnotationElement`s in the `elements`
+field of `ParsedAnnotation`.
 
-Note that `AnnotationBody` allows for simultaneous `expressions` and
-`properties` entries even though the P6-16 specification grammar does not at this
-writing.
-
-The `AnnotationBody` and `AnnotationMap` messages are defined in `p4types.proto`:
+The `AnnotationElement` and `ParsedAnnotation` messages are defined in `p4types.proto`
 
 ~ Begin Proto
-// A parsed representation of the annotation body.
-// Ex. empty annotation:           @my_anno1()
-// Ex. expression list annotation: @my_anno2(1+1, "Hello", 1 << 3)
-// Ex. KV-pair annotation:         @my_anno3(a=b,c=d)
-// Ex. Cumulative kv-pair          @my_anno3(e=f)  // adds to my_anno3's properties
-message AnnotationBody {
-  repeated string expressions = 1;
-  map <string, string> properties = 2;
+// One parsed element within an annotation's body.
+message AnnotationElement {
+// Key is empty if and only if no key was specified
+  string key = 1;
+  string value = 2;
 }
 
-// A map of annotations keyed by the annotation tag following the @-sign
-message AnnotationMap {
-  map <string, AnnotationBody> annotations = 1;
+// An ordered list of annotations identified by the name
+message ParsedAnnotation {
+  string name = 1;
+  repeated AnnotationElement elements = 2;
 }
 ~ End Proto
 
+The following invariants hold:
+
+1. For any P4 entity with annotations, there are no two parsed_annotations that have the same name.
+2. Within a `ParsedAnnotation`, there are no two `AnnotationElement`s that have the same key, except for the empty key, which can appear any number of times.
+
+
 Given the following P4 source fragment:
 ~ Begin P4Example
-@my_anno1()
-@my_anno2("Some random content: for i in range(10)")
-@my_anno3(1+1, "Hello", 1 << 3)
-@my_anno4(a=b,c=d)
-@my_anno4(e=f)
-@my_anno5(w=x, "expression one", y=z, expression two)
+@MyAnno1()
+@MyAnno2("Some random content: for i in range(10)")
+@MyAnno3(1+1, "Hello", 1 << 3)
+@MyAnno4(a=b,c=d)
+@MyAnno4(e=f)
+@MyAnno5(a=1,a=2)
+@MyAnno6(e=f)
+@MyAnno6(1)
+@MyAnno7(
+- What do you get if you multiply six by nine?
+- Six by nine. Forty two.
+- That's it. That's all there is.
+- I always thought there was something fundamentally wrong with the
+  universe.
+0xdeadbeef
+)
 table test1 {
     key = { hdrs.data.f1 : ternary; }
     actions = {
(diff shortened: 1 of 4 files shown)


chrispsommers:
Hey peeps, any time to review this? Thanks!

antoninbas:
Hi Chris. I’m on vacation which is why I haven’t reviewed this yet. I should be able to take a look at it in the upcoming days.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5472,6 +5472,78 @@ the purpose of adding features for the P4Runtime API.
 | `@p4runtime_translation` | See sections [#sec-user-defined-types], [#sec-translation-of-port-numbers] |
 +--------------------------+---------------------------------------+
 ~
+### Custom Annotations in P4Info
+P4 allows flexible annotations. Users may add their own annotations as long as they obey the language syntax.
```
antoninbas: I suggest adding a reference to the P4_16 spec here: https://p4.org/p4-spec/docs/P4-16-v1.1.0-spec.html#sec-annotations

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5472,6 +5472,78 @@ the purpose of adding features for the P4Runtime API.
 | `@p4runtime_translation` | See sections [#sec-user-defined-types], [#sec-translation-of-port-numbers] |
 +--------------------------+---------------------------------------+
 ~
+### Custom Annotations in P4Info
```
antoninbas: I think this should not be in this appendix, but instead moved to section 6 ("The P4Info Message"). The appendix was just meant as a way to quickly lookup which annotations are in the realm of P4Runtime.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5472,6 +5472,78 @@ the purpose of adding features for the P4Runtime API.
...
+P4 allows flexible annotations. Users may add their own annotations as long as they obey the language syntax.
+To facilitate consumption of user-defined annotations, P4Info has an `anno_map` for any element which can accept
+such annotations, e.g. the `PkgInfo`, `Preamble`, `MatchField`, `ActionRef`, `Action` and
```
antoninbas: I suggest replacing "e.g." with "&ie;" as it seems you are giving the exhaustive list of elements for which we have annotations

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5472,6 +5472,78 @@ the purpose of adding features for the P4Runtime API.
...
+types of annotation bodies, namely: empty, free-form string, and a key-value-map.
+
+Given the following P4 source fragment:
+
+~ Begin P4Example
+    @my_anno1()
+    @my_anno2("Some random content: for i in range(10)")
+    @my_anno3(a=b,c=d)
+    @my_anno3(e=f)
+    table test1 {
```
antoninbas: Can we add one more example? I believe the following one is legit (but useless) P4:
```
@my_anno4(1+1, "Hello", 1 << 3)
```
I believe that with this proposal we would use a free-form string for this.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5472,6 +5472,78 @@ the purpose of adding features for the P4Runtime API.
...
+P4 allows flexible annotations. Users may add their own annotations as long as they obey the language syntax.
+To facilitate consumption of user-defined annotations, P4Info has an `anno_map` for any element which can accept
+such annotations, e.g. the `PkgInfo`, `Preamble`, `MatchField`, `ActionRef`, `Action` and
+`ControllerPacketMetadata` messages. The `anno_map` contains a suitable parsed representation for the different
+types of annotation bodies, namely: empty, free-form string, and a key-value-map.
+
+Given the following P4 source fragment:
+
+~ Begin P4Example
+    @my_anno1()
```
antoninbas: nit: any reason for the 4-space indentation here (and 2-space indentation in the "Prototext" block below)?

proto/p4/config/v1/p4info.proto
```diff
@@ -65,6 +86,8 @@ message PkgInfo {
   string contact = 7;
   // url for more information, e.g. "http://support.p4.org/ref/p4/switch.p4_v1.0"
   string url = 8;
+  // Annotation map derived from raw annotations
+  map <string,AnnotationBody> anno_map = 9;
```
antoninbas: I would actually prefer spelling it out: `annotation_map`

proto/p4/config/v1/p4info.proto
```diff
@@ -46,6 +48,24 @@ message Documentation {
...
+// Ex. empty annotation:   @my_anno1()
+// Ex. string annotation:  @my_anno2("Any text at all", for (i = 1 to 10) {print i} )
+// Ex. KV-pair annotation: @my_anno3(a=b,c=d)
+// NOTE: multiple KV annotations with the same tag are cumulative, so @my_anno3(a=b,c=d) in conjunction with
+//       @my_anno3(e=f) would result in 3 kv-pairs for my_anno3.
+message AnnotationBody {
+  oneof anno {
+    google.protobuf.Empty empty = 1;          // The annotation body is empty
+    google.protobuf.StringValue string = 2;   // The annotation body contains free-form string or expression 
+    KVmap kv_map = 3;                         // The annotation body contains one or more KV-pairs
```
antoninbas: It's not a big deal but I believe defining a separate `KVmap` message is superfluous here, we can just put the `map<string, string>` directly in the `oneof`.

proto/p4/config/v1/p4info.proto
```diff
@@ -46,6 +48,24 @@ message Documentation {
...
+}
+
+// A parsed representation of the annotation body.
+// Ex. empty annotation:   @my_anno1()
+// Ex. string annotation:  @my_anno2("Any text at all", for (i = 1 to 10) {print i} )
+// Ex. KV-pair annotation: @my_anno3(a=b,c=d)
+// NOTE: multiple KV annotations with the same tag are cumulative, so @my_anno3(a=b,c=d) in conjunction with
+//       @my_anno3(e=f) would result in 3 kv-pairs for my_anno3.
+message AnnotationBody {
+  oneof anno {
```
antoninbas: what about `body` instead of `anno`?

proto/p4/config/v1/p4info.proto
```diff
@@ -46,6 +48,24 @@ message Documentation {
...
+    google.protobuf.Empty empty = 1;          // The annotation body is empty
+    google.protobuf.StringValue string = 2;   // The annotation body contains free-form string or expression
```
antoninbas: I believe `string` could be used here instead of `google.protobuf.StringValue` since the field is already in a `oneof`, unless there is something I'm missing.

proto/p4/config/v1/p4info.proto
antoninbas: Could we wrap the `map` in a message maybe, since it is used in quite a few places?

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Agreed, will do.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Agreed, will do.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Agreed, will do.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Nope, lazy cut and paste from source code containing indents. Will fix.

proto/p4/config/v1/p4info.proto
chrispsommers: OK, will do.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Item 1 - repeated expressions - great idea, I'll make the change. Also, I'm removing the  `empty` field because the container `AnnotationBody` can simply be empty. I'm removing the `oneof` to allow for both expressions and kv-pairs in an annotation body, even though P4-16 currently precludes it. The `expressions` can simply be a `repeated string`, no need for a container now that `oneof` is gone. Also removing the `KVmap` message and renaming to `properties` since that is a more conventional term.
```
// A parsed representation of the annotation body.
// Ex. empty annotation:           @my_anno1()
// Ex. ExpressionList annotation:  @my_anno2(1+1, "Hello", 1 << 3)
// Ex. KV-pair annotation:         @my_anno3(a=b,c=d)
// Ex. Cumulative kv-pair          @my_anno3(e=f)  // adds to my_anno3's properties
message AnnotationBody {
  repeated string expressions = 1;
  map <string,string> properties = 2;
}
```

Item 2 - reference - will do, great idea.

proto/p4/config/v1/p4info.proto
chrispsommers: Agreed, see earlier comment. However, be warned a `oneof` doesn't allow a `map` so I couldn't have done this without removing the `oneof`, which I did.

proto/p4/config/v1/p4info.proto
chrispsommers: Good idea, but I removed the `oneof` so it's moot.

proto/p4/config/v1/p4info.proto
chrispsommers: Good point, I've greatly simplified `AnnotationBody` anyway and now it doesn't need any wrappers.

proto/p4/config/v1/p4info.proto
chrispsommers: Originally I had that, it looked nicer in the protobuf definition, but it added another message layer which has only one field, i.e. not a lot of value. Do you still feel it's worth it?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1016,6 +1016,104 @@ table my_ipv4_lkup {
   ...
 }
 ~ End p4example
+### Custom Annotations in P4Info
```
antoninbas: can we add an empty tine before the section?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1016,6 +1016,104 @@ table my_ipv4_lkup {
...
+    actions = {
+        setb1;
+        noop;
+    }
+    default_action = noop;
+}
+~ End P4Example
+The corresponding P4Info `Preamble` fragment would be as follows. Note the appearance of the raw `annotations` fields which are deprecated in favor of the `annotation_map` fields.
+~ Begin Prototext
+  preamble {
```
antoninbas: this block is still indented for some reason?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1107,7 +1205,6 @@ Built for data-center profile.")
 PSA_Switch(IgPipeline, PacketReplicationEngine(), EgPipeline,
            BufferingQueueingEngine()) main;
 ~ End p4example
-
```
antoninbas: ?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5472,7 +5569,6 @@ the purpose of adding features for the P4Runtime API.
 | `@p4runtime_translation` | See sections [#sec-user-defined-types], [#sec-translation-of-port-numbers] |
 +--------------------------+---------------------------------------+
 ~
-
```
antoninbas: ?

proto/p4/config/v1/p4info.proto
```diff
@@ -46,6 +46,16 @@ message Documentation {
...
 }
 
+// A parsed representation of the annotation body.
+// Ex. empty annotation:           @my_anno1()
+// Ex. expression list annotation: @my_anno2(1+1, "Hello", 1 << 3)
+// Ex. KV-pair annotation:         @my_anno3(a=b,c=d)
+// Ex. Cumulative KV-pair          @my_anno3(e=f)  // adds to my_anno3's properties
+message AnnotationBody {
+  repeated string expressions = 1;
+  map <string,string> properties = 2;
```
antoninbas: could we leave a space after the comma for all the maps (just like we did for the `P4TypeInfo` message)?

proto/p4/config/v1/p4info.proto
```diff
@@ -56,6 +66,7 @@ message PkgInfo {
   // brief and detailed descriptions
   Documentation doc = 3;
   // Miscellaneous metadata, free-form; a way to extend PkgInfo
+  // raw annotations - deprecated, use annotation_map
   repeated string annotations = 4;
```
antoninbas: I suggest adding the `[deprecated=true]` option to all `repeated string annotations` fields.

See https://developers.google.com/protocol-buffers/docs/proto3#options

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1016,6 +1016,104 @@ table my_ipv4_lkup {
   ...
 }
 ~ End p4example
+### Custom Annotations in P4Info
+P4 allows flexible annotations[@P4Annotations]. Users may add their own annotations as long as they obey the language syntax. To facilitate consumption of user-defined annotations, P4Info has an `annotation_map` for any element which can accept such annotations, &ie; the `PkgInfo`, `Preamble`, `MatchField`, `ActionRef`, `Action` and `ControllerPacketMetadata` messages. The `annotation_map` contains a suitable parsed representation for the different types of annotation bodies, namely: `expression_list` and `kv_map`. Empty annotation bodies are signified by an empty `value` field  in the `annotation_map` entry.
```
antoninbas: looks like I forgot about p4types.proto...

so me may need to rephrase:

> PkgInfo, ..., ControllerMetadata and the type specification messages in p4types.proto

proto/p4/config/v1/p4info.proto
chrispsommers: OK, I added wrapper  AnnotationMap and named  the `map` field within `annotations`. The protobuf looks nice and it has symmetry with the deprecated '`annotations` field.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Done

docs/v1/P4Runtime-Spec.mdk
chrispsommers: I added the missing `table {` opening line which now makes the indent sensible.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: DOh!

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Double-Doh!

proto/p4/config/v1/p4info.proto
chrispsommers: Great idea, done.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Fixed, thanks.

proto/p4/config/v1/p4info.proto
chrispsommers: Done.

chrispsommers:
@antoninbas I appreciate all  the great feedback, it resulted in a much better PR. If this is acceptable, perhaps we can review at the next WG meeting.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1017,6 +1019,129 @@ table my_ipv4_lkup {
...
+    annotation_map {
+      annotations {
+        key: "my_anno1"
+        value {
+        }
+      }
+      annotations {
+        key: "my_anno2"
+        value {
+          expressions: "\"Some random content: for i in range"
```
antoninbas: I think you are missing the end of the annotation body here: `(10)\"`

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1017,6 +1019,129 @@ table my_ipv4_lkup {
...
+writing.
+
+The `AnnotationBody` and `AnnotationMap` messages are defined in `p4types.proto`:
+
+~ Begin Proto
+// A parsed representation of the annotation body.
+// Ex. empty annotation:           @my_anno1()
+// Ex. expression list annotation: @my_anno2(1+1, "Hello", 1 << 3)
+// Ex. KV-pair annotation:         @my_anno3(a=b,c=d)
+// Ex. Cumulative kv-pair          @my_anno3(e=f)  // adds to my_anno3's properties
```
antoninbas: I took a look at the rendered HTML (https://s3-us-west-2.amazonaws.com/p4runtime/travis/chris/add-anno-property-maps/P4Runtime-Spec.html#sec-custom-annotations-in-p4info) and this line was too long for the text box

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1017,6 +1019,129 @@ table my_ipv4_lkup {
...
+message AnnotationBody {
+  repeated string expressions = 1;
```
stefanheule: Should this be a oneof?  I think it can only be one or the other, never both, correct?

stefanheule:
> One change is that with AnnotationMap, we lose the order of annotations. I can't say I have a super clear use-case for why preserving that odering is important, but I'm generally skeptical when we parse some language and just forget about the ordering of elements, even if the order should not matter.
> 
> Questions:
> 
> * Can we think of reasons that would rely on ordering?  Would we ever want to "deparse" our representation and expect to get the same as the original program?
> * Is there an easy way to add ordering back in?  What do would we lose if we don't use a map, but instead a repeated list?

In addition, we also lose the order for KV lists, which again does not seem great, even if I can't actually think of a example where it matters necessarily.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: @stefanheule , Originally in my proposal it was `oneof`, but later I removed `oneof` to allow for future enhancement to the P4-16 spec.If you look at the `@my_anno5` example I cited, I can value in allowing a mix of kv-pairs and expressions

stefanheule: Ah my bad, I should have followed the previous discussion more closely.  But my understanding is correct that P4-16 spec right now only allows one or the other, correct?  Are we aware of plans that they want to generalize this?

We might not want to have an example of something that isn't actually valid according to the current spec (even if we decide not to use a oneof).

chrispsommers:
@stefanheule Re: maintaining the ordering or being able to regenerate the original code so to speak - it's a good point, but one could say the same thing for most P4 constructs: once parsed, something is lost. If this is important, we could retain the original annotations and remove the deprecation.

Here's an alternative along the lines of your comment:
```
// One parsed element within an annotation's body
message AnnotationElement {
  message AnnotationProperty {
    string key = 1;
    string value = 2;
  }
  oneof element {
    string expression = 1;
    AnnotationProperty property = 2;
  }
}

// An ordered list of annotations identified by the tag
message ParsedAnnotation {
  string annotation_tag = 1;
  repeated AnnotationElement elements = 2;
}
```
Then, instead of the original proposal, replace all occurrences of 
```
  AnnotationMap annotation_map = ...;
```
with
```
 repeated ParsedAnnotation parsed_annotations = ...;
```
Note `AnnotationProperty` is a like a one-element `map`; since there's  only ever one pair it is not worth using a `map` field type.

The benefits to this alternative are, ordering of everything is preserved exactly, and you could presumably regenerate the original source code, subject to some normalization. The drawback is  more work in any application consuming these. For example, one want to put all the properties into a map and all  the expressions into their own list, mimicking the original proposal. With this approach, it remains a choice whether to coalesce annotations with duplicate tags. If we do, we lose reversibility.  If we don't, we have to accumulate kv-pairs (and expressions) in the application.

This also points out my original proposal should have specified that expressions get coalesced just like kv-pairs.  I'll comment my own PR as a reminder.

antoninbas:
I'm fine with the second proposal as well. It is still better than what we have currently, since the application does not have to actually parse the raw annotations anymore. With the new proposal we can support duplicates and preserve the order, but deny the application the convenience of an associative container. It is probably good not to make assumptions about the semantics of user-defined annotations. By preserving order and duplicates, we enable the consumer to choose: accumulate duplicates or override, and that can be defined on a per-annotation basis. So I'm leaning towards the second one at this time. Let's make a final decision at the next meeting.

On a side note @chrispsommers I realized that this change has impacts on other parts of the spec (e.g. the description of the `Preamble` message, you can search the spec for "annotation"). These will need to be updated as part of this PR. I suggest you wait until we reach an agreement, or you may need to update the same paragraphs multiple times...

docs/v1/P4Runtime-Spec.mdk
stefanheule: Just to close this thread, I was actually wrong, the P4-16 spec does (sort of) allow mixing them, by accumulating through the use of multiple annotations.

> Multiple annotations of either form may be used and are cumulative, i.e. the list of k-v pairs will accumulate.

To me, this reads that `@Foo(1) @Foo(a=b)` is allowed, and will result in (conceptually) a single annotation `Foo` that has both `1` and `a=b` as it's elements.  Writing it as a single annotation (as in `@Foo(1, a=b)`) is forbidden by the grammer, though, not sure if that's on oversight or on purpose.

stefanheule:
Thanks for working on this, Chris, I really appreciate it.  I agree that we probably won't be able to have parse-deparse be the identity (there's whitespace, too, for example), so that was a bad example and should not be a goal.

I have a few more thoughts:

1. Besides order that I think is important to preserve, there are two more points I'd like us to consider: repetition of keys and repetition of the same annotation.  Unlike what Antonin says, it looks like the consumer is not free to choose override vs duplication, as the P4 spec says what the semantics are:

> Multiple annotations of either form may be used and are cumulative, i.e. the list of k-v pairs will accumulate. If a given key appears more than once, then the latest value will replace all earlier ones.

Since this is in the spec, we should probably also do that.

2. I would like to propose that we treat KV lists and expression lists more uniform, since the two can be mixed (at least through accumulation, though the P4 spec seems to forbid mixing in a single annotation).  We can treat expression lists as a KV list where they key is just empty.

3. The P4 spec says annotation names starting with a lowercase character are reserved, so at the very least we should not use that in our examples.

----------------

Here's my proposal (I also change annotation_tag to name, since that's what the P4 spec uses):

```
// One parsed element within an annotation's body.
message AnnotationElement {
  // key can be empty, in which case the annotation was just an expression
  string key = 1;
  string value = 2;
}

// An ordered list of annotations identified by the name
message ParsedAnnotation {
  string name = 1;
  repeated AnnotationElement elements = 2;
}

// to be used as:

repeated ParsedAnnotation parsed_annotations = ...;
```

The following invariants hold:

1. For any P4 entity with annotations, there are no two `parsed_annotations` that have the same `name`.
2. Within a `ParsedAnnotation`, there are no two `AnnotationElement`s that have the same `key`, except for the empty key, which can appear any number of times.

Here's an example of how this would be used (showing accumulation and shadowing and mixing):

```
@MyAnno1()
@MyAnno2("Some random content: for i in range(10)")
@MyAnno3(1+1, "Hello", 1 << 3)
@MyAnno4(a=b,c=d)
@MyAnno4(e=f)
@MyAnno5(a=1,a=2)
@MyAnno6(e=f)
@MyAnno6(1)
table test1 {
    key = { hdrs.data.f1 : ternary; }
    actions = {
        setb1;
        noop;
    }
    default_action = noop;
}

tables {
  preamble {
    id: 33619894
    name: "ingress.test1"
    alias: "test1"
    annotations: "@MyAnno1()"
    annotations: "@MyAnno2("Some random content: for i in range(10)")"
    annotations: "@MyAnno3(1+1, "Hello", 1 << 3)"
    annotations: "@MyAnno4(a=b,c=d)"
    annotations: "@MyAnno4(e=f)"
    annotations: "@MyAnno5(a=1,a=2)"
    annotations: "@MyAnno6(a=b)"
    annotations: "@MyAnno6(1)"
    parsed_annotations {
      name: "MyAnno1"
    }
    parsed_annotations {
      name: "MyAnno2"
      elements {
        value: "\"Some random content: for i in range(10)\""
      }
    }
    parsed_annotations {
      name: "MyAnno3"
      elements {
        value: "1+1"
      }
      elements {
        value: "\"Hello\""
      }
      elements {
        value: "1 << 3"
      }
    }
    parsed_annotations {
      name: "MyAnno4"
      elements {
        key: "a"
        value: "b"
      }
      elements {
        key: "c"
        value: "d"
      }
      elements {
        key: "e"
        value: "f"
      }
    }
    parsed_annotations {
      name: "MyAnno5"
      elements {
        key: "a"
        value: "2"
      }
    }
    parsed_annotations {
      name: "MyAnno6"
      elements {
        key: "a"
        value: "b"
      }
      elements {
        value: "1"
      }
    }
  }
}
```

Thoughts?

chrispsommers:
@antoninbas and @stefanheule Great input, I like Stefan's revised proposal, thanks for the improvements and providing a well-worked-out example. It's a good compromise on all the issues so far raised - fewest assumptions,  mostly-reversible transformations, reasonably convenient. It would take just a few lines of e.g. Python code to build a consolidated dict, for example. I would like to see feedback from @konne-google if possible, since he was strongly in favor of the need for something like this, based on actual use-cases.

I'll wait for this to simmer awhile, then submit a new PR prior to the next WG meeting. @antoninbas thanks for the reminder to comb the spec for affected areas.

docs/v1/P4Runtime-Spec.mdk
stefanheule: Thinking about this some more, this is at least my interpretation of what the P4 spec says.  However, we may want to ask if that's correct, and why there is this inconsistency between accumulating expression lists and KV lists, but not allowing mixing of them directly.

Another thing we might want to check is if the compiler currently accumulates/overwrites as is described in the P4 spec.  If it does not, we should ask why.  If it does, a P4RT implementation could reuse that when creating the proto.

chrispsommers:
@stefanheule 

> Thinking about this some more, this is at least my interpretation of what the P4 spec says. However, we may want to ask if that's correct, and why there is this inconsistency between accumulating expression lists and KV lists, but not allowing mixing of them directly.

I added kv-pairs to the language spec in https://github.com/p4lang/p4-spec/commit/ede4c872875ad2b878c2b434a5551482fad0e40e. I didn't think of these other cases at the time. If we do approve this PR, I'll bring suggestions to the LDWG to enhance the grammar accordingly.

>Another thing we might want to check is if the compiler currently accumulates/overwrites as is described in the P4 spec. If it does not, we should ask why. If it does, a P4RT implementation could reuse that when creating the proto.

 `p4c` does the kv-pair accumulation correctly (I believe there are even test cases for it), but no expression accumulation.

stefanheule:
Ah great, I didn't know you also did this in the P4 spec! :)  For p4c, maybe we should open a bug then to track that it doesn't implement the accumulation for expressions.  Does it shadow KVs correctly?  That is, does @A(a=1,a=2) become @A(a=2) like the spec says?

chrispsommers:
@stefanheule The spec currently only covers the accumulated kv-pairs and `p4c` already handles it. The PR was quite some time ago. If we approve this change to P4RT, I'll open issues for enhancing the language and compiler.

stefanheule:
I'm not sure I understand what you are saying, I think the spec does explicitly also cover shadowing: "If a given key appears more than once, then the latest value will replace all earlier ones.".  We should make sure the compiler implements this, or if not, open a bug.  But in any case, sounds like we are all in agreement, thanks for the explanations and all your work!

konne-google:
Thanks guys for the very thorough work here, looks very good to me!

One minor nitpick, the comment:
// key can be empty, in which case the annotation was just an expression
was hard for me to understand, because annotations aren't expressions. 

Maybe:
// Key is empty, if and only if no key was specified

chrispsommers:
@konne-google Actually annotations *are* expression lists or kv-lists. From https://p4.org/p4-spec/docs/P4-16-v1.1.0-spec.html#sec-annotations:
```
annotation
    : '@' name
    | '@' name '(' expressionList ')'
    | '@' name '(' keyValueList ')'
    ;
```
We'll probably want to enhance the grammar to allow arbitrary mixtures of expressions and kv-pairs.

stefanheule:
You are right Chris that annotations can use expressions in their element list (whether it's a KV list or just expressions without a key), but annotations themselves are not expressions.  That's why Konstantin thinks my sentence is confusing, and I agree that his proposal is better.  Let's use that (minus the stray comma).

`// Key is empty if and only if no key was specified`

liujed:
As a heads-up, there are some changes to P4 annotations coming down the pipe. With the next version of the spec, annotation bodies will be largely unstructured: https://github.com/p4lang/p4-spec/blob/c357d17bd8db3f3f3c3a71f6611062d2ae461ff4/p4-16/spec/P4-16-spec.mdk#L6954-L6979

```
annotation
    : '@' name
    | '@' name '(' annotationBody ')'
    ;
~ End P4Grammar

## Bodies of annotations

The flexibility of P4 annotations comes from the minimal structure mandated by
the P4 grammar: annotation bodies may contain any sequence of terminals, so
long as parentheses are balanced. In the following grammar fragment, the
`annotationToken` non-terminal represents any terminal produced by the lexer,
including keywords, identifiers, string and integer literals, and symbols, but
excluding parentheses.

~ Begin P4Grammar
annotationBody
    : /* empty */
    | annotationBody '(' annotationBody ')'
    | annotationBody annotationToken
~ End P4Grammar

Annotations may impose additional structure on their bodies, and are not
confined to the P4 language. For example, the P4Runtime specification defines a
`@pkginfo` annotation that takes key-value pairs, which are not found in the
main P4 language.
```

chrispsommers:
@liujed Thanks for flagging this. I do recall  you making proposals in the past towards this end. The intent of this PR is to make it easy to consume typical annotations like KV-pairs and expressions. I'm wondering how we can still accomplish this PR's intent without tripping on unstructured annotations. Do folks think there should be a new choice in the message below, e.g. `unstructured` for anything which doesn't match `expression` or `property` grammar?
```  oneof element {
    string expression = 1;
    AnnotationProperty property = 2;
  }
 ```

**Comment by chrispsommers** [comment]
OK folks, I updated `p4types.proto`, `p4info.proto` and the spec, to align with the very latest suggestions, just prior to @liujed 's https://github.com/p4lang/p4runtime/pull/238#issuecomment-536092720. I realize this is evolving but I wanted to consolidate it so we have a  clear view of where we are. I reexamined the history of the P4 spec and `p4c` commits in this regard, you might refer to https://github.com/p4lang/p4-spec/pull/724 and https://github.com/p4lang/p4c/pull/1614. I think this PR will work with free-form annotations as developed by Jed. Note, in the latest commit I added `@MyAnno7(...)` borrowed from Jed's example in https://github.com/p4lang/p4c/pull/1614/files#diff-8e279f4569ce347a6bf5278b97c851bc (file: testdata/p4_16_samples_outputs/free-form-annotation-first.p4)

@liujed I see you've made quite a few additional commits to `p4c` to handle various annotation body use-cases. I'm wondering if P4Info needs to accommodate these cases, e.g. via enhancements to the proposed `ParsedAnnotation` message. Are we missing something?

**Comment by liujed** [comment]
The proposed `ParsedAnnotation` looks mostly there, but I have a couple of concerns.

First, it appears to be missing the ability to distinguish between annotations that are recognized by the compiler, and ones that are not. I don't know if this is important in this context, but p4c has different internal representations for `@MyAnno3("Hello", 1 << 3, 1 + 1)`, depending on whether it recognizes (and parses) `MyAnno3` annotations, analogous to the difference between
```
    parsed_annotations {
      name: "@MyAnno3"
      elements {
        value: "\"Hello\""
      }
      elements {
        value: "1 << 3"
      }
      elements {
        value: "1 + 1"
      }
    }
```
and
```
    parsed_annotations {
      name: "@MyAnno3"
      elements {
        value: "\"Hello\", 1 << 3, 1 + 1"
      }
    }
```
In the former case, the annotation body is parsed into a list of expressions; in the latter case, the body is just an uninterpreted sequence of tokens. With the current proposal in this PR, these two cases are indistinguishable for an annotation like `@AmIInterpreted(1 + 1)`. Both cases would result in
```
    parsed_annotations {
      name: "@AmIInterpreted"
      elements {
        value: "1 + 1"
      }
    }
```
I can imagine a few ways of addressing this, if this is indeed a concern. My suggestion would be to not deprecate the `annotations` element; only annotations that are recognized and parsed by the compiler would have a corresponding entry in `parsed_annotations`. This is the smallest and cleanest change I can think of.

A second, perhaps lesser concern, is that I'd be careful to preserve whitespace in uninterpreted annotations. The annotation in the [example](https://github.com/p4lang/p4runtime/pull/238/files#diff-73eaf31ba21b55b225f937a3975df025R1073-R1080) appears as follows:
```
@MyAnno7(
- What do you get if you multiply six by nine?
- Six by nine. Forty two.
- That's it. That's all there is.
- I always thought there was something fundamentally wrong with the
  universe.
0xdeadbeef
)
```
I would expect the following result (perhaps with `\n` instead of actual new-lines):
```
    annotations: "@MyAnno7(
- What do you get if you multiply six by nine?
- Six by nine. Forty two.
- That\'s it. That\'s all there is.
- I always thought there was something fundamentally wrong with the
  universe.
0xdeadbeef
")
```
and
```
    parsed_annotations {
      name: "@MyAnno7"
      elements {
        value: "
- What do you get if you multiply six by nine?
- Six by nine. Forty two.
- That\'s it. That\'s all there is.
- I always thought there was something fundamentally wrong with the
  universe.
0xdeadbeef
"
      }
    }
```

chrispsommers:
@liujed Thanks for the detailed feedback. I agree with your second point - preserve whitespace. And in my example, I substituted hard linebreaks for \n for readability, perhaps I should put in '\n' before the typographical linebreaks for clarity.

I also now feel that we should not deprecate the old-style annotations. As yourself and @stefanheule  both pointed out in different ways, we should avoid "lossy" transformations and making too many assumptions about how annotations will be used. 

You wrote:
>My suggestion would be to not deprecate the annotations element; only annotations that are recognized and parsed by the compiler would have a corresponding entry in parsed_annotations. 

When I first read this I interpreted "recognized and parsed by the compiler" to mean "has a well-known/reserved name," from the list of `p4c` reserved annotations in the P4-spec Section 18 (ex. : "@name"), or from the list of `P4info` reserved annotations, Table A.1 in the P4Runtime spec (ex.: "@brief").

Then I realized you probalby were paraphrasing my language in https://github.com/p4lang/p4runtime/pull/238#issuecomment-536104501:
>  If an unstructured annotation doesn't contain a conforming AnnotationElement then it won't generate any.

This brings up a point. I believe `p4c` already suppresses outputting recognized annotations from either reserved list into the `annotations` fields. However, I'm not sure where this is documented. If so, can someone point me to it? I think it deserves mention in the P4Runtime spec. In any case, the language describing the generation of parsed_annotations need some clarification.

To summarize, I propose that I do the following:
* Remove deprecations on old annotations - too many gotchas
* Clarify that only non-reserved annotations generate `annotations` and `parsed_annotations`
* Stipulate the preservation of whitespace and linebreaks
* Clarify that P4Info doesn't emit reserved annotations in `annotations` or `parsed_annotations`

stefanheule:
When I last commented I wasn't aware that P4 changed the grammar for annotations.  I think the new unstructured P4 annotations are a big step backwards, but it looks like it was already decided to go with that new proposal.

If we have to live with these new unstructured annotations, though, I don't think this PR makes much sense any more.  At most we can parse the name and the completely unstructured list of tokens, but I don't know how much value there is in that.  If we think there is, we can do it, but otherwise it might be easiest to just not do anything.

chrispsommers:
@stefanheule My understanding is there can still be parsing of expressions and kv-pairs for purposes of generating `P4Info`. The language spec is simply removing any grammar requirements for the body of  an annotation, i.e. anything inside the matching `()` is legal. The P4Runtime spec, not the P4-16 Language spec, controls the generation of P4Info and thus can declare, for a certain annotation format, it will indeed be parsed as stated in the PR. `p4c` strives to accommodate both the language requirements and P4Info generation. Anyone agree/disagree?

antoninbas:
@chrispsommers I'm assuming we can close this old PR now?

chrispsommers:
@antoninbas yes, this is obsolete. Thanks.

PR #239, opened 2019-09-21
Add linter for P4Runtime Madoko spec

antoninbas:
It is quite a primitive tool at the moment but it makes sure that:
 * there is no trailing whitespace
 * lines longer than 80 chars are wrapped
 * `&eg;` and `&ie;` are used


antoninbas:
I wrote it specifically for this. You know, just some holiday fun :)

Issue #232, opened 2019-08-20
Resetting counters of const table entries


konne-google:
How do I reset the counters of a const table entry? It sounds like P4RT doesn't support this. If so, is there a reason why? Can we support it. The relevant spec quotes:

"The only write updates which are allowed for constant tables are MODIFY operations on the default action"

"we assume that constant tables cannot be assigned direct resources"

konne-google:
Andy & Antonin, do you remember why this is what we chose?

antoninbas:
We had a very short discussion about it there: https://github.com/p4lang/p4-spec/issues/723
We can revisit the topic at the next WG meeting. I think the reason why we chose this is that 1) we didn't have a use case at the time (but it seems it is no longer the case since you are trying to use a direct counter with a constant table), and 2) it seemed non-trivial to define good semantics for reading / writing direct resource entries using the `TableEntry` message (`DirectCounterEntry` and `DirectMeterEntry` are more straightforward).

The current spec states that

> The `TableEntry` `action` field must be set for every `INSERT` and `MODIFY` update, except when resetting the default entry

so when trying to reset a counter with a `MODIFY` update using a `TableEntry` message, the client would need to provide the full & correct action specification (based on the P4 table declaration).

Personally, I don't see any particular issue with supporting resetting the counters of a const table using the `DirectCounterEntry`. If we want to support `TableEntry` messages as well, it may be trickier.

antoninbas:
This was discussed at the 08/28/2019 P4 API WG meeting (https://docs.google.com/document/d/1qzeAnKDDOIgBzq3sLS1C1CacunTtrV5x5T5xPXUFRGM/edit#heading=h.1luy7m2zeqb), and we decided to include support for this in P4Runtime v1.1.0.

PR #240, opened 2019-09-25
Add support for modifying direct resources of const tables

konne-google:


jafingerhut:
The changes look reasonable to me.

It might be helpful to have an example of a request that modifies the direct resources of such a table.

In particular, it seems that such a MODIFY message must contain the field match criteria, action, and action parameters.

Should it be an error if the field match criteria do not match any of the entries of the table?

Should it be an error if the field match criteria do match an entry of the table, but the action and/or action parameters do not match the action in the table?

jafingerhut:
Thinking a bit more about my previous comment, I guess MODIFY should already fail for any table if the field match criteria does not correspond to an entry in the table, regardless of whether it is const entries or not?

If so, that might be good to remind a read here explicitly.

It does seem "new" for const entries tables that it is also an error if the action and/or action parameters are different than what is already in the entry.

konne-google:
jafingerhut, I clarified which match/action fields need to be specified.

jafingerhut:
Looks like a good addition.  Thanks.  Should it go slightly further and specify that the server must return an error code of a particular name if the client does include an action or action parameters in the MODIFY request?

konne-google:
Andy, I think it's already specified what the switch should do, namely, if you try to modify anything but the direct resources, you'll get a PERMISSION_DENIED_ERROR. I've added a sentence to call that out.

jafingerhut:
Konstantin, sorry, I do not know how I could have missed that.  Thanks for the additional sentence.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2750,16 +2750,23 @@ hope that it would simplify the implementation of the P4Runtime service.
...
-the server must return a `PERMISSION_DENIED` error. However, the contents of
-such tables can be queried by the client through a `ReadRequest`. When reading
-static (immutable) entries from a constant table, the following fields --- and
-only these fields ---, must be set by the server: `table_id`, `match`, `action`,
-`priority` and `is_default_action`. In particular, we assume that constant
-tables cannot be assigned direct resources and idle timeout is not supported for
+are identified by the `is_const_table` flag in P4Info.
+
+The only write updates which are allowed for constant tables are `MODIFY`
+operations on direct resources; and the default action, assuming the default
```
antoninbas: semi-colon?

konne-google: use parentheses instead

konne-google:
> The following sentence in section 9.1.2 must be updated:
> 
> > The TableEntry action field must be set for every INSERT and MODIFY update, except when resetting the default entry.
> 
> We agreed (because of static entries) that `action` would no longer be required for MODIFY. So the sentence should read something like this:
> 
> > The TableEntry action field must be set for every INSERT update but may be left unset for MODIFY updates, in which case the action specification for the table entry will not be modified (if a MODIFY update has an unset action field and does not modify any direct resource attached to the table then the MODIFY update is a no-op).
> 
> Later on in the same section, this sentence should also be updated:
> 
> > If the action field is not set (and if is_default_action is false) or if the oneof does not match the table description in the P4Info (e.g. the oneof is action_profile_member_id for a direct table), the server must return an INVALID_ARGUMENT error code.

Thanks for spotting these, and the suggested updates!

antoninbas:
Fixes #232

PR #242, opened 2019-10-08
Add CONTRIBUTING.md file

antoninbas:
Also add a check to the Madoko lint tool to make sure that a small
subset of "keywords" (can be expanded later) are highlighted with
backticks.

### DIFF
--- a/tools/madokolint.py
+++ b/tools/madokolint.py
@@ -26,17 +26,22 @@
 
 import argparse
 from collections import namedtuple
+import json
 import os.path
 import re
 import sys
+import traceback
 
 
+DEFAULT_CONF = 'madokolint.conf.json'
 LINE_WRAP_LENGTH = 80
 
 
 parser = argparse.ArgumentParser(description='Lint tool for Madoko code')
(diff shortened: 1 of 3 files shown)


tools/madokolint.py
```diff
@@ -216,10 +260,24 @@ def check(line, lineno):
...
 
+def check_keywords(path):
+    def check(line, lineno):
+        for word in line.split():
+            if word not in lint_conf.keywords:
+                continue
+            category = lint_conf.keywords[word]
+            lint_state.error(
+                path, lineno, line,
+                "'{}' is a known keyword ({}), highlight it with backticks".format(word, category))
```
chrispsommers: Out of curiosity, does this actually verify backticks surround a keyword, or does it just detect "naked" keywords by virtue of the `split()` behavior?

antoninbas: Just looks for naked keywords at the moment, so `'keyword'` will not trigger an error. I'm sure there is room for improvement :)

chrispsommers: OK, I was just verifying my understandinig, It's still valuable!

PR #248, opened 2019-10-24
Clarify ordering requirements.

### LINKED ISSUE #245: Clarify / fix semantics of Write RPC
* batches should have the serializability property (http://www.bailis.org/blog/linearizability-versus-serializability/)
 * remove sentence about idempotency from spec since it is inaccurate

See minutes from the 10/23/2019 P4 API WG meeting for more details: https://docs.google.com/document/d/1qzeAnKDDOIgBzq3sLS1C1CacunTtrV5x5T5xPXUFRGM/edit#heading=h.fag17iz3x4wo

stefanheule:
Fixes #245.

Note that I also reworded/shortened some of the "advisory" text.  I think I preserved what we wanted to say before, but removed things I don't think we ever intended to say, and removed things that sounded overly prescriptive.  For instance, before we had a sentence:

> In other words, the client must wait until the dependent Write RPC is acknowledged before invoking a Write RPC that depends on it.

I don't think a client *must* do that.  It's certainly _advisable_ if the client wants deterministic (and useful) behavior from the switch, but they are not *required* to do that.  That's an important distinction.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -4383,7 +4383,7 @@ The P4Runtime server may choose to reorder updates in a batch when processing
 them, and/or process updates in parallel.  Updates across multiple concurrent
 `WriteRequest`s can also be processed interleaved and/or in parallel.
 However, **the processing of updates must be strictly serializable**.  That
-is, given a set $S$ of `WriteRequest`s including their responses to those
+is, given a history $S$ of `WriteRequest`s including the responses to those
 requests, there must exist an order $L$ for all updates in $S$, such that:
 
 1. For two updates $u_1$ and $u_2$, if the write request containing $u_1$
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -4390,7 +4390,7 @@ must exist an order L for all updates in S, such that:
    before the write request of u2 was sent, then u1 must appear before u2 in L.
 2. Executing all updates in L sequentially must yield the same response for
    every update as in S.
-3. The observable state of the switch after S (e.g., through the `Read` RPC) is
+3. The observable state of the switch after S (&eg;, through the `Read` RPC) is
    identical to the one obtained by sequentially executing L.
 
 The `Write` RPC demarcates the batch boundary, and can be used to ensure
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -4383,15 +4383,16 @@ The P4Runtime server may choose to reorder updates in a batch when processing
 them, and/or process updates in parallel.  Updates across multiple concurrent
 `WriteRequest`s can also be processed interleaved and/or in parallel.
 However, **the processing of updates must be strictly serializable**.  That
-is, given a set S of `WriteRequest`s and the responses to those requests, there
-must exist an order L for all updates in S, such that:
+is, given a set $S$ of `WriteRequest`s including their responses to those
+requests, there must exist an order $L$ for all updates in $S$, such that:
 
-1. For two updates u1 and u2, if the write request containing u1 completed
-   before the write request of u2 was sent, then u1 must appear before u2 in L.
-2. Executing all updates in L sequentially must yield the same response for
-   every update as in S.
-3. The observable state of the switch after S (&eg;, through the `Read` RPC) is
-   identical to the one obtained by sequentially executing L.
+1. For two updates $u_1$ and $u_2$, if the write request containing $u_1$
+   completed before the write request of $u_2$ was sent, then $u_1$ must appear
+   before $u_2$ in $L$.
+2. Executing all updates in $L$ sequentially must yield the same response for
+   every update as in $S$.
+3. The observable state of the switch after $S$ (&eg;, through the `Read` RPC)
+   is identical to the one obtained by sequentially executing $L$.
 
 The `Write` RPC demarcates the batch boundary, and can be used to ensure
 ordering between dependent updates. When the `Write` RPC returns, it is required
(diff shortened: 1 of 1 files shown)


jafingerhut:
Instead of saying this: "In other words, the client must wait until the dependent Write RPC is acknowledged before invoking a Write RPC that depends on it."

Would you consider this more accurate? "In other words, if the client wishes to guarantee that Write RPC B does not begin processing on the server until after the last operation of Write RPC A has completed, the client must wait until Write RPC A has been acknowledged before invoking Write RPC B."

That is, it is not required, but as far as I can tell from the rest of the P4runtime specification, it _is_ necessary in order to achieve that stated goal.  No other mechanism seems to be available for a client to guarantee the "if" condition.

stefanheule:
That's right, and I tried to say that with "For this reason, most clients are advised to split dependent
updates across separate, non-overlapping `Write` calls."  but maybe that's not clear enough.  Would

> For this reason, most clients are advised to split dependent updates across separate `Write` calls.  Additionally, each `Write` call has to be sent sequentially, waiting for the previous call to be acknowledged before sending the next one.

be better?  I can also use your sentence, but with the changes I made it no longer has an obvious place it would go.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4385,31 +4379,32 @@ P4Runtime supports batching of `Write` operations. The list of updates in a
...
-The P4Runtime server may arbitrarily reorder message within a batch to maximize
-performance, and clients should not depend on a specific processing order (&eg;
-FIFO or inferring implicit dependencies within a batch). In particular, P4
-entities (&eg; table entries) may be inserted in the data plane in an order
-different than what is received in the `WriteRequest`.
+The P4Runtime server may choose to reorder updates in a batch when processing
+them, and/or process updates in parallel.  Updates across multiple concurrent
+`WriteRequest`s can also be processed interleaved and/or in parallel.
+However, **the processing of updates must be strictly serializable**.  That
+is, given a set S of `WriteRequest`s and the responses to those requests, there
```
antoninbas: I think it will look better if you typeset S, L, u1 and u2 between `$` characters

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4385,31 +4379,32 @@ P4Runtime supports batching of `Write` operations. The list of updates in a
...
+must exist an order L for all updates in S, such that:
+
+1. For two updates u1 and u2, if the write request containing u1 completed
+   before the write request of u2 was sent, then u1 must appear before u2 in L.
+2. Executing all updates in L sequentially must yield the same response for
+   every update as in S.
```
antoninbas: I think "must yield the same response for every update as executing S." is clearer?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4385,31 +4379,32 @@ P4Runtime supports batching of `Write` operations. The list of updates in a
...
+3. The observable state of the switch after S (&eg;, through the `Read` RPC) is
```
antoninbas: suggestion: "after *executing* S"

docs/v1/P4Runtime-Spec.mdk

docs/v1/P4Runtime-Spec.mdk
stefanheule: S is meant to be the set of updates + their response, so you don't really need to execute S.  Tried to clarify that when introducing S.

docs/v1/P4Runtime-Spec.mdk
stefanheule: See above.

stefanheule:
> That's right, and I tried to say that with "For this reason, most clients are advised to split dependent
> updates across separate, non-overlapping `Write` calls." but maybe that's not clear enough. Would
> 
> > For this reason, most clients are advised to split dependent updates across separate `Write` calls.  Additionally, each `Write` call has to be sent sequentially, waiting for the previous call to be acknowledged before sending the next one.
> 
> be better? I can also use your sentence, but with the changes I made it no longer has an obvious place it would go.

I added the more explicit version.  Let me know if this is not sufficient.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4385,31 +4379,35 @@ P4Runtime supports batching of `Write` operations. The list of updates in a
...
-performance, and clients should not depend on a specific processing order (&eg;
-FIFO or inferring implicit dependencies within a batch). In particular, P4
-entities (&eg; table entries) may be inserted in the data plane in an order
-different than what is received in the `WriteRequest`.
+The P4Runtime server may choose to reorder updates in a batch when processing
+them, and/or process updates in parallel.  Updates across multiple concurrent
+`WriteRequest`s can also be processed interleaved and/or in parallel.
+However, **the processing of updates must be strictly serializable**.  That
+is, given a set $S$ of `WriteRequest`s including their responses to those
+requests, there must exist an order $L$ for all updates in $S$, such that:
```
antoninbas: sorry, I missed that you were including the responses in `S`. However, I don't think that "including their responses to those requests" makes sense. What is "theirs" referring to? Seems that it should be either "including their responses" or "including the responses to those requests".

jafingerhut:
> That's right, and I tried to say that with "For this reason, most clients are advised to split dependent
> updates across separate, non-overlapping `Write` calls." but maybe that's not clear enough. Would
> 
> > For this reason, most clients are advised to split dependent updates across separate `Write` calls.  Additionally, each `Write` call has to be sent sequentially, waiting for the previous call to be acknowledged before sending the next one.
> 
> be better? I can also use your sentence, but with the changes I made it no longer has an obvious place it would go.

Your suggested replacement does seem a little better to me, if you wouldn't mind using it.  I'm not in love with my sentence -- I wrote it primarily to see if we were on the same page there, so all good.

docs/v1/P4Runtime-Spec.mdk
stefanheule: Agreed, fixed.

konne-google:
Thanks Stefan, this makes it much clearer what the semantics are of P4RT! I think we should submit this.

PR #251, opened 2019-11-09
Add Capabilities RPC to P4Runtime service

### LINKED ISSUE #249: Consider adding Capabilities RPC for P4Runtime v1.1.0
Similar to the one in gNMI: https://github.com/openconfig/gnmi/blob/master/proto/gnmi/gnmi.proto#L49

This would enable a P4Runtime client to determine which version is currently supported by the server for a given major version number.

```
service P4Runtime {
  // ...
  rpc Capabilities(CapabilitiesRequest) returns CapabilitiesResponse;
}

message CapabilityRequest { }

message CapabilitiesResponse {
  string semver = 1;
  // potentially other fields in the future
}
```

Moving forward this would enable the client to determine whether some features introduced by minor version releases are actually supported by the server.

In  the future the `Capabilities` RPC could be useful to provide additional information to the client. See https://github.com/p4lang/p4runtime/issues/213.

The earlier this RPC is added (even in a very simple form as proposed above), the better. Otherwise it doesn't provide a reliable mechanism for the client to determine the P4Runtime version supported by the server, since the RPC itself may not be supported by the service (in which case the client would get an `UNIMPLEMENTED` error). This is why I would be in favor of adding this RPC in the upcoming 1.1.0 release.

Thoughts?

antoninbas:
The P4Runtime API WG decided to include this to P4Runtime v1.1. For now,
this RPC will only enable the client to query the version of the
P4Runtime specification implemented by the server.

Fixes #249

### DIFF
--- a/proto/p4/v1/p4runtime.proto
+++ b/proto/p4/v1/p4runtime.proto
@@ -751,6 +751,6 @@ message CapabilitiesRequest {
 
 message CapabilitiesResponse {
   // The full semantic version string (e.g. "1.1.0-rc.1") corresponding to the
-  // version of the P4Runtime specification currently implemented by the server.
-  string p4runtime_version = 1;
+  // version of the P4Runtime API currently implemented by the server.
+  string p4runtime_api_version = 1;
 }
--- a/proto/p4/v1/p4runtime.proto
+++ b/proto/p4/v1/p4runtime.proto
@@ -53,6 +53,9 @@ service P4Runtime {
   rpc StreamChannel(stream StreamMessageRequest)
       returns (stream StreamMessageResponse) {
   }
+
+  rpc Capabilities(CapabilitiesRequest) returns (CapabilitiesResponse) {
+  }
 }
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


chrispsommers:
LGTM. How about a reference to https://semver.org/ in the biblio and text?

proto/p4/v1/p4runtime.proto
```diff
@@ -741,3 +744,13 @@ enum SdnPort {
...
 }
+
+//------------------------------------------------------------------------------
+message CapabilitiesRequest {
+}
+
+message CapabilitiesResponse {
+  // The full semantic version string (e.g. "1.1.0-rc.1") corresponding to the
+  // version of the P4Runtime specification currently implemented by the server.
+  string p4runtime_version = 1;
```
stefanheule: This sounds like it's the version of the p4runtime server (which presumably also has a version).  Would p4runtime_language_version or p4runtime_protocol_version be better?

antoninbas: How about `p4runtime_api_version` since it is shorter and matches the terminology used in this webpage that we reference: https://cloud.google.com/apis/design/versioning

antoninbas:
@chrispsommers There is a reference already (line 5177 of the current diff)

proto/p4/v1/p4runtime.proto
stefanheule: That's an even better suggestion than mine, I like it. :)

antoninbas: Done.

PR #254, opened 2019-11-29
Prepare release 1.1.0-rc.1

antoninbas:
* Update references to P4_16 spec from 1.1.0 to 1.2.0
* Use 1.1 / 1.1.0 when referring to the current P4Runtime version
* Update version to 1.1.0-rc.1 for release

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -230,7 +230,8 @@ format and available at
 P4Runtime is designed to be implemented in conjunction with the P4~16~ language
 version or later. P4~14~ programs should be translated into P4~16~ to be made
 compatible with P4Runtime. This version of P4Runtime utilizes features which are
-not in P4~16~ 1.0, but were introduced in P4~16~ 1.1.0 [@P4Spec].
+not in P4~16~ 1.0, but were introduced in P4~16~ 1.1.0 [@P4Revisions110]. For
+this version of P4Runtime, we recommend using P4~16~ 1.2.0 [@P4Spec].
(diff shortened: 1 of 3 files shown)


antoninbas:
@jafingerhut I would appreciate it if you could take a quick look at the changes I made to your document (I didn't change much per se, just the version number for P4_16 / P4Runtime).

antoninbas:
@chrispsommers I made the edit you suggested in the specification document (this was the only occurrence in the doc where the current version number was hardcoded). For the P4Info generation document, I'll defer to @jafingerhut. Andy, do you think we should remove hardcoded references to the current version number there, or do you think they help? (Maybe keeping the version numbers there makes it easier to detect when the doc is no longer in sync with the spec?).

jafingerhut:
@antoninbas My intent on including version numbers in the file `guidance-for-generating-p4info.md` was to make it clear what versions of other specifications it was checked for.  Without any version numbers at all in that file, all it would take is for another one of those references to be updated with a new versioned release, and it is no longer clear to me whether that document has been checked against those versions, or not.

It does seem reasonable to mention the versions of other documents it has been checked against in one place, near the beginning of the file, and not keep repeating version numbers throughout, if that would make things easier to maintain in the future.

antoninbas:
@jafingerhut Sounds good. I don't think there is any real benefit in changing the way `guidance-for-generating-p4info.md` does things at this stage. The document needs to be updated for every release any way, so the version numbers themselves are not a maintenance burden IMO (they actually make it easier to figure out what needs to be updated in the doc for each P4Runtime release).

PR #261, opened 2020-01-29
Fix inconsistency in spec regarding Write consistency guarantees

### LINKED ISSUE #258: Inconsistency in P4Runtime spec regarding consistency guarantees
Section 13.4 (Parallelism of Read and Write Requests) includes the following:
> The processing of any two WriteRequest messages W1 and W2 must result in the same state as if one of them was completed before the other began.

The more recently-updated section 12.1 (Batching and Ordering of Updates) states the following:

The P4Runtime server may choose to reorder updates in a batch when processing them, and/or process updates in parallel. Updates across multiple concurrent WriteRequests can also be processed interleaved and/or in parallel. However, the processing of updates must be strictly serializable. That is, given a history `S` of WriteRequests including the responses to those requests, there must exist an order `L` for all updates in `S`, such that:
1. For two updates `u_1` and `u_2`, if the write request containing `u_1` completed before the write request of `u_2` was sent, then `u_1` must appear before `u_2` in `L`.
2. Executing all updates in `L` sequentially must yield the same response for every update as in `S`.
3. The observable state of the switch after `S` (e.g., through the Read RPC) is identical to the one obtained by sequentially executing `L`.

It seems that the 2 sections are a bit in contradiction. Consider a data plane with a TCAM table `T`, which is at capacity (cannot hold an additional entry). Consider also 2 WriteRequests `W1` and `W2` such that:
* `W1` deletes 2 exiting entries from table `T`: `E1` and `E2`
* `W2` inserts 2 new entries into table `T`: `E3` and `E4`.

If there is no strong ordering of both WriteRequests (i.e. they are sent by the client in parallel and bullet point 1. above does not apply), then it seems that section 12.1 states that the server is allowed to process all 4 updates in the following order: `delete(E1)`, `insert(E3)`, `insert(E4)`, `delete(E2)`, while section 13.4 prevents that kind of interleaving.

State of table `T` after both WriteRequests have been processed:
a. `W1` then `W2`: `T` is at capacity, and contains `E3` and `E4`
b. `W2` then `W1`: `T` has room for 2 new entries, and contains neither `E3` not `E4`
c. `delete(E1)`, `insert(E3)`, `insert(E4)`, `delete(E2)`: `T` has room for 1 new entry, and it contains `E3` but not `E4`.

Only a. and b. are acceptable under 13.4. All orders are acceptable according to 12.1.

antoninbas:
This updates section 12.1 (Batching and Ordering of Updates) to match
the stronger guarantees in section 13.4 (Parallelism of Read and Write
Requests).

Fixes #258

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -4383,16 +4383,18 @@ entity or table (in the case of `TableEntry` entities).
 The P4Runtime server may choose to reorder updates in a batch when processing
 them, and/or process updates in parallel.  Updates across multiple concurrent
 `WriteRequest`s can also be processed interleaved and/or in parallel.
-However, **the processing of updates must be strictly serializable**.  That
+However, **the processing of requests must be strictly serializable**.  That
 is, given a history $S$ of `WriteRequest`s including the responses to those
-requests, there must exist an order $L$ for all requests in $S$, such that:
+requests, there must exist an order $L$ for all updates in $S$, such that:
 
-1. For two `WriteRequest`s $r_1$ and $r_2$, if $r_1$ completed before $r_2$ was
-   sent, then $r_1$ must appear before $r_2$ in $L$.
-2. Executing all requests in $L$ sequentially must yield the same response for
-   every request as in $S$, &ie; the status reported by the server for every
-   individual update in every request must be the same.
-3. The observable state of the switch after $S$ (&eg;, through the `Read` RPC)
+1. All updates from the same write request must appear as a contiguous
+   subsequence in $L$, but the order within that subsequence can be arbitrary.
+2. For two updates $u_1$ and $u_2$, if the write request containing $u_1$
+   completed before the write request of $u_2$ was sent, then $u_1$ must appear
+   before $u_2$ in $L$.
+3. Executing all updates in $L$ sequentially must yield the same response for
+   every update as in $S$.
+4. The observable state of the switch after $S$ (&eg;, through the `Read` RPC)
    is identical to the one obtained by sequentially executing $L$.
 
 The `Write` RPC demarcates the batch boundary, and can be used to ensure
@@ -4406,8 +4408,9 @@ clients can invoke multiple outstanding `Write` RPCs. If the updates across
 these RPCs have dependencies, the observed behavior may be non-deterministic, as
 the server can process these RPCs in any arbitrary order, providing strict
 serializability is enforced. For this reason, most clients are advised to split
-dependent updates across separate `Write` calls. Additionally, each `Write` call
-has to be sent sequentially, waiting for the previous call to be acknowledged
+dependent updates across separate `Write` calls. Additionally, if the client
+wants to enforce that batches are applied in a specific order, each `Write` call
+should be sent sequentially, waiting for the previous call to be acknowledged
 before sending the next one.
 
 
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -4385,29 +4385,30 @@ them, and/or process updates in parallel.  Updates across multiple concurrent
 `WriteRequest`s can also be processed interleaved and/or in parallel.
 However, **the processing of updates must be strictly serializable**.  That
 is, given a history $S$ of `WriteRequest`s including the responses to those
-requests, there must exist an order $L$ for all updates in $S$, such that:
+requests, there must exist an order $L$ for all requests in $S$, such that:
 
-1. For two updates $u_1$ and $u_2$, if the write request containing $u_1$
-   completed before the write request of $u_2$ was sent, then $u_1$ must appear
-   before $u_2$ in $L$.
-2. Executing all updates in $L$ sequentially must yield the same response for
-   every update as in $S$.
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4385,29 +4385,30 @@ them, and/or process updates in parallel.  Updates across multiple concurrent
 `WriteRequest`s can also be processed interleaved and/or in parallel.
 However, **the processing of updates must be strictly serializable**.  That
 is, given a history $S$ of `WriteRequest`s including the responses to those
-requests, there must exist an order $L$ for all updates in $S$, such that:
+requests, there must exist an order $L$ for all requests in $S$, such that:
```
stefanheule: By changing this to not talk about updates anymore, we lose the formal definition about what kind of reordering of updates are allowed. I think we should keep this at the update level and just add the necessary restriction.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4385,29 +4385,30 @@ them, and/or process updates in parallel.  Updates across multiple concurrent
...
 
-1. For two updates $u_1$ and $u_2$, if the write request containing $u_1$
-   completed before the write request of $u_2$ was sent, then $u_1$ must appear
-   before $u_2$ in $L$.
-2. Executing all updates in $L$ sequentially must yield the same response for
-   every update as in $S$.
+1. For two `WriteRequest`s $r_1$ and $r_2$, if $r_1$ completed before $r_2$ was
+   sent, then $r_1$ must appear before $r_2$ in $L$.
+2. Executing all requests in $L$ sequentially must yield the same response for
+   every request as in $S$, &ie; the status reported by the server for every
```
stefanheule: Another problem with doing this at the request level is that "Executing all requests in $L$ sequentially must yield the same response for every request as in $S$" is not really defined, what does it mean to execute a request sequentially in terms of the updates within a request?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4385,29 +4385,30 @@ them, and/or process updates in parallel.  Updates across multiple concurrent
...
+from the client depend on each other (&eg; inserting an `ActionProfileMember`
+followed by pointing a `TableEntry` to it) and the updates are not split into
+separate batches, then the behavior may be non-deterministic.  Similarly,
+clients can invoke multiple outstanding `Write` RPCs. If the updates across
+these RPCs have dependencies, the observed behavior may be non-deterministic, as
+the server can process these RPCs in any arbitrary order, providing strict
+serializability is enforced. For this reason, most clients are advised to split
+dependent updates across separate `Write` calls. Additionally, each `Write` call
+has to be sent sequentially, waiting for the previous call to be acknowledged
+before sending the next one.
```
stefanheule: I know you didn't write this last sentence here, but should make this less strong? "has to be sent sequentially" sounds like a requirement on clients, but I think it is only meant as a suggestion.

antoninbas: I rephrased, PTAL.

docs/v1/P4Runtime-Spec.mdk
antoninbas: That makes sense, even though I feel the definition becomes a bit too complicated compared to something like this meant for databases: http://www.bailis.org/blog/linearizability-versus-serializability/. I guess it makes sense in a way since DB transactions are usually assumed to execute operations in order, whereas we want updated within a request to be able to be processed in parallel.

antoninbas: I did change "the processing of updates must be strictly serializable" to "the processing of requests must be strictly serializable" though, as the first one is not accurate IMO.

jafingerhut: My mental P4runtime API specification glossary cache is fairly empty right now.  Looking to refresh it here :-)

Is the intent within the P4Runtime spec is that "update" refers to a single insert/update/delete operation, whereas request refers to an entire `WriteRequest` message?

jafingerhut: OK, my mental cache is getting filled again.  A request is a write request or read request, and write requests contain unordered sets of updates.  Still reading and thinking on these changes, but should be done soon.

antoninbas: this section mostly concerns itself with write requests, but I guess the statement "the processing of requests must be strictly serializable" holds even if requests == writes + reads.

yes update refers to an individual write operation in a write request batch, the name comes from the field name in p4runtime.proto

antoninbas:
Thanks @jafingerhut; I'll wait for @stefanheule's review as well before I merge, to ensure that I have addressed his comments.

docs/v1/P4Runtime-Spec.mdk
stefanheule: Sorry for the delay, I was sick last week.

Looks good, thanks for the updates. One last comment: Right now we say "there must exist an order $L$ for all requests in $S$", but I think $L$ is now again an order over updates, right? Should we revert that back to updates, or maybe "all updates of all requests"?

antoninbas: Yes, I definitely forgot to revert that one

Issue #264, opened 2020-02-11
Add structured annotations per https://github.com/p4lang/p4-spec/pull/796


chrispsommers:

antoninbas:
Support was added in the P4Runtime spec and in p4c

PR #265, opened 2020-02-11
Added Structured Annotations

chrispsommers:
This PR updates p4info.proto and p4types.proto as well as the P4Runtime specification to add structured annotations as described in https://github.com/p4lang/p4runtime/issues/264 and https://github.com/p4lang/p4-spec/pull/796. The P4_16 lang spec PR is not yet ratified, it is awaiting a practical implementation in `p4c` so one goal of this PR is to inform an implementation. Step one is to parse the syntax, step two is to generate the new `P4info`.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1042,7 +1042,7 @@ representations of the original annotation source. This parsing includes
 expression-evaluation, so the resulting `P4Info` may contain a simplified
 replica of the original structured annotations.
 
-The structured annotation messages are defined in `p4types.proto`.
+The structured annotation messages are defined in p4types.proto.
 
 ~ Begin Proto
 message KeyValuePair {
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2477,7 +2477,7 @@ the `TableEntry` entity, which has the following fields:
   match and with which argument values.
 
 * `priority`, a 32-bit integer used to order entries when the table's match key
-  includes a ternary or range match.
+  includes an optional, ternary or range match.
 
 * `controller_metadata`, a 64-bit cookie value which is opaque to the
   target. There is no requirement of where this is stored, but it must be
@@ -2502,7 +2502,8 @@ the `TableEntry` entity, which has the following fields:
 
 The `priority` field must be set to a non-zero value if the match key includes a
 ternary match (&ie; in the case of PSA if the P4Info entry for the table
-indicates that one or more of its match fields has a `TERNARY` or `RANGE` match
+indicates that one or more of its match fields has an `OPTIONAL`, `TERNARY` or
+`RANGE` match
 type) or to zero otherwise. A higher priority number indicates that the entry
 must be given higher priority when performing a table lookup. Clients must allow
 multiple entries to be added with the same priority value.  If a packet can
@@ -2545,6 +2546,8 @@ to ensure [read-write symmetry](#sec-read-write-symmetry). For PSA match types,
 a "don't care" match for a specific match key field is defined as follows:
 
 * For a `TERNARY` match, it is logically equivalent to a mask of zeros.
+
+* For a `OPTIONAL` match, it is logically equivalent to a wildcard match.
 
 * For an `LPM` match, it is logically equivalent to a prefix_len of zero.
 
@@ -2666,6 +2669,14 @@ assert(low <= high)
 assert(low != min_field_value || high != max_field_value)
 ~ End Pseudo
 
+* `OPTIONAL` match
+    * The binary string encoding of the value must conform to the
+      [Bytestrings](#sec-bytestrings) requirements.
+
+~ Begin Pseudo
+assert(BytestringValid(match.optional().value()))
+~ End Pseudo
+
 ### Action Specification
 
 The `TableEntry` `action` field must be set for every `INSERT` update but may be
@@ -2773,9 +2784,9 @@ static entries. If the table requires a priority value for entries, the server
 must populate the `priority` field appropriately, starting at 1 for the lowest
 priority entry and incrementing the value by 1 for each successive entry. Note
 that P4~16~ does not support assigning explicit priorities to static
-entries. When a priority value is required (&eg; for tables including `RANGE`
-and / or `TERNARY` matches in the case of PSA), it is inferred based on the
-order in which entries appear in the table declaration.
+entries. When a priority value is required (&eg; for tables including `RANGE`,
+`TERNARY` or `OPTIONAL` matches), it is inferred based on the order in which
+entries appear in the table declaration.
 
 ### Wildcard Reads { #sec-table-wildcard-reads}
 
@@ -2899,7 +2910,7 @@ entities {
 }
 ~ End Prototext
 
-This issue also exists for tables with `TERNARY` and / or `RANGE`
+This issue also exists for tables with `TERNARY`, `RANGE`, and `OPTIONAL`
 matches. However, in this case the priority is also taken into account for
 wildcard reads, and because a priority of 0 is not valid, in practice only the
 entries with the same priority as the "don't care" entry will be returned to the
@@ -3952,10 +3963,10 @@ can perform a `MODIFY` update with an empty `members` repeated field.
 
 To facilitate [read-write symmetry](#sec-read-write-symmetry), the server must
 return an `ALREADY_EXISTS` error in case of duplicate members. Unlike for match
-tables, a priority value is not required for ternary and range matches:
-overlapping entries do not need to be ordered and the parse state transition
-is determined by whether or not the packet matches at least one entry in the
-set.
+tables, a priority value is not required for ternary, range and optional
+matches: overlapping entries do not need to be ordered and the parse state
+transition is determined by whether or not the packet matches at least one entry
+in the set.
 
 See Appendix [#sec-value-set-example] for a more complex Value Set example.
 
@@ -5595,8 +5606,9 @@ An architecture may introduce new table match types [@P4MatchTypes]. P4Runtime
 accounts for this by providing the following hooks:
 
 * The `match` field in `p4.config.v1.MatchField` (p4info.proto) is a `oneof`
-  which can be either one of the default match types (`EXACT`, `LPM`, `TERNARY`
-  or `RANGE`) or an architecture-specific match type encoded as a string.
+  which can be either one of the default match types (`EXACT`, `LPM`, `TERNARY`,
+  `RANGE`, or `OPTIONAL`) or an architecture-specific match type encoded as a
+  string.
 
 * The `field_match_type` field in `p4.v1.FieldMatch` (p4runtime.proto) is a
   `oneof` which includes an `Any` Protobuf message [@ProtoAny] field
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1106,15 +1106,11 @@ table t {
 }
 ~ End P4Example
 
-The generated P4Info will contain the following. An empty structured annotation
-body generates an empty `expression_list` to remain consistent with the P4
-Language Spec.
+The generated P4Info will contain the following.
 
 ~ Begin Proto
 structured_annotations {
   name: "Empty"
-  expression_list {
-  }
 }
 ~ End Proto
 
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1059,7 +1059,6 @@ message Expression {
     string string_value = 1;
     int64 int64_value = 2;
     bool bool_value = 3;
-    ExpressionList expr_list_value = 4;
   }
 }
 
@@ -1079,15 +1078,13 @@ message StructuredAnnotation {
 The `StructuredAnnotation` message can represent either a `KeyValuePairList`
 or an `ExpressionList`.
 
-The type of an expression is intentionally limited to a practical subset:
-string literal, 64-bit signed integer, boolean, or a nested list of
-`Expression`s. This is more restrictive than the P4 language but provides
-reasonable types for known annotation use-cases. The `p4c` compiler frontend
-which generates `P4Info` will evaluate all expressions and simplify them to
-one of the valid types. Any expressions which don't match one of the valid
-types will generate an error. For integers exceeding 64 bits, besides issuing
-an error, the compiler *could* print a suggestion to use a string
-representation, and the `P4Info` consumer could perform any necessary
+The type of an expression is intentionally limited to one of the following
+base types: string literal, 64-bit signed integer, or boolean. The `p4c`
+compiler frontend which generates `P4Info` will evaluate all expressions and
+simplify them to one of the valid types. Any expressions which don't match one
+of the valid types will generate an error. For integers exceeding 64 bits,
+besides issuing an error, the compiler *could* print a suggestion to use a
+string representation, and the `P4Info` consumer could perform any necessary
 conversions.
 
 The following invariants hold:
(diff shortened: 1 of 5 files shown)


chrispsommers:
@stefanheule Thanks for the fast review, that's a nice suggestion. I'm open to anything reasonable. Let's get more feedback, I'm happy to refactor to the best consensus solution.

stefanheule:
Actually, thinking about it, it is not clear to me from the spec if constant list expressions are allowed or not. If they are not, then the protos can be even more simplified. If they can, the spec should probably be updated to say that. Right now it says:

> All `expression`s (and nested expressions) within a `stucturedAnnotationBody`
> must evaluate at compile-time to string literals, infinite-precision integers or
> booleans.

That, to me, says there are no lists? But I'm not sure what the "nested expressions" means.

chrispsommers:
@stefanheule "Nested expressions" was meant to include the case where an expression is an {expressionList}, to handle casaes like MyAnno[x=10,y={a,b,c}] - it might be very desirable to have the value of a kv-pair be a list of constant epressions.

stefanheule:
I think that probably needs to be made more precise then.  There are a number of nested expressions in P4, and most of them are not allowed (e.g. {kvlist}).

chrispsommers:
@stefanheule Good feedback, Could you be so kind as to make that comment in the P4-Spec PR then? I welcome improvements to it. Thanks!

stefanheule:
> @stefanheule Good feedback, Could you be so kind as to make that comment in the P4-Spec PR then? I welcome improvements to it. Thanks!

Done.

chrispsommers:
@stefanheule I tried out your proposed revised syntax against the examples I'd provided and indeed it is cleaner and more concise. Here is the a slightly-revised version  of it, followed by the original example P4info and the revised P4info using the new syntax. For just these 4 simple examples, the revised  P4info is 25 lines shorter (around 25% reduction). I'm in favor of changing to it,  thanks for the suggestion. I'll wait a bit for more feedback, then  proceed to update the PR accordingly.

Revised `p4types.proto`:
```
message KeyValuePair {
  string key = 1;
  Expression value = 2;
}

message KeyValuePairList {
  repeated KeyValuePair kv_pairs = 1;
}

message Expression {
  oneof value {
    string string_value = 1;
    int64 int64_value = 2;
    bool bool_value = 3;
    ExpressionList list_value = 4;
  }
}

message ExpressionList {
  repeated Expression expressions = 1;
}

message StructuredAnnotation {
  string name = 1;
  oneof body {
    ExpressionList expression_list = 2;
    KeyValuePairList kvpair_list = 3;
  }
}
```
Old P4Info for the examples:
```
=== Generate Empty structured anno ===
structured_annotations {
  name: "Empty"
}

=== Generate Structured expressionList ===
Generate expressionList ===
structured_annotations {
  name: "MixedExprList"
  elements {
    value {
      expression {
        int64_value: 1
      }
    }
  }
  elements {
    value {
      expression {
        string_value: "hello"
      }
    }
  }
  elements {
    value {
      expression {
        bool_value: true
      }
    }
  }
  elements {
    value {
      expression {
        bool_value: false
      }
    }
  }
  elements {
    value {
      expression {
        int64_value: 11
      }
    }
  }
}

=== Generate Structured kvList ===
structured_annotations {
  name: "Labels"
  elements {
    key: "short"
    value {
      expression {
        string_value: "Short Label"
      }
    }
  }
  elements {
    key: "hover"
    value {
      expression {
        string_value: "My Longer Table Label to appear in hover-help"
      }
    }
  }
}

=== Generate Complex kvList ===
structured_annotations {
  name: "MixedKV"
  elements {
    key: "label"
    value {
      expression {
        string_value: "text"
      }
    }
  }
  elements {
    key: "my_bool"
    value {
      expression {
        bool_value: true
      }
    }
  }
  elements {
    key: "list_key"
    value {
      expression_list {
        expressions {
          string_value: "a"
        }
        expressions {
          int64_value: 11
        }
        expressions {
          bool_value: true
        }
      }
    }
  }
  elements {
    key: "int_val"
    value {
      expression {
        int64_value: 6
      }
    }
  }
}
```
Revised P4Info:
```
=== Generate Empty structured anno ===
structured_annotations {
  name: "Empty"
}

=== Generate Mixed expressionList ===
structured_annotations {
  name: "MixedExprList"
  expression_list {
    expressions {
      int64_value: 1
    }
    expressions {
      string_value: "hello"
    }
    expressions {
      bool_value: true
    }
    expressions {
      bool_value: false
    }
    expressions {
      int64_value: 11
    }
  }
}

=== Generate kvList of  Strings===
structured_annotations {
  name: "Labels"
  kvpair_list {
    kv_pairs {
      key: "short"
      value {
        string_value: "Short Label"
      }
    }
    kv_pairs {
      key: "hover"
      value {
        string_value: "My Longer Table Label to appear in hover-help"
      }
    }
  }
}

=== Generate Complex kvList ===
structured_annotations {
  name: "MixedKV"
  kvpair_list {
    kv_pairs {
      key: "label"
      value {
        string_value: "text"
      }
    }
    kv_pairs {
      key: "my_bool"
      value {
        bool_value: true
      }
    }
    kv_pairs {
      key: "list_key"
      value {
        list_value {
          expressions {
            string_value: "a"
          }
          expressions {
            int64_value: 11
          }
          expressions {
            bool_value: true
          }
        }
      }
    }
    kv_pairs {
      key: "int_val"
      value {
        int64_value: 6
      }
    }
  }
}
```

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1020,6 +1021,257 @@ table my_ipv4_lkup {
...
+`MyAnno3[]` or `MyAnno4[kvList|expressionList]` have a more prescribed syntax,
+which allows declaring key-value lists or expression lists. Both unstructured
+and structured annotations may be used simultaneously on a P4 element and
+`P4Info` supports this.
+
+The annotations described up to this point, &eg; `@brief()`, have all been
+unstructured annotations, or simply annotations. These are represented in
+`P4Info` as `repeated string annotations` fields in the various `message`s.
+Similarly, structured annotations are represented in
+`repeated structured_annotations` fields which are siblings to the unstructured
```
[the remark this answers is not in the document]
chrispsommers: I made the suggested clarifications on `ints` and also added an example of nested expressions, which are indeed legal.

chrispsommers:
@stefanheule @antoninbas (and anyone else) please review simplified `Expression` in `StructuredAnnotation`, thanks. If this is satisfactory we will strive to get things approived in both the LDWG and the P4-API WG and start implementing. Thanks for all the engagement and feedback.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1020,6 +1021,173 @@ table my_ipv4_lkup {
...
+@Empty[]
+table t {
+    ...
+}
+~ End P4Example
+
+The generated p4Info will contain:
+
+~ Begin Proto
+structured_annotations {
```
stefanheule: This is missing the empty expression list.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1020,6 +1021,173 @@ table my_ipv4_lkup {
...
+  name: "MixedKV"
+  kvpair_list {
+    kv_pairs {
+      key: "label"
+      value {
+        string_value: "text"
```
stefanheule: These should be wrapped in a "value" message, I believe. One for kvPair and one for Expression.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1048,6 +1216,8 @@ message PkgInfo {
   string contact = 7;
   // url for more information, e.g. "http://support.p4.org/ref/p4/switch.p4_v1.0"
   string url = 8;
+  // Miscellaneous metadata, structured; a way to extend PkgInfo
```
stefanheule: I'm not sure I understand the "a way to extend PkgInfo" comment. Isn't PkgInfo a unstructured annotations?

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Good catch, thanks. Will fix.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: @stefanheule I don't understand this comment. I generated this output using Python generated _pb2.py libraries so I believe the structure is correct. Can you show ,e what you expected?

docs/v1/P4Runtime-Spec.mdk
chrispsommers: `PkgInfo` is a first-class annotation which has some predefined elements like `name` and `version`, but also allows you to add arbitrary content via annotations to `main()`. Previously only unstructured annotations existed. This PR  adds structured annotations as well.

docs/v1/P4Runtime-Spec.mdk
stefanheule: Sorry, my bad, it is correct.

docs/v1/P4Runtime-Spec.mdk
antoninbas: I personally see no reason to default to an `expression_list` here. Instead I "vote" for using an empty oneof if the structured annotation has an empty body:
```
structured_annotations {
  name: "Empty"
}
```

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1020,6 +1021,175 @@ table my_ipv4_lkup {
...
+
+The following invariants hold:
+
+1. For any P4 entity, there are no two `StructuredAnnotation`s that have the
+same name.
+
+2. Within a `KeyValuePairList`, there are no two `KeyValuePair`s that have the
+same `key.`
+
+### Structured Annotation Examples
```
```suggestion
#### Structured Annotation Examples
```

I think it makes more sense as a subsection?

docs/v1/references.bib
```diff
@@ -191,6 +191,11 @@ @ONLINE { ProtoOneOfBackwardsCompatibility
     url = "https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility-issues"
 }
 
+@ONLINE { P4Annotations,
+    title = "P4 Annotations",
+    url = "https://p4.org/p4-spec/docs/P4-16-v1.2.0.html#sec-annotations"
```
antoninbas: Feel free to change that link so that it refers to the "latest" (working draft) version of the P4 spec, since v1.2.0 does not include anything about structured annotations. This way when your p4-spec PR is merged, at least the link will point to some text that talks about structured annotations.

docs/v1/P4Runtime-Spec.mdk
stefanheule: I think at some point we said an empty list (since it's ambiguous) will default to an empty expression_list, but now I'm not sure why we thought that is necessary. Antonin is probably right and we should just not have the oneof.  I also don't see any text in the spec saying it defaults to an empty expression list, so that seems fine?

jafingerhut: @stefanheule I do not know if it is necessary, but the idea was to avoid ambiguity of whether an empty structured annotation was considered somehow both an empty expression list, and also an empty key/value pair list.  This spec pull request explicitly suggests making the empty structured annotation to be considered an empty expression list, at least partly to make the parsing grammar unambiguous: https://github.com/p4lang/p4-spec/pull/796/files

In particular, this proposed new line in that PR: "To avoid ambiguity, an empty `structuredAnnotationBody` is by definition considered to contain an empty `expressionList`."

jafingerhut: That said, the P4_16 language spec does not mandate how the P4Info data structure is defined.  If P4Runtime API folks would prefer a 3rd alternative in the oneof for the empty case, the P4_16 language spec writers need never be informed of that fact, and they will sleep well at night.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1020,6 +1021,175 @@ table my_ipv4_lkup {
...
+    expressions {
+      bool_value: false
+    }
+    expressions {
+      int64_value: 11
+    }
+  }
+}
+~ End Proto
+~ End Proto
```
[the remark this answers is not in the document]
chrispsommers: Thanks, fixed

docs/v1/references.bib
chrispsommers: @antoninbas Good suggestion. Would I link to the .mdk form of the spec in GitHub, e.g. https://github.com/p4lang/p4-spec/blob/master/p4-16/spec/P4-16-spec.mdk instead?

chrispsommers: @antoninbas Regarding the "empty expressionList" topic: I agree it makes more sense, and in my original example I did just that (empty oneof). However @stefanheule commented that it ought to be an empty ExpressionList to be consistent with the P4-16 Spec, namely this cluase:

stefanheule: That's fair, I take the blame for the change. But I do think we want the spec and this to be consistent, but if everyone feels like we want to not disambiguate the two, we should also not disambiguate in the spec.  Or we disambiguate the same way in both places.

docs/v1/P4Runtime-Spec.mdk
stefanheule: Oh, I looked in the changes here to see the text for the empty structured annotation, but couldn't find it. Duh, this is the p4rt repo, not the spec, so no wonder.

As I said elsewhere, I think both are reasonable, but I strongly think they should be consistent. We are proposing a design for more structured annotations and at the same time implementing it in P4RT.  Lets make sure we know what we want, and then make it consistent in both places. This is not a case of P4-spec already having something, and the P4RT community using that in some way, we are literally designing it at the same time in both places and it's the same set of people doing the design in both places. There is no good reason in my opinion to not make the same decision in both places.

chrispsommers: @stefanheule good points. Someone new to this would wonder why they differed. Given the formal language reasons for disambiguating, the "safest" course is to leave the P4Info as-is with the explicit empty `ExpressionList`. This way we can be criticized for the slightly wordier corner-case of a empty StructuredAnno (but I gotta wonder - why would someone use one in the first place?), but at least we won't be criticized  for  imprecision or inconsistency.

antoninbas: So in a comment in the p4c PR to support structured annotations, Mihai wrote:

> If both are empty it's the kv that is assumed to be used.

https://github.com/p4lang/p4c/pull/2227/files#r389943636

But it could be a typo, or it could be confusion on my side :)

I understand the need to disambiguate for the parser generator, I don't think it is very relevant for P4Info, but I'm fine with keeping it the way it is. I'd like us to add a sentence about this case in the P4Runtime spec though.

docs/v1/references.bib
antoninbas: @chrispsommers you can use this link: https://p4.org/p4-spec/docs/P4-16-working-spec.html#sec-annotations

let's use the other thread to discuss empty annotations

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1020,6 +1021,176 @@ table my_ipv4_lkup {
...
+
+The `StructuredAnnotation` message can represent either a `KeyValuePairList`
+or an `ExpressionList`.
+
+The type of an expression is intentionally limited to one of the following
+base types: string literal, 64-bit signed integer, or boolean. The `p4c`
+compiler frontend which generates `P4Info` will evaluate all expressions and
+simplify them to one of the valid types. Any expressions which don't match one
+of the valid types will generate an error. For integers exceeding 64 bits,
+besides issuing an error, the compiler *could* print a suggestion to use a
```
antoninbas: s/the compiler could/the compiler may

to follow RFC style

chrispsommers: Sure, I'll be happy to oblige both, thanks.

chrispsommers:
I believe all  issues discussed have been resolved in this PR and it is ready for merge.

antoninbas:
Alright, I think this PR has been open for long enough. Merging it now and we can make adjustments later if needed.

chrispsommers:
@antoninbas Thanks, I was about to throw it a birthday party! :)

chrispsommers:
BTW, the issues in p4-spec are settling down, I've clarified a number of things.  There will need to be some minor changes in handling duplicate annotations.

PR #266, opened 2020-02-11
Update P4Runtime version to 1.2.0-dev

antoninbas:
To indicate that this is an unreleased version of P4Runtime, and avoid
confusion.


antoninbas:
@chrispsommers No need to track every change (dev.1 -> dev.2 -> dev.3, ...) just like we are not going to create a Github release every time we make a change. The purpose of this change is just to avoid confusion between the latest release (currently 1.1.0) and the version under development (currently 1.2.0).

PR #271, opened 2020-03-12
Add metadata bytes field

### LINKED ISSUE #267: Support a string "cookie" in TableEntry
The current "controller_metadata" has type uint64: https://github.com/p4lang/p4runtime/blob/v1.1.0/proto/p4/v1/p4runtime.proto#L160. This is probably legacy from OpenFlow, where each flow can have a 64-bit integer cookie. We would like to be able to store an arbitrary string as metadata with each table entry.

stefanheule:
Fixes #267.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2482,6 +2482,12 @@ the `TableEntry` entity, which has the following fields:
 * `controller_metadata`, a 64-bit cookie value which is opaque to the
   target. There is no requirement of where this is stored, but it must be
   returned by the server along with the rest of the entry when the client
+  performs a read on the entry. This is deprecated in favor of the more flexible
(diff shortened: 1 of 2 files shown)


konne-google:
LGTM

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2480,6 +2480,12 @@ the `TableEntry` entity, which has the following fields:
   includes an optional, ternary or range match.
 
 * `controller_metadata`, a 64-bit cookie value which is opaque to the
+  target. There is no requirement of where this is stored, but it must be
+  returned by the server along with the rest of the entry when the client
+  performs a read on the entry. This is deprecated in favor of the more flexible
+  metadata field.
```
```suggestion
  `metadata` field.
```

stefanheule: I did both, thanks.

Issue #268, opened 2020-03-06
P4Info ID generation


stefanheule:
There appears to be confusion between the P4RT spec and the compiler (and me?) when it comes to how IDs are allocated. My understanding from reading the spec is that all IDs are 32 bits, the top most 8 bits are indicating the object type, and the rest is either generated by the compiler or set by using @id.  Specifically, the spec says "It is possible to statically set the least-significant 24 bits of the ID in the P4 program source by annotating the object with @id".

If I take this program and run it through p4c, I get the following P4Info.

```
action a(@id(100) bit<32> arg) {
}

control MyIngress(inout headers hdr, inout metadata meta, inout standard_metadata_t standard_metadata) {
  table t1 {
    key = {
      meta.f1 : exact;
      meta.f2 : exact @id(100); // this is ignored
    }
    actions = {
      a;
      NoAction();
    }
    default_action = NoAction();
  }
  @id(100) // ID does not have the right first 8 prefix bits
  table t2 {
    key = {
      meta.f1 : exact;
    }
    actions = {
      NoAction();
    }
    default_action = NoAction();
  }
  apply {
    t1.apply();
    t2.apply();
  }
}
```

```
pkg_info {
  arch: "v1model"
}
tables {
  preamble {
    id: 33584214
    name: "MyIngress.t1"
    alias: "t1"
  }
  match_fields {
    id: 1
    name: "meta.f1"
    bitwidth: 4
    match_type: EXACT
  }
  match_fields {
    id: 2
    name: "meta.f2"
    bitwidth: 4
    match_type: EXACT
  }
  action_refs {
    id: 16815170
  }
  action_refs {
    id: 16800567
  }
  size: 1024
}
tables {
  preamble {
    id: 100
    name: "MyIngress.t2"
    alias: "t2"
  }
  match_fields {
    id: 1
    name: "meta.f1"
    bitwidth: 4
    match_type: EXACT
  }
  action_refs {
    id: 16800567
  }
  size: 1024
}
actions {
  preamble {
    id: 16815170
    name: "a"
    alias: "a"
  }
  params {
    id: 1
    name: "arg"
    bitwidth: 32
  }
}
actions {
  preamble {
    id: 16800567
    name: "NoAction"
    alias: "NoAction"
  }
}
```

Some observations/questions:
[...]

PR #272, opened 2020-03-13
Add the ability to use @id for action params, table keys and packetio headers

stefanheule:
This is as proposed in #268, but also adding it for packetio headers, which seems natural.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1173,7 +1173,9 @@ statically-assigned ID suffixes lead to non-unique IDs (&ie; if the P4
 programmer tries to assign the same ID suffix to two different P4 objects of the
 same type by annotating them with the same `@id` value). Note that it is not
 possible for the P4 programmer to change the value of the 8-bit ID prefix, which
-encodes the object type.
+encodes the object type. The programmer is free to include or not include the
+8-bit prefix; the compiler will add it if not already present, and will check
+an existing prefix against the correct prefix and error if it is not correct.
 
 ~ TableFigure { #tab-exmpl-p4-obj-ids; \
     caption: "Example of statically-assigned P4Info object IDs"; \
@@ -1190,6 +1192,12 @@ encodes the object type.
 | `@id(0x12ab34) action act1...` | 0x0112ab34                                                 |
 +--------------------------------+------------------------------------------------------------+
 ~
+
+The `@id` annotation can also be used to choose the ID for match fields,
+action parameters, packet metadata, and . In this case, there is no 8-bit prefix
+and the programmer is free to choose any 32-bit number. The compiler must fail
+if the IDs chosen by the programmer are not unique (within a table and action,
+respectively).
 
 ## P4Info Objects
 
@@ -1208,7 +1216,9 @@ control plane. This message contains the following fields:
       this table. No rules are prescribed on the way `MatchField` IDs should be
       allocated, as long as two `MatchField` of the same table do not have the
       same ID. Nonetheless, we recommend that the IDs be assigned incrementally,
-      starting from 1, in the same order as in the P4 key declaration.
+      starting from 1, in the same order as in the P4 key declaration. The
+      programmer can chose the IDs using the `@id` annotation, or let the
+      compiler choose them.
 
     * `name`, the string representing the name of this `MatchField`.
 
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5680,6 +5693,8 @@ man-in-the-middle attacks between the server and client.
 * Add new `OPTIONAL` match kind. At the moment, `OPTIONAL` is only supported by
   the v1model architecture [@v1model], and not by PSA. We hope that it will be
   included in the core P4 language by the time P4Runtime v1.2.0 is released.
+* Added the ability to change the ID of table match fields, action parameters,
+  and metadata for packetio in P4Info by using the `@id` annotation.
```
```suggestion
  and Packet IO metadata fields in P4Info by using the `@id` annotation.
```
for consistency with the text in the section below

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1173,7 +1173,9 @@ statically-assigned ID suffixes lead to non-unique IDs (&ie; if the P4
 programmer tries to assign the same ID suffix to two different P4 objects of the
 same type by annotating them with the same `@id` value). Note that it is not
 possible for the P4 programmer to change the value of the 8-bit ID prefix, which
-encodes the object type.
+encodes the object type. The programmer is free to include or not include the
+8-bit prefix; the compiler will add it if not already present, and will check
```
jafingerhut: I know what you mean by "include or not include the 8-bit prefix", but perhaps nicer to be slightly more precise in a spec, e.g. "The programmer is free to leave the 8-bit prefix as 0, in which case the compiler will replace the 0 with the correct value for the kind of object the annotation is annotating.  The programmer may also fill in the 8-bit prefix with a non-0 value, in which case the compiler will give an error if the 8-bit prefix does not contain the correct value, or leave it as is if it is correct."

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1191,6 +1193,12 @@ encodes the object type.
 +--------------------------------+------------------------------------------------------------+
 ~
 
+The `@id` annotation can also be used to choose the ID for match fields,
+action parameters, and packet metadata. In this case, there is no 8-bit prefix
+and the programmer is free to choose any 32-bit number. The compiler must fail
+if the IDs chosen by the programmer are not unique (within a table, action, or
+header, respectively).
```
jafingerhut: I do not fully know the reasons for the 8-bit prefix identifying the type of object, but is allowing the most significant 8 bits of the id of a match field, action parameter, and/or packet metadata to be the same as that of one of the 'defined prefixes' in the spec actually what you want?  i.e. is anyone planning on taking an arbitrary id, selecting out its most significant 8 bits, and using that to determine what type of object it is?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1306,7 +1316,8 @@ The `Action` message defines the following fields:
       on the way `Param` IDs should be allocated, as long as two `Param` of the
       same action do not have the same ID. Nonetheless, we recommend that the
       IDs be assigned incrementally, starting from 1, in the same order as in
-      the P4 action declaration.
+      the P4 action declaration.  The programmer can chose the IDs using the
```
jafingerhut: minor typo: chose -> choose

jafingerhut: and similar in the next "hunk" of diff below, and perhaps others I have not read yet.

docs/v1/P4Runtime-Spec.mdk
stefanheule: Done, thanks for the great suggestion.

docs/v1/P4Runtime-Spec.mdk
stefanheule: Done

docs/v1/P4Runtime-Spec.mdk
stefanheule: I'm not sure what the reason is (I feel having that tag there in the first place seems unnecessary), so I don't know. I personally think it is fine as it is, but I'm open to hearing others give their opinion.

antoninbas: These ids are a bit specific in that their scope is limited to their parent P4 entity (e.g. table for match field parameters). That being said, it could make sense to restrict the prefix to be 0 to avoid confusion.

stefanheule: I don't feel strongly either way, would people like me to restrict these IDs to 24 bit?

stefanheule: In the WG meeting on 3/25 we decided not to restrict these IDs to keep the specification simple.

antoninbas:
@stefanheule please fix the conflict so we can merge

stefanheule:
> @stefanheule please fix the conflict so we can merge

Done.

antoninbas:
@stefanheule the diff shows that you have made the files you modified executable by changing their access permissions from 644 to 755. Could you revert that?

The diff also shows your earlier "metadata" changes for some reason, which we already have merged in master. This should not be a big deal when we merge, but you may want to `git fetch --all && git rebase origin/master` before pushing again.

stefanheule:
Fixed, I think. Sorry for the mishap.

PR #274, opened 2020-03-26
Support translation to strings

### LINKED ISSUE #270: Support strings in @p4runtime_translation
Referring to objects by a string name instead of an integer id makes P4RT flows a lot more readable, which for example greatly helps with debugging. We therefore propose to generalize the `@p4runtime_translation` annotation, to not just support fixed bit width integers, but also strings. 

So in addition to:
```
@p4runtime_translation("p4.org/psa/v1/PortId_t", 32)
```

We propose to also support:
```
@p4runtime_translation("p4.org/psa/v1/PortId_t", string)
```

Since values are already represented as `bytes` in P4RT messages, there is no need for any changes of those protos.

However, we do need to change the P4Info definition. Currently, `match_fields` and action `params` refer to translated values as follows:

```
  match_fields {
    id: 1
    name: "ingress_port"
    bitwidth: 32
    match_type: EXACT
    type_name {
      name: "PortId_t"
    }
  }

params {
    id: 1
    name: "ingress_port"
    bitwidth: 32
    type_name {
      name: "PortId_t"
    }
  }

new_types {
    key: "PortId_t"
    value {
      translated_type {
        uri: "p4.org/psa/v1/PortId_t"
        sdn_bitwidth: 32
      }
    }
  }
```

To support strings, we propose extending the `new_types` proto in https://github.com/p4lang/p4runtime/blob/master/proto/p4/config/v1/p4types.proto as follows:

```
message P4NewTypeTranslation {
  message SdnString {}

  // The URI uniquely identifies the translation in order to enable the
  // P4Runtime agent to perform value-mapping appropriately when required. It is
  // recommended that the URI includes at least the P4 architecture name and the
  // type name.
  string uri = 1;
  // The object is either represented as an unsigned integer with a bitwidth of `sdn_bitwidth`,
  // or as a string.
  int32 sdn_bitwidth = 2;
  SdnString sdn_string = 3;
}
```

So in the above example, the P4Info would be:

```
new_types {
    key: "PortId_t"
    value {
      translated_type {
        uri: "p4.org/psa/v1/PortId_t"
        sdn_string: {}
      }
    }
  }
```

When translating from a string, the `bitwidth` field in `match_fields` and action `params` is ignored.

konne88:
Fixes #270

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2369,14 +2369,15 @@ whose underlying data plane representation may vary on different targets, but
 for which it may be convenient to present a unified representation and numbering
 scheme to the control plane. **The `@p4runtime_translation` annotation can only
 be used if the underlying P4 built-in type is a fixed-width unsigned bitstring
-type (`bit<W>`)** . The type exposed to the control plane can be either a
-fixed-width unsigned bitstring, with a potentially different bitwidth, or a
-string. The annotation takes two parameters: a *URI* (Uniform Resource
-Identifier) which uniquely identifies the translation being performed on
-entities of the new type to the P4Runtime server and the *controller_type* of
-the values exposed the control plane. The *controller_type* can be either
-`bit<W>` where `W` is any positive integer, or `string`,
-or a positive integer `W` which has the same meaning as `bit<W>`.
+type (`bit<W>`)**. The type exposed to the control plane (referred to as the
+*controller_type*) can be either a fixed-width unsigned bitstring, with a
+potentially different bitwidth, or a string. The annotation takes two
+parameters: a *URI* (Uniform Resource Identifier) which uniquely identifies the
+translation being performed on entities of the new type to the P4Runtime server
+and the *controller_type* of the values exposed to the control plane. The
+*controller_type* can be either `bit<W>` where `W` is any positive integer, or
+`string`, or a positive integer `W` which has the same meaning as `bit<W>`.
+Specifying just an integer is deprecated.
 
 It is recommended that the URI includes at least the P4 architecture name and
 the type name.
@@ -2412,15 +2413,15 @@ For example, an architecture --- in this case PSA --- may introduce a new type
 for port numbers:
 ~ Begin P4Example
 // Controller refers to ports as a string, e.g. "eth0".
-@p4runtime_translation("p4.org/psa/v1/PortId_t", string)
+@p4runtime_translation("p4.org/psa/v1/PortId_String_t", string)
 type bit<9> PortId_String_t;
 
-// Controller refers to ports as a 32bit integer, e.g. 0xffffffff.
-@p4runtime_translation("p4.org/psa/v1/PortId_t", bit<32>)
+// Controller refers to ports as a 32-bit integer, e.g. 0xffffffff.
+@p4runtime_translation("p4.org/psa/v1/PortId_Bit32_t", bit<32>)
 type bit<9> PortId_Bit32_t;
 
 // Same as above.
-@p4runtime_translation("p4.org/psa/v1/PortId_t", 32)
+@p4runtime_translation("p4.org/psa/v1/PortId_32_t", 32)
 type bit<9> PortId_32_t;
 ~ End P4Example
 
@@ -2433,7 +2434,7 @@ type_info {
     key: "PortId_String_t"
     value {  # P4NewTypeSpec
       translated_type {  # P4NewTypeTranslation
-        uri: "p4.org/psa/v1/PortId_t"
+        uri: "p4.org/psa/v1/PortId_String_t"
         sdn_string: {}
       }
     }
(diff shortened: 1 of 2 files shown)


konne88:
@stefanheule @jafingerhut @antoninbas

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2369,13 +2369,17 @@ whose underlying data plane representation may vary on different targets, but
...
-is recommended that the URI includes at least the P4 architecture name and the
-type name.
+type (`bit<W>`)** . The type exposed to the control plane can be either a
+fixed-width unsigned bitstring, with a potentially different bitwidth, or a
+string. The annotation takes two parameters: a *URI* (Uniform Resource
+Identifier) which uniquely identifies the translation being performed on
+entities of the new type to the P4Runtime server and the *controller_type* of
+the values exposed the control plane. The *controller_type* can be either
+`bit<W>` where `W` is any positive integer, or `string`,
+or a positive integer `W` which has the same meaning as `bit<W>`.
```
stefanheule: Should we explicitly make this deprecated?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2369,13 +2369,17 @@ whose underlying data plane representation may vary on different targets, but
...
-the translation being performed on entities of the new type to the P4Runtime
-server and the *bitwidth* of the bitstring type exposed to the control plane. It
-is recommended that the URI includes at least the P4 architecture name and the
-type name.
+type (`bit<W>`)** . The type exposed to the control plane can be either a
+fixed-width unsigned bitstring, with a potentially different bitwidth, or a
+string. The annotation takes two parameters: a *URI* (Uniform Resource
+Identifier) which uniquely identifies the translation being performed on
+entities of the new type to the P4Runtime server and the *controller_type* of
+the values exposed the control plane. The *controller_type* can be either
```
jafingerhut: maybe insert "to", e.g. "values exposed to the control plane"?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2396,27 +2400,61 @@ the following fields:
...
 
 * `annotations`, a repeated field of strings, each one representing a P4
   annotation associated to the type declaration.
 
 For example, an architecture --- in this case PSA --- may introduce a new type
 for port numbers:
 ~ Begin P4Example
+// Controller refers to ports as a string, e.g. "eth0".
+@p4runtime_translation("p4.org/psa/v1/PortId_t", string)
+type bit<9> PortId_String_t;
```
jafingerhut: This question need not cause this example to be changed, but to satisfy my own curiosity, if the control plane type is different, wouldn't you perhaps expect the URI to be different, too?  Having all of the 3 URIs in this example be identical to each other seems like something one would never want to do.  e.g. wouldn't you expect some code that performs the translation to use the URI to decide how to do it, and perhaps only the URI?

docs/v1/P4Runtime-Spec.mdk
konne88: done

docs/v1/P4Runtime-Spec.mdk
konne88: done

docs/v1/P4Runtime-Spec.mdk
konne88: done

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2369,13 +2369,17 @@ whose underlying data plane representation may vary on different targets, but
...
 scheme to the control plane. **The `@p4runtime_translation` annotation can only
 be used if the underlying P4 built-in type is a fixed-width unsigned bitstring
-type (`bit<W>`)** and the type exposed to the control plane will also be a
-fixed-width unsigned bitstring, with a potentially different bitwidth. It takes
-two parameters: a *URI* (Uniform Resource Identifier) which uniquely identifies
-the translation being performed on entities of the new type to the P4Runtime
-server and the *bitwidth* of the bitstring type exposed to the control plane. It
-is recommended that the URI includes at least the P4 architecture name and the
-type name.
+type (`bit<W>`)** . The type exposed to the control plane can be either a
```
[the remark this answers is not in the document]
konne88: done

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2396,27 +2400,61 @@ the following fields:
...
   annotation associated to the type declaration.
 
 For example, an architecture --- in this case PSA --- may introduce a new type
 for port numbers:
 ~ Begin P4Example
+// Controller refers to ports as a string, e.g. "eth0".
+@p4runtime_translation("p4.org/psa/v1/PortId_t", string)
+type bit<9> PortId_String_t;
+
+// Controller refers to ports as a 32bit integer, e.g. 0xffffffff.
```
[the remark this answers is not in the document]
konne88: done

konne88:
Just opened https://github.com/p4lang/p4c/issues/2283

PR #276, opened 2020-03-28
Support @id annotations for value_set match fields

antoninbas:
https://github.com/p4lang/p4runtime/pull/272 added support for table
match fields, action parameters, and controller header metadata, but we
forgot value_set match fields, as pointed out by @jafingerhut.


antoninbas:
I also made the language in all 4 sections more uniform, and fixed a random typo.

PR #279, opened 2020-03-30
Clarifications on action profiles and selectors

jafingerhut:
I believe that these are clarifications to the existing P4Runtime API
specification, not changes to its behavior.

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -3292,8 +3292,9 @@ table t {
 
 When programming table `t` in the example above, a P4Runtime client should
 specify the `TableAction` in the `TableEntry` to be a reference to either an
-action profile member or group. The reference is a `uint32` identifier that
-uniquely identifies a member or group programmed in the action selector `as`.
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3292,8 +3292,9 @@ table t {
 
 When programming table `t` in the example above, a P4Runtime client should
 specify the `TableAction` in the `TableEntry` to be a reference to either an
-action profile member or group. The reference is a `uint32` identifier that
-uniquely identifies a member or group programmed in the action selector `as`.
+action profile member or group. The reference is a non-0 `uint32` identifier
```
stefanheule: I think this is more commonly spelled non-zero. I don't feel super strongly though.

antoninbas:
@jafingerhut did you want to address Stefan's suggestion?

jafingerhut:
@antoninbas Done with second commit a minute ago, including replacing an occurrence of "non-0" that I did not add with this PR, already in the document, because my text editor found it.

PR #281, opened 2020-04-02
Add document with P4Runtime id notes

jafingerhut:


jafingerhut:
If you would like to read it with all of the easier-to-read table formatting, you can look at it here: https://github.com/jafingerhut/p4runtime/blob/doc-add-p4runtime-id-notes/docs/v1/p4runtime-id-notes.md

jafingerhut:
@chrispsommers There is absolutely no need for you to read this or provide review comments, but since you gave a thumbs-up on the issue this is intended to address, I thought I'd ping you in case you were interested.

docs/v1/p4runtime-id-notes.md
```diff
@@ -0,0 +1,483 @@
...
+
+Section 6.3 "ID Allocation for P4Info Objects" of the P4Runtime API
+specification says:
+
+    The most significant 8 bits of the ID encodes the object type (as
+    per Table 1).  [ ... ]  The remaining 24 bits must be generated in
+    such a way that the resulting IDs must be globally unique in the
+    scope of the P4Info message.
+
+So all of these id values must be globally unique for a single P4Info
```
stefanheule, line 75: It might be worth saying that the yare unique w.r.t to all IDs of this type, not really globally unique. You can have a table ID that is the same as a field match ID, or role ID, etc.

jafingerhut: Makes sense.  Maybe introduce a name to describe this "scope", e.g. "top level P4Info object scope"?  There is probably a better name.

jafingerhut: I have added another commit attempting to address this comment.  I called the new scope the "P4Info top level scope".

docs/v1/p4runtime-id-notes.md
```diff
@@ -0,0 +1,487 @@
...
+
+```
+commit c9cd4af17c2d562d6e4ec760237f16fea9558fcb
+Author: Chris Sommers <31145757+chrispsommers@users.noreply.github.com>
+Date:   Fri Mar 27 13:43:31 2020 -0700
+```
+
+This is after version 1.1.0 of the specification was released, and it
+is expected that version 1.2.0 will be released soon after this time,
+with very few (if any) changes from the version above.
```
antoninbas, line 24: @jafingerhut please remind me to update this text when I do the v1.2.0 release

docs/v1/p4runtime-id-notes.md
```diff
@@ -0,0 +1,487 @@
...
+
+## Numeric ids that are unique within the P4Info top level scope
+
+The messages listed below are all of those that are sub-messages of
+the `P4Info` message, and they contain a `Preamble` sub-message.  Each
+`Preamble` message contains an `id` field, as well as several others
+like `name`, `alias`, and `annotations`.  See the `Preamble` message
+definition for the full list.
+
+Section 6.3 "ID Allocation for P4Info Objects" of the P4Runtime API
```
antoninbas, line 67: suggestion: you can put a link to the section here, this way you ensure you won't need to update the section number in the future.

docs/v1/p4runtime-id-notes.md
```diff
@@ -0,0 +1,487 @@
...
+program and the controller software.)
+
+
+### Clone session ids
+
+Scope: A single P4Runtime device.
+
+`session_id` field values appear in the message types listed below.
+The allowed set of values is determined by the P4Runtime server,
+typically in a contiguous range `[0, C-1]` for some maximum number of
```
antoninbas, line 287: 0 is no longer valid since P4Runtime v1.1.0

docs/v1/p4runtime-id-notes.md
jafingerhut: Do you mean a link to the section heading of the HTML version of the latest auto-generated version, e.g. this?  https://p4.org/p4runtime/spec/master/P4Runtime-Spec.html#sec-id-allocation

docs/v1/p4runtime-id-notes.md
jafingerhut: Thanks for catching that.  Changed to `[1, C]`

docs/v1/p4runtime-id-notes.md
jafingerhut: Sure, I can do that, and/or create a PR to update it myself.

jafingerhut:
> Did you want to include a section at the top with the special meaning of 0 for all ids:
> 
> * invalid for P4Info ids - used for wildcard reads when 0 is used in P4Runtime messages in lieu of a P4Info id reference
> * election_id / role_id: special meaning
> * ...
> 
> Or do you think you already cover this enough in the different sections of your doc?

I could add something for this, but does the spec explicitly say already what 0 means for the following kinds of ids?  I could not find anything mentioned about that.

+ election id
+ device id
+ digest list id
+ member_id and group_id for action profiles and action selectors.  I am guessing the intent is that 0 is invalid for these?

+ role_id of 0 is explicitly specified
+ 0 is explicitly reserved/invalid for P4 object id values, including their wildcard behavior in ReadRequest messages (at least in most cases -- I did not check for all of them exhaustively)
+ const_default_action_id=0 and implementation_id=0 are defined
+ SDN port number 0 invalid is specified explicitly

jafingerhut:
Minor update to my previous comment: I myself added some changes making it more explicit that member_id and group_id must be non-0, but can be 0 in ReadRequest as a wildcard, just as is done for many other things.

I still do not know if the current P4Runtime spec has anything to say on any specialness of 0 for device id, election id, and/or digest list id values.

antoninbas:
@jafingerhut 

**election_id**
a bit hard to find maybe but 0 has a special meaning, which is different based on whether the message is sent by the server or by a client:
*by a client:*
> The sender need not specify an election_id. If the election_id is not specified, the sender's election_id is considered lower than any election_id, and the sender will thus never become master. This way, a controller can choose to be a standby controller, in order to avoid master “flapping” (if a standby controller connects to the switch shortly before the actual master controller, therefore becoming master temporarily).
*by the server:*
See section 5.4. I feel like the spec may not be super clear, but my understanding is that if there is no master yet, an election_id of 0 is used in the `MasterArbitrationUpdate` sent by the controller whenever a controller with no election id (case above) connects.

**device_id**
I thought that 0 was invalid but I could not find anything in the spec to support that. Maybe it is invalid in Stratum specifically and I got confused? Potentially something to discuss at the next meeting. I don't see any obvious reason why 0 should not be a valid device id. @konne88 @stefanheule 

**digest list_id**
You're right. I don't see anything preventing the use of 0 by the server, and I don't see any reason to prevent it explicitly - except perhaps that it may be the last id on all P4Info / P4Runtime for which 0 is a valid value, at least if we declare 0 to be invalid for device_id :)

docs/v1/p4runtime-id-notes.md
antoninbas: Yes, that's what I meant

antoninbas:
actually for digest `list_id` it may be a good idea to reserve 0. It could be used by a controller to trigger a deletion of all cache entries on the server side (in a `DigestLickAck` message), which may or may not be useful :) However, that's not something that's covered by the spec today, so digest `list_id` 0 currently does not have any special meaning, and no change is required in your document for that one.

jafingerhut:
I have added another commit (the 19th one in this PR) that gives notes on whether 0 is special for several kinds of id values.  I mentioned that this is still TBD for device id and digest list id values.

PR #284, opened 2020-04-15
Add generated Go code for P4Runtime

### LINKED ISSUE #260: Host Go generated code for P4Runtime in this repo
To enable users to quickly import and use P4Runtime in their Go projects using an import like this one: `import p4v1 "github.com/p4lang/p4runtime/p4/v1"`.

antoninbas:
The generated files are placed under the go/ directory, and the 2 Go
packages which can be imported by other projects are
"github.com/p4lang/p4runtime/go/p4/v1" and
"github.com/p4lang/p4runtime/go/p4/config/v1".

As part of CI, we now check that the generated Go files are up-to-date
with respect to the checked-in Protobuf files. We also check that the
go.mod file is "tidy".

When updating the Protobuf files, a contributor can re-generated the Go
files with `./build/update_go.sh`.

Fixes #260


CI/check_go.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -eo pipefail
+
+THIS_DIR=$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)
```
smolkaj, line 5: You may want to add quote to be safe: `THIS_DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"`

CI/check_go.sh
```diff
@@ -0,0 +1,21 @@
...
+
+pushd $THIS_DIR/.. >/dev/null
```
smolkaj, line 7: Add quotes: `pushd "$THIS_DIR/.." >/dev/null`

build/update_go.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -e
+
+THIS_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
```
smolkaj, line 5: See comments above about quotes, here and in all bash scripts.

Issue #285, opened 2020-04-22
Add source locations to annotations


smolkaj:
## Motivation
The type of (unstructured) annotations in `p4info.proto` is currently just `string`, e.g.:
```
message Preamble {
  uint32 id = 1;
  string name = 2;
  string alias = 3;
  repeated string annotations = 4;
  Documentation doc = 5;
}
```
Since annotations can be arbitrary strings, consuming them will typically involve parsing them first. Thus, an annotation consumer may want to point out syntax/type errors when parsing fails. For example, our [p4-constraints prototype](https://github.com/p4lang/p4-constraints) produces messages such as the following:
```
path/to/example.p4:60:5-15:
   |   @entry_restriction("
60 |     ternary_key::prefix_length == 32
   |     ^^^^^^^^^^^
Type error: expression of type ternary<32> has no field 'prefix_length'
```
This requires knowledge of the location of the annotation in the source file.

## Proposal
I propose adding a more informative `Annotation` message as follows:
```
message Annotation {
  string tag = 1;                        // The part between '@' and '('.
  string payload = 2;                    // The part between '(' and ')'.
  SourceLocation source_location = 3;
}
message SourceLocation {
  string file = 1;  // Path to source file (total or relative to working directory).
  int line = 2;     // Offset of payload in source file.
  int column = 3;   // Offset of payload in source file.
}
```
`Preamble` (and other messages carrying annotations) would then be updated as follows:
```
message Preamble {
  ...
  repeated Annotation annotations = 4;
  ...
}
```

## Workaround
Our prototype currently obtains source locations using the builtin preporcessor macros `__FILE__` and `__LINE__` as follows:
```
@file(__FILE__)
@line(__LINE__)
@my_annotation("
  foo bar baz
")
```
This pollutes the .p4 source file with clutter, is a burden to the user, and is brittle.

stefanheule:
Shouldn't we add this to both structured and unstructured annotations?

smolkaj:
We discussed this proposal at the P4 API working group meeting. We agreed on the following points:
* Source locations should be added to both structured and unstructured annotations.
  * While syntax errors seem unlikely in structured annotations, there will likely still exists semantically invalid inputs; being able to point out the source location of such invalid inputs seems useful.
* The source location should refer to the location of the '@' symbol in all cases.
* To make source locations more useful, [p4c needs to preserve whitespace](https://github.com/p4lang/p4c/issues/2333).
[...]

PR #287, opened 2020-04-24
[breaking change] use common data structure for all annotations

smolkaj:
This is non-backward-compatible solution to issue #285. To be included in 2.0.

### DIFF
--- a/proto/p4/config/v1/p4types.proto
+++ b/proto/p4/config/v1/p4types.proto
@@ -181,6 +181,17 @@ message StructuredAnnotation {
     ExpressionList expression_list = 2;
     KeyValuePairList kv_pair_list = 3;
   }
+  // Location of the '@' symbol of this annotation in the source code.
+  SourceLocation source_location = 4;
+}
+
+// Location of code relative to a given source file.
+message SourceLocation {
+  // Path to the source file (total or relative to the working directory).
+  string file = 1;
+  // Line and column numbers within the source file, 1-based.
+  int32 line = 2;
+  int32 column = 3;
 }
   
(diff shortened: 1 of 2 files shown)


antoninbas:
@smolkaj this should probably be an issue at this stage and not a PR against master, which we cannot merge for a while.

It is possible that time has come for us to create a development branch for P4Runtime v2.0, but even then it makes sense to have a corresponding issue that we can use for tracking.

jafingerhut:
It is entirely up to the working group how they want to handle this in the Github repo files, but an alternative to creating a branch is to have separate directories for v1 and v2 of the P4Runtime API spec, and initially populate v2 files with copies of the latest v1 files.

There are disadvantages to such an approach, but a potential advantage is to make it easier to continue to make small corrections/clarifications to the v1 files as time goes on, and easily keep them "in the file system" for operations like auto-generating HTML/PDF from both v1 and v2 without doing git checkout operations.

smolkaj:
I created #288 to track this.

smolkaj:
I will close this PR until it becomes relevant again.

proto/p4/config/v1/p4types.proto
```diff
@@ -189,11 +201,11 @@ message P4EnumTypeSpec {
   message Member {
     string name = 1;
     repeated string annotations = 2;
```
chrispsommers, line 203: Why isn't this deleted? (same comment for other occurrences of `string annotations`

smolkaj: Good catch, fixed :)

smolkaj:
Just opening this temporarily to incorporate updates pointed out by @chrispsommers .

PR #292, opened 2020-05-07
Mention that device_id 0 is a normal device_id in P4Runtime v1.x

jafingerhut:


jafingerhut:
This PR does not change the spec, only the side document with notes on P4Runtime API id values.

PR #302, opened 2020-06-23
fix badge in README

smolkaj:
There was a typo in the badge link, causing the bade not to display correctly.


smolkaj:
We actually also need to address the TODOs in p4runtime/bazel/example/WORKSPACE.bazel. Will do so tomorrow.

PR #297, opened 2020-06-09
Fix Bazel build rules, addressing #250

smolkaj:
Also add CI to prevent future breakage and update the BAZEL_NOTES.
Add '.bazel' suffixes to Bazel files to be respectful to people unfamiliar with
Bazel.

PR #303, opened 2020-06-23
Fix Bazel example

antoninbas:
The "bazel" branch no longer exists after merging #297. This is
temporary until tag v1.2.0 is available.


antoninbas:
Merging this since the CI build is broken because of a branch which no longer exists.

smolkaj:
Thanks for taking care of this!

PR #304, opened 2020-06-24
CI should build latest commit, not latest release.

smolkaj:
We want `bazel/example` to accomplish two things:
* It should serve as a template for people who want to use p4runtime in their own Bazel project.
* It should ensure that the current master commit of p4runtime builds when included in third-party Bazel projects.

This PR fixes the second point. Before this PR, we would build a fixed commit rather than the latest commit.


smolkaj:
Sounds good!

PR #315, opened 2020-09-25
Updating spec to include IANA-assigned port: 9559

### LINKED ISSUE #306: What is the default / recommended TCP port to run a P4Runtime server on?
I could not find a reference to that in the spec.
Is it even desirable to define a standard port?

For reference: [IANA](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=9339) lists TCP port 9339 for gNMI.

bocon13:
closes #306

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -6079,15 +6079,6 @@ properties, but we may include on in future versions of the API.
   in particular, there is no way for a client to query the supported minor +
   patch version numbers.
 
-# Security concerns for P4Runtime
-
-Appropriate measures and security best practices need to be in place to protect
-the P4Runtime server and client, and the communication channel between the two.
-For example, firewalling and authenticating the incoming connections to the
-P4Runtime server can prevent a malicious actor from taking over the switch.
-Similarly, using TLS to authenticate and encrypt the gRPC channel can prevent
-man-in-the-middle attacks between the server and client.
-
 # Appendix { @h1:"A"}
 
 ## Revision History
@@ -6309,22 +6300,5 @@ the server did not deem necessary to increase the default value ---, unless the
 clients are aware that the server is using a larger maximum receive message
 size. The gRPC server running the P4Runtime service must not set the maximum
 receive message size to a value smaller than the default (4MB).
-
-### gRPC Default Port
-
-IANA has allocated TCP port number 9559 for the P4Runtime service, and
-P4Runtime server implementations must adopt port 9559 as the default. Users
-should be allowed to override the default using a configuration file or flag
-when starting the server. Uses of other port numbers as the default should be
-discontinued.
-
-For example, in C++, one can set the listening port as follows:
-~ Begin CPP
-::grpc::ServerCredentials server_credentials = /*...*/;
-::grpc::ServerBuilder server_builder;
-builder.AddListeningPort("0.0.0.0:9559", server_credentials);
-builder.RegisterService(/*...*/);  // register P4Runtime service
-builder.BuildAndStart();
-~ End CPP
 
 [BIB]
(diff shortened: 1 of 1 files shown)


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6284,5 +6284,21 @@ clients are aware that the server is using a larger maximum receive message
 size. The gRPC server running the P4Runtime service must not set the maximum
 receive message size to a value smaller than the default (4MB).
 
+### gRPC Default Port
+
+IANA has allocated TCP port number 9559 for the P4Runtime service, and
+P4Runtime server implementations should adopt port 9559 as the default. Users
```
antoninbas, line 6290: maybe we could use a stronger verb here, s/should/must/?

antoninbas (commented):
Thanks for taking care of this @bocon13. I think we should consider moving this out of the appendix section and to a more visible place of the spec. As a suggestion, maybe this could go to section 3 (Reference architecture) - I have been considering moving the section on security concerns to a subsection there as well, since we have a lot of top-level sections right now. But I am open to other solutions.

antoninbas:
Also, please update the "Revision History" as part of the PR - see https://github.com/p4lang/p4runtime/pull/317/files for an example

docs/v1/P4Runtime-Spec.mdk
bocon13: updated to "must"

bocon13:
@antoninbas I updated the Revision History and added as new subsection under Section 3 to cover the definition of the P4Runtime server. PTAL

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -414,6 +414,27 @@ caption: "P4Runtime Reference Architecture." }
...
+### Security concerns
+
+Appropriate measures and security best practices must be in place to protect
+the P4Runtime server and client, and the communication channel between the two.
+For example, firewalling and authenticating the incoming connections to the
+P4Runtime server can prevent a malicious actor from taking over the switch.
+Similarly, using TLS to authenticate and encrypt the gRPC channel can prevent
+man-in-the-middle attacks between the server and client. Mutual TLS (mTLS) may
+be used to facilitate the authentication of the client by the server and
+vice-versa.
```
antoninbas, line 436: Thanks for moving this, but you forgot to remove the original "Security concerns for P4Runtime" section?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6284,5 +6310,21 @@ clients are aware that the server is using a larger maximum receive message
 size. The gRPC server running the P4Runtime service must not set the maximum
 receive message size to a value smaller than the default (4MB).
 
+### gRPC Default Port
+
+IANA has allocated TCP port number 9559 for the P4Runtime service, and
+P4Runtime server implementations must adopt port 9559 as the default. Users
+should be allowed to override the default using a configuration file or flag
+when starting the server. Uses of other port numbers as the default should be
+discontinued.
```
antoninbas, line 6319: This information is now redundant - I'm personally fine with either removing this altogether or moving the C++ sample code to "P4Runtime Service Implementation" in Section 3.

bocon13: This info is pretty obvious I think; I just added it because of the precedent in the previous sections. I will remove the whole thing.

docs/v1/P4Runtime-Spec.mdk
bocon13: oops good catch.

antoninbas (approved):
LGTM

antoninbas:
Will merge this tomorrow if there are no other comments

PR #320, opened 2020-09-25
Replace references to 'master/slave' in the spec doc

bocon13:
Using the following mapping:
    - master -> primary
    - slave -> backup
    - mastership arbitration -> client arbitration
    - mastership -> primary client

**TODO**
- [x] clean up line lengths
- [x] update revision history


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -375,11 +375,11 @@ The following are not in scope of this specification document:
 
 Figure [#fig-reference-architecture] represents the P4Runtime Reference
 Architecture. The device or target to be controlled is at the bottom, and one or
-more controllers is shown at the top. A multi-master protocol allows more than
-one controller to participate, and a role-based arbitration scheme ensures only
-one controller has write access to each read/write entity, or the pipeline
-config itself. Any controller may perform read access to any entity or the
-pipeline config. Later sections describe this in detail. For the sake of
+more controllers is shown at the top. An active-active primary protocol allows
```
bocon13, line 378: This might not quite be right.

antoninbas:
As an additional TODO, some figures in the spec need to be updated

docs/v1/P4Runtime-Spec.mdk
antoninbas: I feel like "multi-master" protocol was not right to begin with. I am not sure what it's supposed to mean here. The figure shows a single primary controller. Maybe we should just replace "A multi-master protocol allows more than one controller to participate" with "More than one controller can participate".

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -603,11 +603,11 @@ build/embedded-plus-two-remote-controllers.[svg,png] \
 ## Embedded Controller + Two High-Availability Remote Controllers
 
 Figure [#fig-embedded-plus-two-remote-ha-controllers] illustrates a single
-embedded controller plus two remote controllers in an active-standby HA
+embedded controller plus two remote controllers in an active-standby (&ie; primary-backup) HA
 (High-Availability) configuration. Controller #1 is the active controller and is
 in charge of some entities. If it fails, Controller #2 takes over and manages
 the tables formerly owned by Controller #1. The mechanics of HA architectures
-are beyond the scope of this document, but the P4Runtime multi-master
+are beyond the scope of this document, but the P4Runtime active-active primary
```
antoninbas, line 610: Same as above. I don't think "multi-master" was accurate in the first place, and I am not sure "active-active" applies here since the backup controllers don't do anything until the primary fails. So I would suggest simply "the P4Runtime client arbitration scheme"

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5329,30 +5329,30 @@ message PacketMetadata {
...
-## Master Arbitration Update
+## Client Arbitration Update
 
-P4Runtime's master arbitration mechanism ensures that only the current master
+P4Runtime's client arbitration mechanism ensures that only the current primary
 can modify state on the switch, and that the `election_id` is monotonically
 increasing. For example, the switch must finish all previous write operations
-before changing masters, and must only accept write requests from the current
-master.
+before changing primarys, and must only accept write requests from the current
```
antoninbas, line 5337: s/primarys/primaries

or "before selecting a different primary"

antoninbas (commented):
I think the "active-active" does not really apply, otherwise changes LGTM. Some figures also need to be updated.

docs/v1/P4Runtime-Spec.mdk
bocon13: How about this?

"P4Runtime only allows for a single primary controller to actively write to each read/write entity. P4 entities can be divided into different groups, called roles. P4Runtime allows for a primary controller for each role, and a role-based arbitration scheme..."

docs/v1/P4Runtime-Spec.mdk
bocon13: That make sense, we can also do: "but the P4Runtime role-based client arbitration scheme"

docs/v1/P4Runtime-Spec.mdk
bocon13: I like the latter better.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -357,7 +357,7 @@ The following are not in scope of this specification document:
   : Refers to a gRPC Stream, which is a RPC on which several messages can be
     sent and received. P4Runtime defines one Stream RPC (`StreamChannel`), which
     is a bidirectional stream (both the client and the server can send messages)
-    which is used for packet I/O and master-slave arbitration, among other
+    which is used for packet I/O and primary-backup arbitration, among other
```
bocon13, line 360:
```suggestion
    which is used for packet I/O and client arbitration, among other
```

antoninbas (commented):
Apart from one typo, this looks good to me. In the future, please consider creating a branch directly in this repo since you have write access (it lets us look at the rendered spec HTML at `https://s3-us-west-2.amazonaws.com/p4runtime/travis/<your_branch_name>/P4Runtime-Spec.html` before merging the PR, in order to catch formatting issues. I generated the docs manually on my laptop and everything looks fine, including the 2 modified figures.

bocon13:
@antoninbas I made the one fix for the typo that you caught and rebased against latest master. Unfortunately, I don't think that I have write access to the repo which is why I was using my personal branch.

```ERROR: Permission to p4lang/p4runtime.git denied to bocon13.```

antoninbas (approved):
LGTM

antoninbas:
@bocon13 My bad I thought you already had write access, I sent you an invite

antoninbas:
Merging this PR as it has been open for a while, and there was no objection at the 10/23/2020 WG meeting.

bocon13:
Great, thanks @antoninbas! Looks like write access is working, too.

PR #322, opened 2020-10-09
Fix codegen/update_go.sh for Mac

bocon13:
-p isn't supported on Mac, but a template can be provided be provided on Linux and Mac


antoninbas (approved):
LGTM

PR #323, opened 2020-10-21
Add optional Scope message

### LINKED ISSUE #316: Add a scope argument to Entity message to target a subset of the HW resources
### Problem

Some architectures like the Tofino Native Architecture (TNA) for the Barefoot Tofino chip consist of multiple parallel hardware pipelines, instead of the usual one. They are attached to distinct sets of front panel ports, can be individually programmed and have independent state. Currently there is no mechanism in P4Runtime to target such a subset of the pipelines.

Examples
- Asymmetric table entries
  In some cases table entries only apply to a subset of ports (e.g. setting VLAN id based on egress port), but all rules still are inserted into all pipelines. By selectively inserting there entries in only the relevant subset, resources can be used more efficiently.
- Registers & counters
  Since the pipelines share no state, the registers and counters in them are not in sync. While this can be mitigated for counters (totaling them up), this does not make sense for general purpose registers (e.g. calculating flow hashes).

### Proposed changes

Adding an optional scope message to the `Entity` allows targeting on both the read and write path. The absence of a Scope message specifies the current behavior.

```proto
message Entity {
  message Scope {
    repeated uint32 scope = 1;
  }
  oneof entity {
    ExternEntry extern_entry = 1;
    TableEntry table_entry = 2;
    ActionProfileMember action_profile_member = 3;
    ActionProfileGroup action_profile_group = 4;
    MeterEntry meter_entry = 5;
    DirectMeterEntry direct_meter_entry = 6;
    CounterEntry counter_entry = 7;
    DirectCounterEntry direct_counter_entry = 8;
    PacketReplicationEngineEntry packet_replication_engine_entry = 9;
    ValueSetEntry value_set_entry = 10;
    RegisterEntry register_entry = 11;
    DigestEntry digest_entry = 12;
  }
  Scope scope = 13;
}
```

### Alternatives

Externs provide a way to extend the P4RT protocol unilaterally for a platform/architecture/vendor. Since the details are then defined by the specific implementation, interoperability between SwitchOSs and control planes might be lower than with a standardized solution.

pudelkoM:
TODOs:

- [ ] Update spec

Fixes #316

### DIFF
--- a/proto/p4/v1/p4runtime.proto
+++ b/proto/p4/v1/p4runtime.proto
@@ -91,7 +91,7 @@ message WriteRequest {
     // Optional. Every data plane packet is guaranteed to be processed according
     // to table contents before the batch began, or after the batch completed
     // and the operations were programmed to the hardware.
-    // The batch is therefore treated as a transaction. 
+    // The batch is therefore treated as a transaction.
     DATAPLANE_ATOMIC = 2;
   }
   Atomicity atomicity = 5;
@@ -120,6 +120,11 @@ message Update {
   Entity entity = 2;
 }
 
+// A scope allows limiting a request to a subset of the hardware resources.
+message Scope {
+  repeated uint32 scope = 1;
+}
+
 message Entity {
   oneof entity {
     ExternEntry extern_entry = 1;
@@ -135,6 +140,7 @@ message Entity {
     RegisterEntry register_entry = 11;
     DigestEntry digest_entry = 12;
   }
+  Scope scope = 13; // Optional
 }
 
 message ExternEntry {
@@ -708,6 +714,7 @@ message ForwardingPipelineConfig {
     uint64 cookie = 1;
   }
   Cookie cookie = 3;
+  Scope scope = 4; // Optional
 }
 
 message GetForwardingPipelineConfigRequest {


antoninbas:
@pudelkoM thanks for submitting this, we discussed the changes at the P4 API WG meeting today (10/23/2020). I think the takeways are:

 * there seems to be consensus that adding the `scope` field to the `Entity` message is useful (based on current use cases on Stratum) and low-risk
 * things are not as straightforward for the `ForwardingPipelineConfig` message change:
    - for devices like Tofino, physical pipes cannot be configured independently: it has to be done at the same time (in other terms, changing the forwarding config for one pipe cannot be done without affecting the config of other pipes); to accommodate this restriction, the `ForwardingPipelineConfig` message could be changed to support multiple "configurations"
    - before even trying to address the above bullet point however, we need to look into which use cases we want to support and decide if we want 1) a global P4Info message which includes object definitions across all logical pipelines, or 2) one P4Info message per logical pipeline definition. Terms have yet to be defined, but a logical pipeline definition can be mapped to multiple physical pipelines, which means that these physical pipelines implement the same P4 logic (same set of P4 objects), and can be programmed (table entries) symmetrically or asymmetrically. It seems that at the moment, 1) covers the known Stratum use case. However, with 1), it is not possible without knowing how to interpret the target-specific device config (`p4_device_config`) to determine whether a write operation on an `Entity` is valid based on the P4 object id & the `scope` value. 2) could support compiling P4 programs from 2 different stakeholders independently and then using them for different pipelines (they can share the same id space).

I believe we can start with adding a `scope` field to the `Entity` message, without any other change. However, we still need to update the specification accordingly to 1) define the appropriate terminology, 2) present some multi-pipe device model that users can refer to, 3) based on the device mode, describe which use cases we are trying to support and which are out-of-scope.

@bocon13 @ccascone @saynb - hopefully this is not to inaccurate based on our discussions, please feel free to commen

chrispsommers:
Hi @pudelkoM do you still want to move this forward, or is it time to retire it?

chrispsommers:
I'm doing some "housecleaning." @jafingerhut @smolkaj @antoninbas @bocon13 Is @pudelkoM still participating? No response to my inquiry 5 weeks ago. I am inclined to close this soon.

bocon13:
@chrispsommers I think we can close this. @pudelkoM is working on other projects now. We can also re-open if needed in the future. Thanks!

PR #326, opened 2020-11-20
Clarify spec: annotation locations are optional.

### LINKED ISSUE #311: spec doesn't mention whether annotation_locations is optional
The spec talks about ``annotations`` and ``annotation_locations`` and I cannot see how it says clearly one way or another whether ``annotation_locations`` is mandatory for each annotation. If locations are meant to be optional, it would be nice to say so. (And if they're meant to be mandatory, then https://github.com/p4lang/p4c/pull/2419 seems more important.)

smolkaj:
Fixes #311.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,9 +146,9 @@ processes.
 
 When updating the Protobuf files in a pull request, you will also need to update
 the generated Go files, which are hosted in this repository under
-[go/](go/). This can be done easily by running `./build/update_go.sh`, providing
-docker is installed and your user is part of the "docker" group (which means
-that the `docker` command can be executed without `sudo`).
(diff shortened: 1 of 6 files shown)


smolkaj:
@antoninbas Can you remind me how to update all generated files?

antoninbas:
> @antoninbas Can you remind me how to update all generated files?

It's documented at the end of the README

> When updating the Protobuf files in a pull request, you will also need to update the generated Go files, which are hosted in this repository under go/. This can be done easily by running ./build/update_go.sh, providing docker is installed and your user is part of the "docker" group (which means that the docker command can be executed without sudo).

smolkaj:
Thanks, done.

antoninbas (approved):
LGTM

PR #328, opened 2020-12-02
Add Python setuptool support

### LINKED ISSUE #324: Add Python package support?
Hello maintainers,

I open this issue to see if anyone is interested in this feature.

Currently, I am maintaining some projects which use the P4Runtime Python library(_pb2.py) and try to improve the way to create the P4Runtime python library.
So I made few code changes to my fork to support the Python package (using pip/setup.py to install)
https://github.com/Yi-Tseng/p4runtime/commit/f623d27aa6ffde2c084e49cceb4ff031b02b0e73

This means developers can use the following command to install the P4Runtime python library:

```bash
pip3 install git+https://github.com/Yi-Tseng/p4runtime.git@pip-support
```

This package only includes the P4Runtime python library, but the `setup.py` script will install some dependencies such as:

 - protobuf
 - grpcio
 - googleapis-common-protos (code_pb2 and status_pb2, required by P4Runtime)

We can also upload the package to `pypi.org` if merged, so people can simply use `pip install p4runtime` to install the library.

Let me know if anyone has any concerns.

Thanks

Yi-Tseng:
Allows developers to install generated Python code via pip or setup.py

Fixes #324


setup.py
```diff
@@ -0,0 +1,25 @@
+from setuptools import setup, find_packages
+
+setup(
+    name = "p4runtime",
+    version = "1.2.0",
```
antoninbas, line 5: s/1.2.0/1.3.0

setup.py
```diff
@@ -0,0 +1,25 @@
...
+    packages = find_packages("py"),
+    install_requires = [
+        "protobuf>=3.5.0",
```
antoninbas, line 8: nit: inconsistent spacing around " >= " compared to the other dependencies below

antoninbas, line 8: how are the versions chosen? I am asking because they don't necessarily match the versions used by p4lang, and the version of protobuf here doesn't exactly match the version of protoc used to generate the Python files.

The version of gRPC required is also much more recent than the version of protobuf that's required. Maybe we could use `"grpcio >= 1.17"` so that the version used by the p4lang toolchain falls into the acceptable range?

README.md
```diff
@@ -145,8 +145,8 @@ processes.
    approve.
 
 When updating the Protobuf files in a pull request, you will also need to update
-the generated Go files, which are hosted in this repository under
-[go/](go/). This can be done easily by running `./codegen/update_go.sh`,
+the generated Go and Python files, which are hosted in this repository under
+[go/](go/) and [py/](py/). This can be done easily by running `./codegen/update_[go/py].sh`,
```
antoninbas, line 149: I think some unification can be done in all scripts (update / check) so that contributors only have one script to run and so that we don't do duplicate work in CI. However, this can wait if you don't want to address this in this PR...

setup.py
```diff
@@ -0,0 +1,25 @@
+from setuptools import setup, find_packages
```
antoninbas, line 1: I image that it's not possible for this file to live under the `py/` directory?

antoninbas (commented):
I realize I didn't do it for Go, but do you think you could add a small section in the README to describe how to use / install the Python package in other projects?

setup.py
Yi-Tseng: The version I am using is based on the version I got from the Stratum project. But I think it is fine to use the one you suggest. (grpcio 1.17.2 and protobuf 3.6.1 according to p4lang/third-party)

setup.py
Yi-Tseng: Yes, we actually can place this file to the `py/` directory, will do.

README.md
Yi-Tseng: Done, now we only have `CI/check_codegen.sh` and `codegen/update.sh`

CI/deploy_pypi.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
antoninbas, line 1: can we automate running this script in Travis or Github actions every time a tag is created?

antoninbas (approved):
LGTM

CI/deploy_pypi.sh
Yi-Tseng: Yes, we should do that, let me try to add that to the Travis config.
By the way, this script requires a PyPI API token.
Currently, I am using mine to upload to the PyPI (https://pypi.org/project/p4runtime/) and I own this package.
I can move the ownership to anyone and add the token for Travis bot to the PyPI

antoninbas: I think Travis has a mechanism to encrypt secrets / tokens, feel free to keep using yours unless this is an issue for you.

If you could add me as a maintainer to the pypi project - AntoninBas (`antonin.bas[at]gmail.com`) - in case some changes are required in the future, it would be great. Thanks.

.travis.yml
```diff
@@ -71,3 +71,13 @@ deploy:
     on:
       repo: p4lang/p4runtime
       branch: master
+  - provider: pypi
+    username: "__token__"
+    password:
+      secure: ""
```
antoninbas, line 77: @Yi-Tseng this should be your encrypted token?

CI/deploy_pypi.sh
Yi-Tseng: Is this yours? (I need to use the username to invite users to the project)
https://pypi.org/user/AntoninBas/

antoninbas: yes that's me

antoninbas (approved):

PR #317, opened 2020-09-25
Deprecate watch in favor of watch_port.

stefanheule:
Fixes https://github.com/p4lang/p4runtime/issues/307

PR #331, opened 2021-01-15
Deprecate egress_port (uint32) in favor of egress (bytes) in Replica

### LINKED ISSUE #329: Should the egress_port field in the Replica message also be changed to a bytes string
https://github.com/p4lang/p4runtime/pull/317 deprecated the `watch` field (`int32`) in favor of a `watch_port` field (`bytes`) for action profile group programming.

However, there is another field used to represent ports, that is still using an integral type (for some reason it is `uint32`, while `watch` was using `int32`...): https://github.com/p4lang/p4runtime/blob/bd19174b0f8a99dd0ea39a185dae311e5f8b3606/proto/p4/v1/p4runtime.proto#L406-L410

It seems that the fields play a similar role and that one may want to leverage translation for ports when programming packet replication / cloning, so it seems that we should make the same change as for `watch`. @stefanheule is there a reason why we didn't make the same change for the `egress_port` field at the time? I don't recall the topic even coming up but I may have forgotten...

antoninbas:
This is to ensure consistency with the watch_port field.
See https://github.com/p4lang/p4runtime/pull/317

Fixes #329


stefanheule (approved):
Thanks!

proto/p4/v1/p4runtime.proto
```diff
@@ -405,15 +405,19 @@ message PacketReplicationEngineEntry {
 
 // Used for replicas created for cloning and multicasting actions.
 message Replica {
-  uint32 egress_port = 1;
+  oneof port_kind {
```
bocon13, line 408: Is the `oneof` necessary here given that `egress_port` is deprecated? It shouldn't be harmful, just wondering what the best practice here is.

bocon13 (approved):
Looks good. One minor question inline.

smolkaj:
Bump. We are starting to use multicast. Would be great to get this submitted.

antoninbas:
> Bump. We are starting to use multicast. Would be great to get this submitted.

I don't mind merging this as it is. I know that other parts of the spec also needs to be tackled: the `Index` message, used to address Counter and Meter entries, should also support translation and its type should be changed to `bytes` (indexing a counter using a port value is a pretty common scenario). However, that can be done in a future PR.

proto/p4/v1/p4runtime.proto
antoninbas: That's a good point, but I'm just sticking to what we did for watch port.

antoninbas:
@smolkaj I rebased the PR, could you take a look?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4309,7 +4309,7 @@ entity {
   packet_replication_engine_entry {
     clone_session_entry {
       session_id: 100
-      replicas { egress_port: 0xfffffffd instance: 1 } # to CPU
+      replicas { port: \xff\xff\xff\xfd instance: 1 } # to CPU
```
smolkaj, line 4312: Nit: add quotes, "\xff\xff\xff\xfd"

smolkaj (approved):
Thanks @antoninbas !

PR #334, opened 2021-01-22
Update versions of dependencies for Bazel build

antoninbas:
Our CI uses the latest available version of Bazel, which is now 4.0.
With this new Bazel release, the build started breaking, which required
upgrading dependencies to the following versions:
 * Protobuf: v3.13
 * gRPC: v1.35
 * googleapis: top-of-tree
 * rules_proto: top-of-tree

We also start running the Github workflow as a CRON job to ensure that
we catch these failures early in the future (in this case, the failure
was detected because of the p4lang/PI Bazel CI build).

### DIFF
--- a/.github/workflows/ci-build-proto.yml
+++ b/.github/workflows/ci-build-proto.yml
@@ -5,6 +5,8 @@ on:
     branches: [ master ]
   pull_request:
     branches: [ master ]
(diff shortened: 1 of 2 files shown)


smolkaj (approved):
LGTM, thanks!

PR #335, opened 2021-01-22
Improve CI build times through smarter caching

smolkaj:
GitHub's caching mechanism never overwrites cache entries for existing keys. Since we were using a static key, this caused our cache to never be refreshed. With this PR, a new cache entry is generated whenever any Bazel files (including dependencies) change.


antoninbas (approved):
LGTM

antoninbas:
@smolkaj could you rebase the PR so that the CI Bazel build can pass?

smolkaj:
I rebased, now it builds. Ready to be merged.

PR #338, opened 2021-02-12
Add redirect from master to main

bocon13:


antoninbas:
doesn't seem to be working, I get a 404

antoninbas:
@bocon13 is there no way to redirect https://p4.org/p4runtime/spec/master/P4Runtime-Spec.html and https://p4.org/p4runtime/spec/master/P4Runtime-Spec.pdf, only https://p4.org/p4runtime/spec/master/ ?

bocon13:
I missed those. Thanks!

The PDF redirect is tricky, let me know if you have another idea:
https://p4.org/p4runtime/spec/master/P4Runtime-Spec.pdf

antoninbas:
Thanks for taking care of this. I think the current solution is fine.

PR #343, opened 2021-03-12
Specify Read behavior without P4Info. Fixes #342.

### LINKED ISSUE #342: Read should return FAILED_PRECONDITION if there is no P4Info
I believe the P4RT spec says nothing about the behavior of the switch for Read if there is no P4Info configured yet. We should clearly specify the desired behavior. I suggest we return FAILED_PRECONDITION, since reading table entries requires to know what tables there are.

stefanheule:


antoninbas (approved):
LGTM

stefanheule:
Could we merge this?

antoninbas:
Yes I think it stayed open long enough to allow time for reviews :)

Issue #345, opened 2021-03-17
Read table entries from a particular role only


konne88:
Right now, reads return all table entries. It's very common that a controller only wants to read the table entries associated with its own role. So it would be great if the Read request could filter all table entries that don't match the specified role.

message ReadRequest {
  uint64 device_id = 1;
  // if specified, only table entries programmed by this role will be returned.
  string role_name = 3;
  repeated Entity entities = 2;
}

jafingerhut:
@konne88 Did you mean to assign this to me?  I suppose I can look into what P4Runtime API spec changes would be required here and try to write them up.  Just didn't expect it :-)

konne88:
Sorry, wasn't so much meant as you need to review this, but more as a "this is what's going on, just in case you want to know".

jafingerhut:
FYI, you can ping someone in a Github comment by putting an @ before their Github user id, too.

konne88:
@antoninbas @stefanheule 
What do you think about this idea? If this seems reasonable, I'm happy to send out the PR to change the spec.

stefanheule:
I fully support this, reading entries not for your own role is pretty annoying and does not seem like it would ever be beneficial.

Issue #344, opened 2021-03-17
Using strings for role


konne88:
Could we add string based roles to P4Runtime. The idea is that a string is much more readable (e.g. "tunnel manager") than to use an id (e.g. 1).

```
message WriteRequest {
  uint64 device_id = 1;
  uint64 role_id = 2;          // deprecated
  string role_name = 6;     // new
  Uint128 election_id = 3;

// ...

message SetForwardingPipelineConfigRequest {
  // ...
  uint64 device_id = 1;
  uint64 role_id = 2;          // deprecated
  string role_name = 6;     // new
  Uint128 election_id = 3;
  Action action = 4;
  ForwardingPipelineConfig config = 5;
}
```

konne88:
@antoninbas @stefanheule 
What do you think about this idea? If this seems reasonable, I'm happy to send out the PR to change the spec.

stefanheule:
Seems very reasonable, I'm a big fan of making things easier to debug/understand. If someone still prefers an integer as the role, they can still do that by encoding it as a byte-string (or even a regular string), so this is not limiting any prior usecases either.

jafingerhut:
It appears that performance concerns are less of an issue for the role, since it is only once per WriteRequest as probably the most frequent message it appears in.

This might tread into "number of angels that can dance on the head of a pin" kinds of discussion, but does this mean that a P4Runtime server must continue to support role_id until and unless it upgrades to a version of P4Runtime API that explicitly breaks backwards compatibility and drops role_id completely?

konne88:
Let's discuss this in tomorrow's meeting. In general, the way we've handled this is by returning UNIMPLEMENTED for any feature that we don't support (of which there are many).

PR #346, opened 2021-04-08
Add support for string roles & roles in reads

konne88:
This addresses bugs:
https://github.com/p4lang/p4runtime/issues/344
https://github.com/p4lang/p4runtime/issues/345


konne88:
Antonin, I just got the tests to pass. Can you review. Thanks!

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5168,7 +5167,7 @@ message SetForwardingPipelineConfigRequest {
     RECONCILE_AND_COMMIT = 5;
   }
   uint64 device_id = 1;
-  uint64 role_id = 2;
+  uint64 role = 2;
```
stolsma, line 5172: Again doesn't compare to pb file, should be 'string role = 6' I think?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4728,7 +4727,7 @@ defined as follows:
 ~ Begin Proto
 message WriteRequest {
   uint64 device_id = 1;
-  uint64 role_id = 2;
+  uint64 role = 2;
```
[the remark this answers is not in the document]
konne88: Oh yeah, sorry I totally forgot that.

docs/v1/P4Runtime-Spec.mdk
konne88: Thanks for catching this!

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4729,7 +4729,7 @@ defined as follows:
 ~ Begin Proto
 message WriteRequest {
   uint64 device_id = 1;
-  uint64 role = 2;
+  string role = 2;
```
stolsma, line 4732: The number two needs to be 6 too...😉

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -5169,7 +5169,7 @@ message SetForwardingPipelineConfigRequest {
     RECONCILE_AND_COMMIT = 5;
   }
   uint64 device_id = 1;
-  uint64 role = 2;
+  string role = 2;
```
stolsma, line 5172: Same here! 2 needs to be 6...😎

docs/v1/P4Runtime-Spec.mdk
konne88: It's a good question of whether we want to keep the protos in the spec completely in sync with the real protos. As a middle ground for now, let me update the ids, but not add the deprecated fields in the spec (even though they are in the proto). I don't feel strongly about this though, happy to add the deprecated fields as well.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -767,10 +767,9 @@ from making changes or even reading the state of the server.
 ## Default Role
 
 A controller can omit the role message in `MasterArbitrationUpdate`. This
-implies the "default role", which corresponds to "full pipeline access". This
-also implies that a default role has a `role.id` of 0 (default). If using a
-default role, all RPCs from the controller (&eg; `Write`) must set the `role_id`
-to 0.
+implies the "default role", which corresponds to "full pipeline access".
+If using a default role, all RPCs from the controller (&eg; `Write`) must
```
antoninbas, line 771: Should we mention explicitly that the empty string is therefore reserved for the default role and that it is not a valid name for a user-defined role? I noticed that you remove the sentence about role id 0.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4728,7 +4729,7 @@ defined as follows:
 ~ Begin Proto
 message WriteRequest {
   uint64 device_id = 1;
-  uint64 role_id = 2;
+  string role = 6;
```
antoninbas, line 4732: I am fine with the current solution of not including deprecated fields here. it is consistent with the fact that we do not describe deprecated fields such as `role_id` in the spec.

antoninbas (commented):
could you squash the commits and include the following in the commit message body:
```
Fixes #344 
Fixes #345 
```
It will link the PR to these issues properly and close them automatically when the PR is merged. Thanks.

docs/v1/P4Runtime-Spec.mdk
konne88: I'm pretty sure that even in protobuf v3, empty string "" is different from unset string. Therefore, I think "" need not be a special case, and can just be like any other role.

konne88:
I squashed the commits.

docs/v1/P4Runtime-Spec.mdk
antoninbas: No, string is a scalar field and empty string is the default
https://developers.google.com/protocol-buffers/docs/proto3#default

antoninbas:
@konne-google you have merged conflicts, as I had to merge a separate PR

docs/v1/P4Runtime-Spec.mdk
konne88: Oh you are right, thanks! I though strings were different.

antoninbas (approved):
LGTM

konne88:
Antonin, I don't think I can merge the request (The base branch restricts merging to authorized users.), could you do it for me? There is still a failing test, but I don't think I can fix it, it says:

docker: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit.

antoninbas:
@konne88 I restarted the job, I'll merge when it passed

PR #347, opened 2021-04-13
Clarify that write interts can use more specific error codes.

rhalstea:
Since MODIFY allows for more specific error codes where applicable it seems like the intent may have been to allow INSERTS the same discretion?

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -4785,10 +4785,10 @@ An update can be one of the following types:
 * `INSERT`: Inserts the given P4 entity in the entity container.
   The `entity` field always specifies the full state of the P4 entity.
   If the entity already exists, an `ALREADY_EXISTS` error is returned, and
-  the existing entity remains unchanged.
-  If the entity is malformed, an `INVALID_ARGUMENT` error is returned.
-  If the entity cannot be inserted because the container is already full,
-  a `RESOURCE_EXHAUSTED` error is returned.
+  the existing entity remains unchanged. If the entity is malformed, an
+  `INVALID_ARGUMENT` error is usually returned (unless a more specific error
+  code applies [@gRPCStatusCodes]). If the entity cannot be inserted because the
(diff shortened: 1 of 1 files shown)


antoninbas (approved):
LGTM
I think there is no harm in merging this. But I wonder if there is ever a case where something else than INVALID_ARGUMENT should be returned (whether for INSERT or MODIFY). Maybe OUT_OF_RANGE, but according to the gRPC documentation:
> Unlike INVALID_ARGUMENT, this error [OUT_OF_RANGE] indicates a problem that may be fixed if the system state changes

stefanheule:
Some examples:
- Insert of an action that doesn't exist: NOT_FOUND.
- Modify of a table entry that uses an action set_nexthop_id where the nexthop doesn't exist (some targets may have checks like this): NOT_FOUND.

rhalstea:
Thanks Antonin & Stefan,

The NOT_FOUND for set_nexthop_id example also works on insert, and could signal the controller isn't ordering it's requests properly.

antoninbas:
> Some examples:
> 
> * Insert of an action that doesn't exist: NOT_FOUND.

I would be more tempted to return INVALID_ARGUMENT here, and save NOT_FOUND for entities created at runtime by the controller

> * Modify of a table entry that uses an action set_nexthop_id where the nexthop doesn't exist (some targets may have checks like this): NOT_FOUND.

That one makes sense to me, although it is target-specific

stefanheule:
> > Some examples:
> > 
> > * Insert of an action that doesn't exist: NOT_FOUND.
> 
> I would be more tempted to return INVALID_ARGUMENT here, and save NOT_FOUND for entities created at runtime by the controller

Why is that? Seems like exactly what NOT_FOUND is supposed to be used for, no? Generally INVALID_ARGUMENT should only be used if there is no more specific code.

antoninbas:
@stefanheule  You're right. If the action doesn't exist at all, I would return NOT_FOUND. If the action exists but is not in the action list for that table, I would return INVALID_ARGUMENT.

PR #348, opened 2021-04-20
Enabling arenas in runtime proto

mkruskal-google:
This will be enabled by default in future versions, and is necessary to generate protobuf arena methods.


antoninbas (commented):
Is there any drawback to enabling arenas unconditionally?

mkruskal-google:
Looking at when this was flipped on by default back in early 2020, it looks like the main drawback is that the binary size increases a bit.  This should be marginal though, and this flag will be deprecated and defaulted to true in future protobuf versions.

stefanheule (approved):

antoninbas (approved):

mkruskal-google:
Are there any outstanding issues?  I don't seem to have permission to merge

antoninbas:
@mkruskal-google, no I was giving other folks a chance to review. I think we can proceed with the merge.

mkruskal-google:
Great, thanks Antonin!

PR #351, opened 2021-05-19
Add Go proto build rule for p4info

sthesayi:
This PR adds a build rule for the p4info.proto to generate the Go library version.


smolkaj (approved):
LGTM.
@antoninbas could you take a look?

antoninbas:
The PR itself looks good to me, but the build is failing: `ERROR: /home/runner/work/p4runtime/p4runtime/proto/BUILD.bazel:63:1: name 'go_proto_library' is not defined (did you mean 'cc_proto_library'?)`

Please also take a look at the CLA

smolkaj (changes requested):
Undo LGTM, look like we need additional changes to make things build.

sthesayi:
I believe to have fixed the build issue. Sorry for the premature request as it was my first CL and I forgot to mark it as draft. I am not sure why it still says "Waiting for status to be reported" but the CL build is passing now. I also fixed the CLA issue.

sthesayi:
> Undo LGTM, look like we need additional changes to make things build.

This is good for second review. Thanks

smolkaj (approved):
LGTM modulo the go.{mod, sum} changes, since I'm not a go user.
@antoninbas could you take a look.

go.mod
```diff
@@ -3,7 +3,11 @@ module github.com/p4lang/p4runtime
...
 require (
-	github.com/golang/protobuf v1.4.0
-	google.golang.org/genproto v0.0.0-20200413115906-b5235f65be36
+	github.com/golang/protobuf v1.4.3
+	github.com/google/go-cmp v0.5.1 // indirect
+	golang.org/x/sys v0.0.0-20200930185726-fdedc70b468f // indirect
+	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
+	google.golang.org/genproto v0.0.0-20200526211855-cb27e3aa2013
 	google.golang.org/grpc v1.28.1
+	google.golang.org/protobuf v1.25.0 // indirect
```
antoninbas, line 12: I am also unsure about the go.mod changes. I tried a few things (e.g., removing all required modules and running `go mod tidy`) but I could never get these indirect dependencies. Could you explain how they got there?

sthesayi: I ran the codegen/update.sh. Without that the build was failing because the Go modules were not up to date.

sthesayi: Also some of these might be due to Google auto updater which is defaulted in my emacs. Is there an issue with keeping all of these dependencies up-to-date? I can always try to handcraft the go.mod to remove the indirect ones and see if it is still ok with the build but I am a bit afraid messing with this by hand.

antoninbas (approved):

Issue #364, opened 2021-10-27
Allow Action Selectors to specify a maximum number of members.


jonathan-dilorenzo:
Currently, Action Selectors can only specify the maximum weight of members within a group and the maximum total weight of all members (see image below). It would be useful to be able to specify the maximum number of members for Action Selectors as well (much like for Action Profiles).

![image](https://user-images.githubusercontent.com/2077197/139162318-97eaab17-4dc3-4aa8-956c-9670b7d8299b.png)

Non-breaking change suggestion:
- Add a `max_num_members` field, which specifies the maximum number of total members for Action Selectors.

Breaking change suggestions:
1. Replace `size` with the `max_num_members` field from the above suggestion, unifying the semantics for Action Selectors and Action Profiles and clarifying the meaning of the name in the presence of both weight and member number. Then, add a `max_total_weight` field which has the semantics that `size` currently has for Action Selectors (i.e. maximum total weight) and should be `0` for Action Profiles.

2. Instead of combining Action Profiles and Action Selectors into one message `ActionProfile`, separate them out. This simplifies `ActionProfile` since it only needs `preamble`, `table_ids`, and `size` (ideally renamed for clarity). Action selectors can then use the suggestion from just above, co-opting all but the `with_selector` field from the current `ActionProfile` message and adding the `max_total_weight` field (and ideally renaming `size`).

antoninbas:
This sounds good to me. As I mentioned in today's meeting, having `max_group_size` be the cumulative sum of member weights was implementation-driven when we first added support for `ActionSelector` in P4Runtime (the implementation was replicating members according to their weight to achieve the desired load-balancing behavior). If an implementation can benefit from having this new piece of information, it is reasonable to add it.

@jonathan-dilorenzo could you submit a PR with this change (the non-breaking one).
I would recommend that for an `ActionProfile`, we define `max_num_members` to be the same as `size`. Also, I assume we need a PSA annotation so that this value can be set from the P4 program for Action Selector extern instances?

PR #369, opened 2021-11-22
Add max_num_members to Action Profiles.

jonathan-dilorenzo:
See issue #364 for more information.

It is a bit unnatural for a proto field to default to the same value as another field, but it seems like the right thing to do if we want the semantics to be the same for `standard` Action Profiles. WDYT @antoninbas and @smolkaj?

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -6167,7 +6167,7 @@ the purpose of adding features for the P4Runtime API.
 | `@description`           | See section [#sec-annotating-p4-entities-with-documentation] |
 | `@id`                    | See section [#sec-id-allocation] |
 | `@max_group_size`        | See sections [#sec-p4info-action-profile], [#sec-action-profile-group-programming] |
-| `@max_num_members`       | See sections [#sec-p4info-action-profile], [#sec-action-profile-group-programming] |
+| `@max_num_members`       | See section [#sec-p4info-action-profile] |
 | `@pkginfo`               | See section [#sec-annotating-p4-code-with-pkginfo] |
 | `@p4runtime_translation` | See sections [#sec-user-defined-types], [#sec-translation-of-port-numbers] |
 +--------------------------+---------------------------------------+
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -1595,6 +1595,12 @@ The `ActionProfile` message includes the following fields:
   Action Profile can hold, or, in the case of an Action Selector, the maximum
   sum of all member weights across all selector groups.
 
+* `max_num_members`, an `int64` representing the maximum number of member
+  entries that an Action Profile or Action Selector can hold. For Action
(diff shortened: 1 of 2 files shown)


jonathan-dilorenzo:
The build error seems to be due to the Go version being outdated. Unfortunately, I can't easily get docker on my work computer. If someone else could run the codegen/update.sh that is suggested (well, update_go.sh is suggested, but it doesn't exist), and update the PR, that'd be fantastic.

antoninbas (commented):
Just to clarify but I am pretty sure that this is what you intended: if the same member is added to 2 different groups, it still counts as only 1 against the `max_num_members` limit?

One more question: when using "one-shot" programming (with `ActionProfileActionSet`), how are members accounted for in that case? There aren't really any `Member` resources in that case, yet the definition of `size` and `max_group_size` can still hold because these limits apply to weights and not individual members. But with `max_num_members`, what happens when the same `Action` is repeated across `ActionProfileActionSet`s or even within the same `ActionProfileActionSet` (discouraged by the spec).

jonathan-dilorenzo:
Excellent questions that I should perhaps clarify.

1. `if the same member is added to 2 different groups, it still counts as only 1 against the max_num_members limit?` 
Yes, that is exactly my intention.

2. We would count every individual action as being a member since this is a conservative approach that should be a lower-bound regardless of how clever the switch is (unless the current definition of 'size' is a better abstraction for its resource constraints than the `max_num_members` is e.g. because it uses X weight 1 members to represent a single weight X action).

antoninbas:
@jonathan-dilorenzo sounds good to me. I think your changes to the spec should explicitly cover my 2nd question (one-shot programming) as it is not obvious.

I can help with Go files generation, but I will either need write access to your fork or you will need to reopen this PR from a branch in this repository (instead of from your fork).

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1595,6 +1595,12 @@ The `ActionProfile` message includes the following fields:
   Action Profile can hold, or, in the case of an Action Selector, the maximum
   sum of all member weights across all selector groups.
 
+* `max_num_members`, an `int64` representing the maximum number of member
+  entries that an Action Profile or Action Selector can hold. For Action
+  Profiles, this should always be equal to the size. PSA programs can use the
```
smolkaj, line 1600: Nit: maybe swap these two sentence? ("For Action Profiles ..." & "PSA programs")

smolkaj: Is `@max_num_members` specific to PSA programs? We're using v1model and are planning to use this, no?

jonathan-dilorenzo: Done on the nit. For PSA programs, I just used the same wording as in `max_group_size`. I'm honestly not sure why we talk about how to specify it in PSA programs in this part of the spec at all, though a significant portion of the section talks about PSA in various ways. 

WDYT?

jonathan-dilorenzo: Done in other PR #386, to be clear.

PR #372, opened 2022-01-20
Update versions of dependencies for Bazel build

antoninbas:
Our CI uses the latest available version of Bazel, which is now 5.0.
With this new Bazel release, the build started breaking, which required
upgrading dependencies to the following versions:
 * Protobuf: v3.18.1
 * gRPC: v1.43.0
 * googleapis: top-of-tree
 * rules_proto: top-of-tree
 * rules_go: v0.29.0


antoninbas:
Waiting for a new version of grpc to be released which would include this fix: https://github.com/grpc/grpc/pull/28406

PR #374, opened 2022-02-10
Use noninteractive mode when installing packages in Docker build

antoninbas:


antoninbas:
Superseded by https://github.com/p4lang/p4runtime/pull/375

PR #380, opened 2022-03-22
Build Python distribution before pushing to PyPI

antoninbas:
The pypa/gh-action-pypi-publish action doesn't take care of building the
distribution.
See
https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

### DIFF
--- a/.github/workflows/tag-uploads.yml
+++ b/.github/workflows/tag-uploads.yml
@@ -54,7 +54,27 @@ jobs:
     runs-on: [ubuntu-latest]
     steps:
     - uses: actions/checkout@v3
-    - uses: pypa/gh-action-pypi-publish@release/v1
+    - name: Set up Python
+      uses: actions/setup-python@v3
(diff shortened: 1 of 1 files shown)


antoninbas:
@Yi-Tseng does this look correct to you?

.github/workflows/tag-uploads.yml
```diff
@@ -54,7 +54,26 @@ jobs:
...
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
```
Yi-Tseng, line 74: I tried this command on my laptop, I think you need to `cd` to `py` directory first.

antoninbas: good catch, thanks

PR #390, opened 2022-05-07
Generate mocks for the C++ grpc library.

jonathan-dilorenzo:
This change will cause Bazel to automatically generate mocks for testing various interactions with gRPC calls.

The only semantic difference is the `generate_mocks = true`, the rest is automatic reordering by my auto-formatter :'(. 

We add and use this internally in Google, meaning that we can't open source some of our tests. I recently read that this should now be supported by Bazel.


smolkaj (approved):
This is great, thanks!

antoninbas (approved):
This is cool, do the mocks end up being generated by `protoc`?

jonathan-dilorenzo:
I believe so, yeah. They are definitely automatically generated, but I'm never quite sure which subcomponents do what when I call build ;).

PR #391, opened 2022-05-07
Redefines ActionProfiles to include the manner in which `size` and `max_group_size` should be calculated.

jonathan-dilorenzo:
This is a new proposal for how to represent the `max_num_members` field that we recently introduced. As I was trying to use it, I figured out that the semantics of `max_group_size` was non-sensical in conjunction with `max_num_members`.

It became clear that some sort of flag determining whether `size` and `max_group_size` should talk about total weight or total members is necessary.

Here I went with a slightly more generalized proposal wherein one might specify different methods of calculating the size of a group through a oneof message. 

I thought I'd just write it up real quick and get feedback on something concrete rather than go through the rounds. 

Also, @antoninbas, would you run the proto updating script again? Sorry about that.


proto/p4/config/v1/p4info.proto
```diff
@@ -265,15 +265,26 @@ message ActionProfile {
...
-  int64 max_num_members = 6;
-  // max number of weighted member entries in any given selector group, or 0 if
-  // the action profile does not have a selector
+  // 0 if the action profile does not have a selector. Otherwise, semantics as
+  // specified by `selector_group_size_computation` below.
   int32 max_group_size = 5;
+  // specifies the semantics of `size` and `max_group_size` above
+  message SumOfWeights {}
+  message SumOfMembers {}
+  oneof selector_group_size_computation {
```
smolkaj, line 278: Any strong reasons why you went with a oneof here?
Perhaps this should be an enum, with default value `SUM_OF_WEIGHTS = 0` for backward compatability?

jonathan-dilorenzo: Just poor memory. Should clearly be an enum. Will fix on Monday!

antoninbas:
@jonathan-dilorenzo let's discuss this on Friday at the WG meeting
I can generate the Proto files for you once we get consensus

smolkaj:
@antoninbas 's feedback:
- Use single annotation with two arguments.
- See if we can standardize those annotations in PSA spec.

jonathan-dilorenzo:
What do we think about `selector_group_size_computation` vs `selector_size_semantics` or something like this?

antoninbas:
It almost feels like `@max_group_size`, along with the new annotation should be standard PSA properties of the ActionSelection extern type. The rationale bing that `size` is a standard property, and that the semantics of `size` already depend on the "semantics mode" chosen (sum of weights vs distinct members). I'm tagging @jafingerhut here, even though he may be missing come context.

jonathan-dilorenzo:
Gentle reminder to take a look at this when you get a chance. Once accepted, I'll need to make some updates to the P4RT compiler to properly interpret the new annotations.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1595,21 +1595,29 @@ The `ActionProfile` message includes the following fields:
...
-  Action Selectors. If the annotation is omitted, the P4Info field will default
-  to 0.
+  Action Selector, its semantics is specified by the
+  `selector_size_semantics` value as described below.
+  The `max_group_size` must be no larger than `size`. PSA programs can use the
+  `@max_group_size` annotation to provide this value for Action Selectors.
+  If the annotation is omitted, the P4Info field will default to 0.
+
+* `selector_size_semantics`, an enum for Action Selectors that
+  specifies how `size` and `max_group_size` are calculated. It can be either:
```
smolkaj, line 1609: Nit: calculated -> interpreted?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1595,21 +1595,29 @@ The `ActionProfile` message includes the following fields:
...
+    * `SUM_OF_WEIGHTS`, indicating that `size` and `max_group_size` represent
+      the maximum sum of weights that can be present across all selector groups
+      and within a single selector group respectively.
+    * `SUM_OF_MEMBERS`, indicating that `size` and `max_group_size` represent
+      the maximum number of members that can be present across all selector
+      groups and within a single selector group respectively.
+
+  PSA programs can use the `@selector_size_semantics` annotation with one of
```
smolkaj, line 1617: Just to confirm, the syntax we have in mind here is
@selector_size_semantics(SUM_OF_WEIGHTS)
@selector_size_semantics(SUM_OF_MEMBERS)
?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3680,17 +3688,21 @@ does not need to keep track of group ids and member ids.
 
 One shots are programmed by choosing the `ActionProfileActionSet` message as the
 `TableAction`. The `ActionProfileActionSet` message consists of a set of
-`ActionProfileAction` messages, which in turn have the following fields:
+`ActionProfileAction` messages. This set should have cardinality no greater than
+`max_group_size` (if greater than 0) if `selector_size_semantics` is
```
smolkaj, line 3692: Can we replace "(if greater than 0)" with "(if max_group_size is nonzero)" to make the object explicit? 
I initial thought the implicit object was "cardinality".

proto/p4/config/v1/p4info.proto
```diff
@@ -265,15 +265,25 @@ message ActionProfile {
...
+  // `selector_size_semantics` below.
   int64 size = 4;
-  // max number of member entries across all groups
-  int64 max_num_members = 6;
-  // max number of weighted member entries in any given selector group, or 0 if
-  // the action profile does not have a selector
+  // 0 if the action profile does not have a selector. Otherwise, semantics as
+  // specified by `selector_size_semantics` below.
   int32 max_group_size = 5;
+  // specifies the semantics of `size` and `max_group_size` above
```
smolkaj, line 275: Nit: Move this comment down to the `selector_size_semantics` field.

smolkaj (changes requested):
Small nits, otherwise LGTM.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: I like it.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: Correct.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: Done.

proto/p4/config/v1/p4info.proto
jonathan-dilorenzo: Done

jonathan-dilorenzo:
I get an error when I try to run codegen: `error checking context: 'readdirent: input/output error'.`

Could I ask you to rerun it @antoninbas?

smolkaj (approved):

antoninbas:
@jonathan-dilorenzo I ran codegen and update the branch. The diff was very small compared to what you had:
```
$ git diff
diff --git a/go/p4/config/v1/p4info.pb.go b/go/p4/config/v1/p4info.pb.go
index 43064ee..c9f6707 100644
--- a/go/p4/config/v1/p4info.pb.go
+++ b/go/p4/config/v1/p4info.pb.go
@@ -288,7 +288,6 @@ func (ActionRef_Scope) EnumDescriptor() ([]byte, []int) {
        return file_p4_config_v1_p4info_proto_rawDescGZIP(), []int{9, 0}
 }

-// specifies the semantics of `size` and `max_group_size` above
 type ActionProfile_SelectorSizeSemantics int32

 const (
@@ -1503,7 +1502,8 @@ type ActionProfile struct {
        Size int64 `protobuf:"varint,4,opt,name=size,proto3" json:"size,omitempty"`
        // 0 if the action profile does not have a selector. Otherwise, semantics as
        // specified by `selector_size_semantics` below.
-       MaxGroupSize          int32                               `protobuf:"varint,5,opt,name=max_group_size,json=maxGroupSize,proto3" json:"max_group_size,omitempty"`
+       MaxGroupSize int32 `protobuf:"varint,5,opt,name=max_group_size,json=maxGroupSize,proto3" json:"max_group_size,omitempty"`
+       // specifies the semantics of `size` and `max_group_size` above
        SelectorSizeSemantics ActionProfile_SelectorSizeSemantics `protobuf:"varint,6,opt,name=selector_size_semantics,json=selectorSizeSemantics,proto3,enum=p4.config.v1.ActionProfile_SelectorSizeSemantics" json:"selector_size_semantics,omitempty"`
 }
```

antoninbas (approved):
LGTM
Are we planning to add a sentence about the behavior if a target / P4Runtime implementation doesn't support one semantics mode?

From the May meeting minutes, I see:

> The assumption is that the compiler or P4Runtime server should reject the P4 program or P4Info message if its ActionSelector implementation is at odds with the provided annotation.

I'm fine postponing this to a follow-up PR, or discussing it further if needed

jonathan-dilorenzo:
> Are we planning to add a sentence about the behavior if a target / P4Runtime implementation doesn't support one semantics mode?

Ooh. Good question. I guess I was imagining that being covered by statements like `Returns an INVALID_ARGUMENT error if ... the provided config cannot be realized.` in SetForwardingPipelineConfig. There are a bunch of ways in which a provided config can't be realized, but I'm not sure it's worth calling this one out specifically. What do you think?

antoninbas:
> There are a bunch of ways in which a provided config can't be realized, but I'm not sure it's worth calling this one out specifically. What do you think?

True, but most of the time I would expect it to be because of resource constraints, not necessarily because of missing functionality. In this case, I think it's more about missing functionality.
However, maybe rejecting the config because it cannot be realized is just "common sense", and it doesn't need to be called out explicitly. We can merge as is and see if a question ever comes up :)

jonathan-dilorenzo:
Ah, these are good points. I guess I think of it as common sense, but it's true that an easy mistake to make (without mentioning this explicitly) would be to just ignore this field, given that it hasn't existed until now, and assume that this remained a correct implementation of the specification.

PR #409, opened 2022-12-19
Fix bazel version to 5.4.0

smolkaj:
This is to work around breaking changes introduced by Bazel 6: bazelbuild/bazel#8622 (comment)

We can wait until the next nightly CI to see if we're actually affected, and only merge this if that's the case.


antoninbas:
@stefanheule don't we usually try to support the latest version of Bazel? Can't we follow the "Migration" steps described in the issue you referenced?

stefanheule:
Assuming this is meant for @smolkaj, not me? I'm no longer working on P4RT.

antoninbas:
@stefanheule yes, apologies. I typed `@stef` and selected the first result without thinking :). I'll remove you from the repo.

smolkaj:
Why not a guest appearance @stefanheule? ;)

smolkaj:
Yes @antoninbas, +1, I think we should generally try to keep up to date with the latest Bazel version.

In this case, I (i) didn't have the time to dig sufficiently deeper and (ii) suspect this is an issue caused by upstream dependencies. With a bit of luck, if we wait a couple of weeks, our upstream dependencies may have made the necessary changes, and we can then upgrade to their latest versions and undo my change.

So this is just meant as a temporary workaround, given that I will be OOO for a while.

antoninbas (approved):

Issue #408, opened 2022-12-09
Define a term "data plane volatile" in P4Runtime specification


jafingerhut:
This proposal arose in a P4.org API work group discussion on 2022-Dec-09 about read-write symmetry.

The following types of objects are known to be able to change their state by data plane actions alone, without any action being taken by a controller:

+ counters can be updated
+ P4 registers can be written
+ P4 tables with PNA's new `add_on_miss = true` table property can have new table entries added by data plane initiated actions
+ P4 tables with PNA's new `pna_idle_timeout = PNA_IdleTimout_t.AUTO_DELETE;` table property can have existing table entries removed by the data plane, if they exceed their configured idle time.
+ P4 tables with the previous new property, OR the existing value of `NOTIFY_CONTROL`, can have the idle timeout value configured for an entry changed by the data plane on its own.

Note: add-on-miss and auto-delete-on-idle-timeout are NOT exclusive to NICs.  There are many switch ASICs that can do this, too, often because of Ethernet MAC address learning of source MAC addresses, but other features like Netflow/sFlow, and others.

It would be good to define a term that covers this category of objects that break read-write symmetry because they are data plane volatile, and perhaps introduce a new P4Info property that indicates whether a particular P4 object is data plane volatile or not.  While it is true that ALL counters and P4 registers are data plane volatile, there can be P4 programs where some P4 tables are data plane volatile, while others are not, and it would be nice to be able to distinguish which P4 tables are which from P4Info file contents.

jafingerhut:
Some notes while thinking about what to say in the spec about data plane volatility.  I may propose a PR based on this soon.  Correction/comments/questions all welcome.

Read/write symmetry really only matters for things that can be written
by the controller.

TODO: Are all such things ones that can appear in a WriteRequest
message?  Or are there others?

Within a WriteRequest, there can be 0 or more Update messages.
Each Update message can contain one Entity:

```
message Entity {
  oneof entity {
    ExternEntry extern_entry = 1;
    TableEntry table_entry = 2;
    ActionProfileMember action_profile_member = 3;
    ActionProfileGroup action_profile_group = 4;
    MeterEntry meter_entry = 5;
    DirectMeterEntry direct_meter_entry = 6;
    CounterEntry counter_entry = 7;
    DirectCounterEntry direct_counter_entry = 8;
    PacketReplicationEngineEntry packet_replication_engine_entry = 9;
    ValueSetEntry value_set_entry = 10;
[...]

PR #412, opened 2022-12-26
Define the term data plane volatile and how it relates to read-write symmetry

jafingerhut:


jafingerhut:
Intended to resolve https://github.com/p4lang/p4runtime/issues/408

jafingerhut:
TODO: Does the MessageDifferencer function mentioned in the spec ignore all of the data plane volatile fields when comparing read values against write values, at least for things that are data plane volatile in PSA?

I would guess it would need to be updated for the new things that can be data plane volatile in PNA.

jafingerhut:
@smolkaj @jonathan-dilorenzo @verios-google FYI please read it over to see if you have any corrections/comments/questions.

smolkaj:
@verios-google for visibility

verios-google (commented):
Great change, good job including the PSA and PNA specific items.
I have a question, would the field `time_since_last_hit` (Protobuf with `elapsed_ns`) still be considered data plane volatile in non-PSA/non-PNA implementations? Or would that depend on the implementation?

jafingerhut:
> Great change, good job including the PSA and PNA specific items. I have a question, would the field `time_since_last_hit` (Protobuf with `elapsed_ns`) still be considered data plane volatile in non-PSA/non-PNA implementations? Or would that depend on the implementation?

It would depend upon the behavior of that architecture, but if you are asking about the v1model architecture, then yes it is data plane volatile.

I have a difficult time imagining an implementation that claims to support idle timeout in a way compatible with the existing defined P4Runtime API messages, including their documentation, where those fields were not data plane volatile.  But it seems safest to qualify the documentation with the architectures we know, rather than ones that we do not.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2185,6 +2185,114 @@ C++ API supports comparing messages while treating repeated fields as sets, so
 that different orderings of the same elements are considered equal. This method
 of comparing Protobuf messages may come at a cost in performance.
 
+### Data plane volatile objects
+
+An exception to read-write symmetry are objects whose contents or
```
smolkaj, line 2190: Nit: Should we briefly mention this exception and add a forward reference to this section at the beginning of the previous section, so people don't miss it?

smolkaj (approved):
This looks great from my PoV. Thanks Andy!

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Done

antoninbas (approved):
sorry for the delay, LGTM

I intend to merge this by end of day

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -2185,6 +2187,114 @@ C++ API supports comparing messages while treating repeated fields as sets, so
...
+field of a `TableEntry` can be modified by the data plane when packets
+match the entry.
+
+For a table with a direct meter associated with it, the
+`meter_counter_data` field of a `TableEntry` can be modified by the data
+plane when packets match the entry.
+
+For a PSA [@PSA] table with property `psa_idle_timeout` equal to
+`PSA_IdleTimeout_t.NOTIFY_CONTROL`, the data plane can modify the
+`elapsed_ns` field of a `TabeEntry` when _no_ packets match the entry
```
jonathan-dilorenzo, line 2220: nit: TabeEntry -> TableEntry

jonathan-dilorenzo (approved):
Super nice! Thanks Andy!

docs/v1/P4Runtime-Spec.mdk
jafingerhut: fixed

PR #414, opened 2023-02-07
changes in p4Runtime for Mirrir_for_splitArch

bharticemk:
Signed-off-by: bharticemk <richa.bharti@intel.com>

### DIFF
--- a/proto/p4/config/v1/p4info.proto
+++ b/proto/p4/config/v1/p4info.proto
@@ -106,6 +106,8 @@ message P4Ids {
 
     // externs for other architectures (vendor extensions)
     OTHER_EXTERNS_START = 0x80;
+    MATCH_LOOKUP_TABLE = 0x81;
+    TERNARY_MATCH_LOOKUP_TABLE = 0x82;
 
     // max value for an unsigned 8-bit byte
     MAX = 0xff;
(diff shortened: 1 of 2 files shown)


proto/p4/config/v1/p4info.proto
```diff
@@ -106,6 +106,8 @@ message P4Ids {
 
     // externs for other architectures (vendor extensions)
     OTHER_EXTERNS_START = 0x80;
+    MATCH_LOOKUP_TABLE = 0x81;
```
saynb, line 109: This will conflict with existing vendors having their externs starting with 0x81. Is this going to be PNA specific or being added to P4 language core ?
If PNA, then we should probably think along the lines of having a pna.p4 to extend this so that other possible vendor architectures can extend from 0x81 onwards.
If core, then why not 0x18 and 0x19 directly ?

proto/p4/config/v1/p4info.proto
```diff
@@ -114,6 +116,72 @@ message P4Ids {
...
+  p4.config.v1.Preamble preamble = 1;
+  repeated p4.config.v1.MatchField match_fields = 2;
+  message Param {
+    uint32 id = 1;
+    string name = 2;
+    uint32 bitwidth = 3;
+  }
+  repeated Param params = 3;
+  uint32 size = 4;
+}
```
saynb, line 143: Apart from this and the extern enum additions, I feel all the other additions to the p4info in this PR aren't required.

proto/p4/config/v1/p4info.proto
```diff
@@ -114,6 +116,72 @@ message P4Ids {
...
+
+message FieldMatch {
```
saynb, line 145: This is already there in p4runtime.proto

proto/p4/config/v1/p4info.proto
```diff
@@ -114,6 +116,72 @@ message P4Ids {
...
+  }
+  repeated Param params = 1;
+}
+
+message MvlutEntry {
+  uint32 mvlut_id = 1;
+  repeated FieldMatch match = 2;
+ MvlutParam param = 3;
+}
+
```
saynb, line 132: This is also to be added in p4runtime.proto

bharticemk:
decline the PR

PR #419, opened 2023-04-14
Support for adding GenericTable

saynb:
* Phase 1: 
  * Not extend any of the current RPC updates. Use INSERT, MODIFY, DELETE and Read RPC as is
  * Define Table categories/ properties
  * Define Entry and Field properties
  * Define Type system
  * Achieve Write/Read for simple PNA specific externs like ExactValueMatchLookupTable
  * Work on generic-table-dev
* Phase 2
  * Entry Handles
  * Table attributes (symmetric vs asymmetric)
  * New APIs
    * reset
    * add_or_mod
    * sync
    * usage_get
    * mod_inc
  * Table custom operations extensions. A way to define new operations via json modeling. For example, sync
* Phase 3
  * String representation of fields. Like IPv4 IPv6
  * Table notifications. Model notifications in a generic way

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2007,6 +2007,15 @@ The `Digest` message defines the following fields:
   notification using a `P4DataTypeSpec` message (see section on [Representation
   of Arbitrary P4 Types](#sec-representation-of-arbitrary-p4-types)).
 
+### `GenericTable` { #sec-p4info-generic-table}
+`GenericTable` messages are used to map non-PSA P4 externs or non-P4 target-
+specific features to their implementation in a generic way. The same can be achieved
+via Extern, but GenericTable provides a structured way in which every feature is
+represented as a set of match-fields and data-fields. The data consists of unions
+wihch are similar to actions. The best use of GenericTable in a server backend is
+with TDI (Table Driven Interface), but targets can use Generic table to map to their
+own specific feature implementations as well.
+
 ### `Extern` { #sec-p4info-extern}
 
 `Extern` messages are used to specify all extern instances across all extern
@@ -2041,6 +2050,9 @@ the following fields:
 
 If the P4 program does not include any instance of a given extern type, the
 `Extern` message instance for that type should be omitted from the P4Info.
+
+
+
 
 ## Support for Arbitrary P4 Types with P4TypeInfo
 
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -2008,47 +2008,7 @@ The `Digest` message defines the following fields:
   of Arbitrary P4 Types](#sec-representation-of-arbitrary-p4-types)).
 
 ### `GenericTable` { #sec-p4info-generic-table}
-`GenericTable` messages are used to map non-PSA P4-externs or non-P4 target-
-specific "fixed" features to their implementation in a generic way. The same can be achieved
-via `Extern`, but GenericTable provides a structured way in which every feature is
-represented as a set of match-fields and data-fields. The data consists of unions
-wihch are similar to actions. One use of GenericTable in a server backend is
-with TDI [Table Driven Interface](https://github.com/p4lang/tdi), but targets
-can use Generic table to map to their own specific feature implementations as
-well.
-It defines the following fields
-
-* `preamble`, a `Preamble` message with the ID, name, and alias of this 
-  GenericTable.
-
-* `match_fields`, a repeated field of type `MatchField` representing the data to
-  be used to construct the lookup key matched in this table. For information check
-  the match_fields info in the [Table section](#sec-table)
-
-* `union_refs`, a repeated `UnionRef` field representing the set of possible
-  unions for this table. Functionally, it behaves same as that of ActionRefs.
-  Hence it has been kept as ActionRefs itself.
-  Please check the action_refs info in the [Table section](#sec-table)
-
-* `const_default_union_id`, if this table has a constant default union, this
-  field will carry the `uint32` identifier of that action, otherwise its value
-  will be 0. A default union is the data when the key is null. It is similar to
-  a default action for Match Tables.
-  Being constant means that the control plane cannot
-  set a different default union at runtime or change the default union's
-  arguments.
-
-* `size`, an `int64` describing the desired number of table entries that the
-  target should support for the table.  See the "Size" subsection within the
-  "Table Properties" section of the P4~16~ language specification for details
-  [@P4TableProperties].
-
-* `is_const_table`, a boolean flag indicating that the table cannot be modified
-  by the control plane at runtime.
-
-* `other_properties`, an `Any` Protobuf message [@ProtoAny] to embed
-  architecture-specific or target-specific table properties [@P4TableProperties]
-  that the target wants to convey.
+See section [GenericTable p4info](#sec-p4info-generic-table) for more info
 
 ### `Extern` { #sec-p4info-extern}
 
@@ -2977,24 +2937,8 @@ tables {
 ~ End Prototext
 
 ### GenericData
-
-`GenericData` is an addition to p4data.proto- which aims at adding support for
-data types which aren't covered by P4Data. This incorporates all the data types
-used as primitive data types like bytes, float, strings.
-
-  * bitstring : Bytes type used for simple unsigned integers or bytes
-  * float : Floating type values
-  * bool : Boolean type values
-  * generic_struct : Used to define a structure that can contain
-    one or more of the other members. So one structure can contain
-    different members of different types in GenericData
-  * generic_list : Used to define a list of same types. Even though the def
-    is same as that of generic_struct, this only allows for a variable list
-    of the same type. It can also be a list of generic_structs
-  * string : Used to define control plane strings. The max string length is
-    defined by the size
-  * enum : String representation with which safe enums are realized
-  * enum_val : bytes value with which unsafe/serializable enums are defined
+See section [GenericData p4types](#sec-p4data-generic-data) for more info on the info type.
+See section [GenericData p4data](#sec-p4types-generic-data) for more info on the runtime type.
 
 # P4 Entity Messages { #sec-p4-entity-msgs}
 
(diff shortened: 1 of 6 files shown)


qobilidop:
I want to better understand this proposal with a concrete use case in mind, which is to [provide a P4Runtime API for custom hash extern to set its hash key from control plane](https://github.com/p4lang/pna/issues/19). Is this an intended use case of this proposal?

The challenge there was that, as @jafingerhut [commented](https://github.com/p4lang/pna/issues/19#issuecomment-1315817970), "if you define a new extern, there is no off-the-shelf way in P4 source code or P4Runtime API to define its control plane API". I think this proposal could enable providing such an off-the-shelf P4Runtime API for externs, right?

For example, for the following extern (adapted from [pna.p4](https://github.com/p4lang/p4c/blob/a56268106d7200490ed0049519a29ed327a595b7/p4include/pna.p4#L280-L300)):
```p4
extern ToeplitzHash<O> {
  Hash();
  void set_rss_key<K>(in K key);
  O get_hash<D>(in D data);
  O get_hash<T, D>(in T base, in D data, in T max);
}
```

I would expect to model an extern `ToeplitzHash` object as a GenericTable in P4Runtime, and use the following entry to set its RSS key:
```
generic_table_entry {
  table_id: ...  // Refers to the extern object
  // No match fields in this case
  table_data_union {
    union_id: ...  // Refers to the set_rss_key method
    params {
      param_id: ...  // Refers to the key param
      value: ...
    }
  }
}
```

I'm not very certain about this usage because there is no match fields in this case.

Another thing I'm not clear here is how to model the generic extern methods like `void set_rss_key<K>(in K key)` above? My tempted answer is that, if this method is to be used in the control plane only, maybe this method level generics shouldn't be allowed, because there is no way in the P4 code to tell what `K` is. Generics at the extern object level like the following one should be fine though:
```p4
extern ToeplitzHash<O, K> {
  Hash();
  void set_rss_key(in K key);
  O get_hash<D>(in D data);
  O get_hash<T, D>(in T base, in D data, in T max);
}
```

smolkaj:
I'll let @saynb answer himself, but regarding:

> Another thing I'm not clear here is how to model the generic extern methods like void set_rss_key<K>(in K key) above?

Normally the methods you see in P4 code should be "dataplane methods" (i.e. methods you apply in your P4 code, not via P4Runtime). @jafingerhut could you please chime in to clarify?

saynb:
@qobilidop If I understand the usage of this extern correctly, then you are right. Is there an example of a control plane API call for this Hash extern? I can see in the current spec that there is Algorithm that you can pass as the constructor over [here](https://github.com/p4lang/pna/blob/main/pna.p4#L346).
```
Hash(PNA_HashAlgorithm_t algo);
```

For the `set_rss_key()` method, let's say the P4 program looks like
```
 ToeplitzHash<0> hash_object;
hash_object.set_rss_key(value);
```
And the compiler decides to generate an object specific to the P4 function invocation, similar to what we do for digest.pack(), then, the p4runtime message (and the corresponding p4info) could look like the below
```
generic_table_entry {
  table_id: ...  // Refers to the extern object
  // No match fields in this case
  table_data_union {
    union_id: ...  // Refers to the set_rss_key method
    params {
      {
        param_id: ...  // Refers to the key param
        value: ...
      },
      {
        param_id: ...  // Refers to the algorithm param
        value: ... // String enum depicting the algorithm
      }
    }
  }
}
```
I will add a refined usecase with both p4info and p4runtime examples. If you can think of an extern with more complicated usage like having multiple key fields and data fields, then that would be great.

qobilidop:
Thanks @saynb! I don't have a more complicated usage in mind right now. I look forward to your examples to learn more.

jafingerhut:
Yes, GenericTable is intended to define control plane APIs for externs, including things like the ToeplitzHash extern mentioned above.

Yes, all methods defined for an extern object in P4 are methods intended to be called in the data plane, from your P4 program, while processing packets.  There is NO standard way in P4 to define a control plane API for extern objects INSIDE the P4 program today.  If you like the idea of being able to modify the control plane API without having to go back and update your P4 code every time (assuming the data plane API is not affected), then you might think of this as an advantage.  (I am not claiming that everyone would view that as an advantage.)

I do not recall the bit and byte level details of the ToeplitzHash control plane API that we might want here, but it is basically something like specifying N different 32-bit unsigned integers as a "secret key" input to the hash.

One possible simple way to represent this as a single GenericTable would be to have unsigned integer exact match keys in the range [0,N-1], each having a value that is a 32-bit unsigned integer.  There would be a separate such table for each extern object instance of a ToeplitzHash in the P4 program, with a name for the GenericTable instance that is probably some simple function of the extern object instance name, e.g. if the extern object instance name is "X", then the GenericTable instance name might be "X.SecretKeyConfig", where "SecretKeyConfig" is chosen at design time of the ToeplitzHash control plane API, and unlikely to ever be changed later unless you decide to completely change the control plane API of the extern in a backwards-incompatible way.

smolkaj:
> There is NO standard way in P4 to define a control plane API for extern objects INSIDE the P4 program today.

That's an interesting point. I do wonder if there should be. Maybe something to discuss in the P4 API WG.

jafingerhut:
jafingerhut> There is NO standard way in P4 to define a control plane API for extern objects INSIDE the P4 program today.

smolkaj> That's an interesting point. I do wonder if there should be. Maybe something to discuss in the P4 API WG.

jafingerhut> I agree it could be an interesting topic.  One viewpoint on this: _If_ GenericTable support is added to the P4Runtime API, then perhaps one way to specify a control plane API in a P4 program, for an extern object type, is to create a syntax for specifying one or more GenericTable instances that will be the control plane API for any extern object instance created (of that type).

That would avoid the need to specify API details for any specific programming language (e.g. C++, Java, Python, Golang, etc.) and instead it would simply be "Whatever the control plane API is for configuring these GenericTable objects, that is the control plane API for this extern."

If there were fancy restrictions on allowed values, sequence of operations, etc., that might get complicated to create a language to specify all of those, but at some point (perhaps very quickly), such restrictions would likely be just part of the documentation of the control plane API in comments.

chrispsommers:
>...create a syntax for specifying one or more GenericTable instances that will be the control plane API for any extern object instance created (of that type).
@jafingerhut Would this be a language construct, an annotation, or something else?

jafingerhut:
@chrispsommers I'm tossing ideas out from the top of my head here, not doing a careful thoughtful design process.  Just warning you that there could be holes in anything I've said on this topic in this comment, and comments above, regarding such a feature.

If one wanted to make this part of the official P4 language definition of a P4 extern object, _without_ using annotations, then making it standard would mean proposing a syntax for it and going through the language design work group.  Possible, certainly, but perhaps more time consuming than using annotations.

Doing it using annotations would not require going through the language design work group for changes (at least, I cannot think of any reason language spec changes would be needed).  One would just need to come up with a use of existing P4 annotations to define such control plane API tables.

chrispsommers:
>Doing it using annotations would not require going through the language design work group for changes (at least, I cannot think of any reason language spec changes would be needed). One would just need to come up with a use of existing P4 annotations to define such control plane API tables.

Recall that P4-16 language spec lists reserved annotations. If we define annotations for this solution discussed above, we'd likely want to add them to the spec. If we vetted a solution in this WG and got it into practice it would probably be an easier sell.

proto/p4/config/v1/p4types.proto
```diff
@@ -292,4 +292,166 @@ message P4NewTypeSpec {
...
+}
+
+// Describes Generic Data types.
+message GenericDataTypeSpec {
+  oneof type_spec {
+    GenericBitstringLikeTypeSpec bitstring = 1;
+    GenericFloatType float = 2;
+    GenericBoolType bool = 3;
+    GenericNamedType struct = 4;
+    GenericNamedType list = 5;
```
saynb, line 318: sets ?

saynb, line 318: separate unordered lists

jafingerhut, line 318: This may be a duplicate with some other comments Sayan or others may add to this PR:

These variants of collections were discussed verbally during the 2023-May-19 meeting:

+ list - ordered, duplicates allowed
+ sets - unordered, no duplicates allowed
+ bag/multiset/some-other-name - unordered, duplicates allowed

That leaves open a fourth possibility that I don't know any name by which it is commonly called:

+ name? - ordered, no duplicates allowed

Andy's personal preference: If all of the first three variants are desired, it seems better to give those different names, rather than having a single `list` type with one flag for ordered/unordered, and another flag for duplicates allowed/not-allowed.

saynb: I like lists, sets and multisets.
Maybe we should think about unions in future as well.

jafingerhut: Ah, I have heard the 4th one called an "ordered set" before, where order is significant, and no duplicates are allowed.  At least in Python and a few other high level programming languages.

proto/p4/GenericTable.md
```diff
@@ -0,0 +1,108 @@
...
+unions {
+  preamble {
+    id: 23557840
+    name: "multicast_group_member_add"
+    alias: "multicast_group_member_add"
+  }
+  params {
+    id: 1
+    name: "instance"
+    repeated: true
```
byllyfish, line 50: If you use a variable list type (GenericListTypeSpec), does that replace the "repeated" here? Is this "repeated" necessary?

proto/p4/GenericTable.md
```diff
@@ -0,0 +1,108 @@
+# GenericTable example p4info
+
+Below is an example for realizing MulticastGroupEntry as a GenericTable.
+The key (MatchFields) comprise of only the group_id. 2 different ways of
+defining the data fields has been presented. Only 1 of them is required.
+
+```
+generic_tables {
+  generic_table_type_id : 145
+  generic_table_type_name : "MulticastGroup"
```
byllyfish, line 10: Why is the "resource" type represented as both an int *and* a string? It might be useful to do a wildcard-read filtered on all generic_table entries of the same type. For that use case, have the type as an integer makes sense.

proto/p4/GenericTable.md
saynb: You are right `repeated` is redundant with List. I was thinking of having type as the below
```
    type {
      repetition_type : "list" // can be list, bag/multiset , set, unordered_set 
      type : "bytes"
      width : 32
    }
  }
```

proto/p4/GenericTable.md
saynb: I am not sure I understood the question here. Are you talking about why we need an ID and a name?

byllyfish: Yes, my question was why do we need both an ID and a name? However, I just noticed that your GenericTable/GenericTableInstance uses the same design as Extern/ExternInstance in `p4info.proto`. So, it is starting to make more sense to me. 😄

saynb: Yes. I will be including a new item called a table_category_type_name and ID as well to make it more confusing before the next meeting tomorrow. Working on the writeup

proto/p4/config/v1/p4types.proto
```diff
@@ -292,4 +292,211 @@ message P4NewTypeSpec {
...
+    GenericNamedType ordered_set = 8;
+    GenericNamedType string = 9;
+    GenericNamedType enum = 10;
+    GenericNamedType serializable_enum = 11;
+    GenericNamedType new_type = 12;
+    // P4 data type spec can help achieve parity with types being used
+    // in P4 externs. For example, if Register is using a type in P4
+    // Data type spec, then it will make it easier to map here when using
+    // GenericTables
+    P4DataTypeSpec p4_type = 13;
```
jafingerhut, line 330: If we keep this line in this definition, it would be a good idea to specify exactly when someone should use p4_type vs. bool, struct, enum, serializable_enum, or new_type above, so that among the two available choices, a consistent one is used.  Ideally the rule for which to use should be simple, e.g. "If the enum or serializable_enum is defined in the user's P4 program, use p4_type, otherwise use enum or serializable_enum instead of p4_type".

I have no idea what the rule should be to distinguish whether to use p4_type for `bool`, or use field `bool` directly above, though.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6457,6 +6492,49 @@ applicable on the table and the APIs.
...
+
+
+## GenericTable properties
+
+Properties exist at 3 different levels - Table, entry and field (match or union)
+
+### Table properties
+
+Table properties are at table level. A combination of these define a certain
+table category
```
chrispsommers, line 6516: Perhaps state clearly that each property can be `true` or `false`

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6457,6 +6492,49 @@ applicable on the table and the APIs.
...
+
+* Read-only
+* Modify-only
+* Reset-only
+* Indexed
+* Keyless
+* Calculation
```
chrispsommers, line 6523: Per the meeting, some properties are mutually-exclusive and some may overlap. Could you show a matrix or truth-table showing what's possible?

jonathan-dilorenzo: And presumably, any use of mutually exclusive properties should be rejected immediately.

chrispsommers (commented):
Misc comments from today's meeting.

proto/p4/config/v1/p4types.proto
saynb: Added a writeup

docs/v1/P4Runtime-Spec.mdk
saynb: Added the matrix

chrispsommers:
Did you want to close this and mention it's superseded by https://github.com/p4lang/p4runtime/pull/435 ?

proto/p4/v1/p4data.proto
```diff
@@ -34,6 +34,30 @@ message P4Data {
     bytes enum_value = 12;  // serializable enums only
   }
 }
+message GenericData {
+  oneof data {
+    bytes bitstring = 1;
+    GenericVarbit varbitstring = 2;
+    float float = 3;
```
byllyfish, line 41: This is a 32-bit float. If intended, it would be best to call it `float32`.  It might be better to use `float64` in a serialization spec for flexibility.

saynb:
Have opened https://github.com/p4lang/p4runtime/pull/435 to continue working on it

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+the `GenericTableEntry` entity, which has the following fields:
+
+* `table_id`, which identifies the table instance; the `table_id` is determined
+  by the P4Info message.
+
+* `match`, a repeated field of `GenericFieldMatch` messages. Each element in the
+  repeated field is used to provide a value for the corresponding element in the
+  p4info.
+
+* `union`, which indicates which of the table's unions to execute in case of
```
byllyfish, line 6404: Nit: I have a concern about the use to the word `union` by itself. I understand this is a C union type; but union is also a set operation. When I first read this spec, I wasn't sure if the GenericTableEntry's data could support a set of actions -- more than one action (like OpenFlow).

GenericTableEntry's data is in the form of a variant record. Each GenericTableEntry can execute exactly one variant. I feel that the word `variant` is an improvement on `union` in this context.

proto/p4/v1/p4data.proto
chrispsommers: Protobuf defines `float` as 32 bits and `double` as 64 bits, see https://protobuf.dev/programming-guides/encoding/#structure
We can just make up a new primitive name like `float32`, right?

Issue #426, opened 2023-05-13
Enhancements to P4Runtime API for P4 table "initial entries" proposal


jafingerhut:
2023-Aug-08 status of this issue: https://github.com/p4lang/p4runtime/pull/432 has several approvals from key people involved, and I hope that it can be merged in during or very soon after the next P4 API work group meeting on 2023-Aug-11.

As of version 1.2.4 of the P4_16 language spec, this PR was merged, enabling P4 programs to specify a list of initial entries in a table, but the runtime is still able to add new entries later during runtime, or to modify or delete any individual entries not marked `const`: https://github.com/p4lang/p4-spec/pull/1180

On 2023-Jun-07, a partial implementation of this feature has been merged into the open source p4c compiler here: https://github.com/p4lang/p4c/commit/b80b2cbc166399ba5ce8f8a31ad1f054f05cd577

I say that it is a partial implementation, because for tables with initial entries, it does not mark those tables in P4Info in any way to distinguish them as having initial entries, which is probably desirable to do in some way, and it does not distinguish in the "entries" output file between const vs. non-const entries yet.  How to do those things are part of the subject of this issue.  When those are decided upon by the P4 API work group, then this p4c issue should be resolved with appropriate changes to the P4Runtime generation code in p4c: https://github.com/p4lang/p4c/issues/4016

This proposed PR on the P4Runtime spec has several TODO items in it that should be discussed and agreed upon in the P4 API work group, in order for it to proceed forward: https://github.com/p4lang/p4runtime/pull/432.   (alternately, some other PR that addresses this issue in some other way should be created and merged in)

Today, `const entries` descriptions of table entries are written by p4c front/mid end to a file of entries.  None of those contain a priority, nor do they contain a per-entry `const` qualifier.  The 'initial entries' approach will likely need extra properties on each entry for those.

There may be interesting questions of what the various SetForwardingPipelineConfig options should do with initial entries, that never arose with `const entries`, because the answer for `const entries` was "they are always there, no matter what".  Initial entries are defined to be there when you initially load the P4 program, but what if the same program is reloaded?  Should the entries of that table be reinitialized back to what the source code says?  Or should they be left as they currently are?  Or should there be multiple options that let the controller choose?

jafingerhut:
Updated status, this PR has now been merged into the language spec, and it seems likely that version 1.2.4 of the spec will be released within a week or so from now: https://github.com/p4lang/p4-spec/pull/1180

[...]

PR #432, opened 2023-05-28
Support for initial entries

jafingerhut:
Fixes https://github.com/p4lang/p4runtime/issues/426


jafingerhut:
As of commit 2 of this PR, this is a first draft of a proposal for addressing issue https://github.com/p4lang/p4runtime/issues/426

There are still several embedded questions marked TODO at this point.  Please comment with suggested answers to those questions if you have any.

proto/p4/v1/p4runtime.proto
```diff
@@ -210,6 +210,28 @@ message TableEntry {
   IdleTimeout time_since_last_hit = 10;
   // Arbitrary metadata from the controller that is opaque to the target.
   bytes metadata = 11;
+  // const is true only for entries that cannot be deleted or
+  // modified, e.g. those in a table with `has_initial_entries = true`
+  // that have been marked `const` _on that individual entry_ in the
+  // P4 source code within its `entries` table property.
+  //
+  // TODO: Should const be true for all entries read from a table
+  // declared with `const entries`?  Would that be considered a
```
chrispsommers, line 219: Existing implementations will not populate this field, evaluating to 0=false, even for constant entries, resulting in bogus information. I don't know if this constitutes "backwards incompatibility," or it simply means existing implementations are suddently non-conformant with a new requirement. If the consensus is to adopt the proposal despite this fact, we would need at a minimum to write an advisory that this field should be ignored unless the implementation conforms with the spec version which adopts this.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3298,6 +3301,68 @@ entries. When a priority value is required (&eg; for tables including `RANGE`,
...
+queried.  Idle timeouts are not supported for such tables.
+
+If the table requires a priority value for entries, the priorities of
+the initial entries are determined according to the P4~16~ language
+specification.  After the P4 program is initially loaded, the entries
+not marked `const` can be modified at run time just as table entries
+in a normal table can.
+
+TODO: Is there already a place in this specification where the
+contents of P4Runtime "entries" files are defined?  If not, should the
```
chrispsommers, line 3367: What are P4Runtime "entries" files?

proto/p4/v1/p4runtime.proto
jafingerhut: That does seem like an undesirable state.  One way to avoid such a situation is to define that when reading entries from a table with `const entries`, or from a default entry declared with `const`, that this new TableEntry `const` flag will always be false, to ensure that old and new servers return the same result for these kinds of read requests both before and after this spec change is made.

If we did that, we would try to document this as clearly as we can, and the reason why (because it would not be immediately obvious to someone new to the P4Runtime spec a year or two from now), and that the only time the server would set the `const` flag to `true` in a response to a read request would be for a table that had a `const` flag on an entry in a table with this new style of initial entries.

docs/v1/P4Runtime-Spec.mdk
chrispsommers: Got it, thanks for the refresher. It seems like these entries are implicitly formatted as P4Runtime messages, which creates an implicit linkage to the P4RT spec, same as P4Info is generated in the front-end, which also depends on the P4RT spec. This should probably be stated clearly. I assume backend implementations normally handle this by consuming the entries and applying them at startup.

proto/p4/v1/p4runtime.proto
chrispsommers: Your logic is sound, but it doesn't feel satisfying or intuitive overall - too many conditions to wrap one's head around. It seems like a long-term solution to a short-term problem. Perhaps we should get others to weigh in on this. I'm not sure how objectionable it is - clients of this older server could simply not try to use the field until the server supports it, same as any other new feature. Perhaps we should get others to weigh in on this.

proto/p4/v1/p4runtime.proto
```diff
@@ -210,6 +210,22 @@ message TableEntry {
...
+  // backwards-compatible change in this specification?  If it is
+  // considered backwards-incompatible, is it an acceptable
+  // incompatibility?
+  //
+  // TODO: Should const be true when reading a default entry from a
+  // table, if it was declared as `const default_action` in the P4
+  // source code?  Would that be considered a backwards-compatible
+  // change in this specification?
+  bool const = 13;
```
chrispsommers, line 228: Is this field actually writable? It seems like it's an invariant based on the P4 code itself. Do we want to be able to write it back (with correct value) just for the sake of read-write symmetry? I'm trying to comprehend the new para. around line 3046 in the .mdk spec itself:

>The `const` field must be `false` in any `INSERT`, `MODIFY`, or
`DELETE` write request of a table entry.  If it is true, the server
must reject the operation and return an `INVALID_ARGUMENT` error.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3298,6 +3305,68 @@ entries. When a priority value is required (&eg; for tables including `RANGE`,
 `TERNARY` or `OPTIONAL` matches), it is inferred based on the order in which
 entries appear in the table declaration.
 
+TODO: With the recent change to the P4~16~ specification for initial
```
chrispsommers, line 3308: Any further insights into this concern?

docs/v1/P4Runtime-Spec.mdk
chrispsommers: I looked through the spec for likely locations to mention the `<program-name.p4.entries.txt` file emitted by the compiler front end. I looked at https://p4.org/p4-spec/docs/p4runtime-spec-working-draft-html-version.html#sec-table-entry section but this would imply initial entries are only for tables, which they are today, but perhaps this could change. So how about a new appendix called "Initial Entries?"

chrispsommers (commented):
More comments.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: A new appendix sounds good.  I can create one as part of this PR, if that seems like a reasonable way to add such a section.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Here is the proposed PR for the language spec that would make it explicit that it is a compile-time error to try to specify priorities in the source code within `const entries`, matching the current open source p4c behavior: https://github.com/p4lang/p4-spec/pull/1259/files

docs/v1/P4Runtime-Spec.mdk
jafingerhut: With commit 7 of this PR I have added a new appendix at the end that documents the contents of entries files generated by p4c.

proto/p4/v1/p4runtime.proto
jafingerhut: I would love to get others to weigh in on this.  @smolkaj @jonathan-dilorenzo @thantry Can you or someone you know at Google who is interested in getting initial entries actually working in an implementation (as opposed to merely specified in P4.org documents) please take a look, or get someone else to take a look, at this PR in detail and comment?

docs/v1/P4Runtime-Spec.mdk
jafingerhut: With commit 11 of this PR, I have removed the paragraph marked TODO, and replaced it with a footnote linking to this PR on the language spec, which is not accepted and merged in yet, but I'm being optimistic about that: https://github.com/p4lang/p4-spec/pull/1259

proto/p4/v1/p4runtime.proto
jafingerhut: I would still like others to weigh in on this question, but thinking about this some more, it does seem in the spirit of Protobuf based messages to consider it still 'backwards compatible' if an old message has a new field with additional information added to it.  If we have a new `const` field in `TableEntry` messages, and new P4Runtime servers in the future return `true` for this field for entries declared with `const entries` in the source code, even though older P4Runtime servers return no such field at all, that seems like something that P4Runtime API clients should just need to handle gracefully.

jafingerhut: With commit 12 on this PR, I have changed the proposed comment being commented on here to say that this new `const` field for TableEntry messages WILL be set to true by the server for all entries read from a table with `const entries`, and also for reading the default action from a table that has `const default_action`.  We can change that before approving, of course, depending on whether anyone else has arguments against this choice.

jafingerhut:
Jonathan DiLorenzo: Should we have tables with `const entries` have `has_initial_entries` AND `is_const_table` both true?

Andy: I will make this change in another commit on this PR and let people know it is ready for review.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6598,4 +6666,53 @@ clients are aware that the server is using a larger maximum receive message
...
+format is a binary P4Runtime API protobuf message format.  The `.txt`
+format is the text encoding of the same Protobuf messages.
+
+Target devices are *not* required to use this file.  For example, if a
+target has a P4 compiler back end that encodes all of the necessary
+details from the P4 source program, including the `entries` of tables,
+in a target-specific binary format, then that target might hae no
+reason to generate these entries files.
+
+Some target devices might choose to generates entries files, and also
```
chrispsommers, line 6697: /generates/generate

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6598,4 +6666,53 @@ clients are aware that the server is using a larger maximum receive message
...
+
+You can replace the `.txt` suffix of the file name `tmp/entries.txt`
+in the example command above with `.json` or `.bin`.  The `.bin`
+format is a binary P4Runtime API protobuf message format.  The `.txt`
+format is the text encoding of the same Protobuf messages.
+
+Target devices are *not* required to use this file.  For example, if a
+target has a P4 compiler back end that encodes all of the necessary
+details from the P4 source program, including the `entries` of tables,
+in a target-specific binary format, then that target might hae no
```
jonathan-dilorenzo, line 6694: hae -> have*

chrispsommers: That's the Scottish version. ;)

jafingerhut:
I have updated this PR to follow the suggestion made during the 2023-Jul-14 API work group meeting, so that tables declared in source code with `const entries` continue to have `is_const_table` true as before, but they also have the new flag `has_initial_entries` true.

The new style of tables that the language spec now supports that can have `entries` without `const` before `entries`, will only have `has_initial_entries` true, but `is_const_table` will be false.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6598,4 +6669,53 @@ clients are aware that the server is using a larger maximum receive message
...
+    cd p4c/testdata/p4_16_samples
+    mkdir tmp
+    p4test --arch v1model \
+        --p4runtime-files tmp/p4info.txt \
+        --p4runtime-entries-files tmp/entries.txt \
+        table-entries-ternary-bmv2.p4
+
+You can replace the `.txt` suffix of the file name `tmp/entries.txt`
+in the example command above with `.json` or `.bin`.  The `.bin`
+format is a binary P4Runtime API protobuf message format.  The `.txt`
```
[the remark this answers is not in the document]
smolkaj: I opened https://github.com/p4lang/p4c/issues/4066 to track this

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6598,4 +6669,53 @@ clients are aware that the server is using a larger maximum receive message
...
+Some target devices might choose to generate entries files, and also
+to require doing so in order to have a correct implementation.  For
+example, a target runtime implementation might take a target-specific
+binary format for the compiled P4 program that does *not* contain any
+data describing the `entries` of tables, plus the entries file
+generated by p4c, and use the entries file to load the initial entries
+of tables into the device.
+
+The format of the entries file is a sequence of 0 or more `Update`
+messages.  All of them have `type` equal to `INSERT`, and `entity` is
```
smolkaj, line 6709: I suspect just a `WriteRequest`?
The protobuf parsers don't really support sequences of messages out of the box AFAIK, so if it was really just a concatenation of parsable protobuf blobs that would make things pretty hard to consume.

proto/p4/config/v1/p4info.proto
```diff
@@ -217,8 +217,16 @@ message Table {
   }
   // is idle timeout supported for this table?
   IdleTimeoutBehavior idle_timeout_behavior = 9;
-  // table with static P4 entries, cannot be modified at runtime
+  // is_const_table is true if and only if the table's entries are
```
smolkaj, line 220: Optional: Perhaps just start the comment with "True if and only if ..." since the subject is implied? But that's just my personal preference and you may prefer it this way.

Ditto for the other new fields.

proto/p4/v1/p4runtime.proto
```diff
@@ -210,6 +210,18 @@ message TableEntry {
...
+  // + Any entry read from a table declared with `const entries`
+  // + The default entry read from a table declared with `const default_action`
+  // + Any entry declared with `entries` without the `const` qualifier
+  //   before `entries`, and thus the table has `has_initial_entries =
+  //   true`, where the individual entry has the `const` qualifier on
+  //   it in the P4 source code.
+  // Note: Older P4Runtime API servers before the `const` field was
+  // added to this message will not return a value for `const` in the
+  // first two cases above, but newer P4Runtime API servers will.
+  bool const = 13;
```
smolkaj, line 224: Nit: Any thoughts on `const` vs `is_const`? Googlers would probably generally prefer the latter, and I also have a slight preference that way.

proto/p4/v1/p4runtime.proto
```diff
@@ -210,6 +210,18 @@ message TableEntry {
   IdleTimeout time_since_last_hit = 10;
   // Arbitrary metadata from the controller that is opaque to the target.
   bytes metadata = 11;
+  // const is true only for entries that cannot be deleted or
+  // modified, e.g. any of the following:
+  // + Any entry read from a table declared with `const entries`
+  // + The default entry read from a table declared with `const default_action`
```
smolkaj, line 216: I didn't realize that default actions generate entries? Is that true for all targets and standardized somewhere, or a target-specific thing?

smolkaj (changes requested):
This looks fantastic! I'm very excited about this new feature. Thank you Andy!

I left a few minor comments, mostly nitpicking.

proto/p4/v1/p4runtime.proto
jafingerhut: Is it perhaps clearer if I said "The default action read from a table" instead of "The default entry read from a table"?

The actual hardware resource utilization of a default action in a P4 table is potentially a little target-specific, e.g. for a ternary table, I can imagine implementations that use a TCAM entry that is completely wildcarded with lowest match priority to implement a default_action, whereas another implementation might have a separate place to store a default action outside of the TCAM completely.  Typically it only makes a difference of one hardware entry used vs. not in implementations I am familiar with.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3293,10 +3306,68 @@ direct resources that are being queried.  Idle timeouts are not supported for
...
+
+For every P4 table, it will fall into one of these three categories:
+
+* Neither `entries` nor `const entries` are declared in the source
+  code, and thus `is_const_table` and `has_initial_entries` will both
+  be false.  This will be called a "normal table" in this section
+  below.
+* The table has `const entries` declared, and thus a separate
+  `entries` property is not permitted by the language.  Such a table
+  will have `is_const_table` true and `has_initial_entries` true.
```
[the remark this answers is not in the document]
jafingerhut: P4 supports declaring a table with `const entries`, or with `entries` (without `const` before it), with an empty list of entries.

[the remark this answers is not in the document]
smolkaj: So technically, we might have a 4th category of tables with `is_const_table = true`  and `has_initial_entries = false`?

proto/p4/v1/p4runtime.proto
jafingerhut: Every P4 table always has a default action.  If it isn't explicitly specified in P4 source code, it defaults to NoAction

smolkaj: But at least on our switches, there is no table entry that I can read back via P4Runtime that implements the default action.

docs/v1/P4Runtime-Spec.mdk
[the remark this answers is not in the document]
jafingerhut: Elsewhere in these changes the definition of `has_initial_entries` true is that the P4 source code has the `entries` property present for the table, vs. not present for the table.  Do you think that is a reasonable definition?

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Here is an example entries file written by p4c in text format: https://github.com/p4lang/p4c/blob/main/testdata/p4_16_samples_outputs/table-entries-ternary-bmv2.p4.entries.txt

It doesn't look like a `WriteRequest` message to me, since there are no fields of a WriteRequest message present there, only a sequence of 0 or more `Update` messages.  If you tell me that is a valid `WriteRequest` message, I am happy to document it that way in the P4Runtime spec.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3293,10 +3306,68 @@ direct resources that are being queried.  Idle timeouts are not supported for
...
+&ie; only `MODIFY` operations on direct resources are allowed.
+
+Unlike a table with `is_const_table = true`, a client may insert
+entries into a table with `has_initial_entries = true` and
+`is_const_table = false`, subject to capacity constraints on the
+number of entries supported by the target for the table.
+
+The contents of tables with initial entries can be queried by the
+client through a `ReadRequest`.  When reading entries from a table
+with initial entris, the following fields must be set by the server:
```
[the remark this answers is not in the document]
jafingerhut: Makes sense.  My next commit should have changes that avoid this redundancy.

docs/v1/P4Runtime-Spec.mdk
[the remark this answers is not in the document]
jafingerhut: Maybe to be more explicit, are you saying that you think `has_initial_entries` should be true if and only if there is an `entries` property for the table in the P4 source code, and there is at least one entry listed there?  So if the list contains 0 entries, `has_initial_entries` should be false?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3293,10 +3306,68 @@ direct resources that are being queried.  Idle timeouts are not supported for
...
+required).  This is in addition to any direct resources that are being
+queried.  Idle timeouts are not supported for such tables.
+
+If the table requires a priority value for entries, the priorities of
+the initial entries are determined according to the P4~16~ language
+specification.  After the P4 program is initially loaded, the entries
+not marked `const` can be modified at run time just as table entries
+in a normal table can.
+
+The contents of all table entries within `entries` table properties in
```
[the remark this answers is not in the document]
jafingerhut: I am changing "within `entries`" to "within the `entries`'"

proto/p4/v1/p4runtime.proto
jafingerhut: I don't have a strong preference either way, so fine with changing it to `is_const`

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3293,10 +3306,68 @@ direct resources that are being queried.  Idle timeouts are not supported for
...
+client through a `ReadRequest`.  When reading entries from a table
+with initial entris, the following fields must be set by the server:
+`table_id`, `match`, `action`, `is_default_action`, and `priority` (if
+required).  This is in addition to any direct resources that are being
+queried.  Idle timeouts are not supported for such tables.
+
+If the table requires a priority value for entries, the priorities of
+the initial entries are determined according to the P4~16~ language
+specification.  After the P4 program is initially loaded, the entries
+not marked `const` can be modified at run time just as table entries
```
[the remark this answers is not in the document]
jafingerhut: I believe commit 16 removes this redundancy.

proto/p4/config/v1/p4info.proto
jafingerhut: I believe this comment is addressed in or before commit 17

proto/p4/v1/p4runtime.proto
jafingerhut: P4Runtime API read requests can read either one (or all) "hit entries", or the default action of a table, never both in the same read request.  See mentions of the field `is_default_action` throughout the P4Runtime spec.

What happens if you send a ReadRequest to one of your switches with `is_default_action` true?

docs/v1/P4Runtime-Spec.mdk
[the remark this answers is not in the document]
jafingerhut: @smolkaj I thought about this some more since my last comment, and came to the conclusion that it does seem to make more sense if `has_initial_entries` is true if and only if the P4 source code has an `entries` table property, _and_ the list of entries is not empty.  With commit 19 on this PR, I believe I have made all changes required to consistently define it this way.

This does mean that a normal table is either:
(a) one that has neither `entries` nor `const entries` properties in the source code
(b) one that has `entries = { }`, with no `const` before `entries`

Two tables declared as (a) or (b) would both have `is_const_table` false and `has_initial_entries` false as proposed by commit 19 version of this PR.  That makes a kind of sense to me, since the things that the client can do to either a table declared like (a) or (b) would be exactly the same, i.e. there are no entries initially loaded into the table when you load the P4 program, and the runtime can insert, modify, or delete any entries it wants to at run time.

jafingerhut:
@smolkaj @jonathan-dilorenzo @chrispsommers I have made 7 commits to this PR since the 2023-Jul-14 API work group meeting, primarily due to comments made during the meeting, and from Steffen's careful review after the meeting, and I think it is the better for it.

I have no current plans to change it any more, until and unless there is additional review feedback that warrants additional changes.

proto/p4/v1/p4runtime.proto
jafingerhut: I may not have been clear above, but note that what P4Runtime API defines is that there are 0 or more entries in a table, which if any of them are matched the device must behave as if table_name.apply().hit is true, and if none of them match, the device must behave as if table_name.apply().hit is false.  In the latter case, the default_action of the table must be executed.

As an implementation detail, one correct way to implement that behavior for a hardware TCAM is to have a separate hardware TCAM hit entry with wildcard value/mask, and an action equal to the currently configured default_action.  A correct P4Runtime API implementation for such a hardware implementation should _never_ return the hardware entry representing the default_action from a P4Runtime API read request that is reading entries installed in the table, because that hardware entry is an implementation detail for how that device chose to implement default_action behavior.

proto/p4/v1/p4runtime.proto
smolkaj: Sorry for missing this until now.

I am not someone who is planning to immediately use initial entries at this point, but can answer from a protobuf / semantic versioning perspective.

This requires a MINOR version bump in terms of https://semver.org/.

This is a "backward compatible API change" in the sense that an "old client" (using the old .proto) will continue to work without problems with a "new server" who adopt the new API:
* READ: When an old client gets a message with the new field (in binary format), protobuf will simply ignore the extra data and the client will not see it. 
* WRITE: When an old client writes a message without the new field (in binary format), a new server will see the field as being set to `false`. That works out fine for this proposal.

Note that a new client won't automagically work with an old server, as mentioned by other above: they will read back all entries with `is_const: false` even if the entries are const. But since the client made a deliberate choice here to adopt the new API, it's also their job to ensure they can support older servers as well, if they wish. ("This is a backward compatible API change" means that upgrading servers to the new API won't break clients -- it doesn't talk about what happens if clients start using the new API with servers that don't support it...)

@jafingerhut Perhaps we should leave a comment saying that the field will be set by P4Runtime servers starting at version x.y.z?

docs/v1/P4Runtime-Spec.mdk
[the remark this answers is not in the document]
smolkaj: > Elsewhere in these changes the definition of has_initial_entries true is that the P4 source code has the entries property present for the table, vs. not present for the table. Do you think that is a reasonable definition?

I hadn't thought about that, and actually do think this is also a reasonable definition (even if the `entries` property specifies the empty set of entries, which is still a set of entries).

> are you saying that you think has_initial_entries should be true if and only if there is an entries property for the table in the P4 source code, and there is at least one entry listed there? So if the list contains 0 entries, has_initial_entries should be false?

I was just asking a question -- sorry if it sounded like I had thought this through to the end. 

> I thought about this some more since my last comment, and came to the conclusion that it does seem to make more sense if has_initial_entries is true if and only if the P4 source code has an entries table property, and the list of entries is not empty.

This also SGTM. Just to check my understanding, the main difference is that we now have "constant tables" with and without `has_initial_entries`.

Actually, it does seem nice to allow a constant table with `has_initial_entries: false`, since clearly this is easier to support by the switch than if `has_initial_entries: true`.

> (b) one that has entries = { }, with no const before entries

Orthogonal to this PR, but perhaps we should ask p4c to disallow this? One weird corner case less?

docs/v1/P4Runtime-Spec.mdk
smolkaj: Yup, the link you provided does show a valid `WriteRequest`.

> It doesn't look like a WriteRequest message to me, since there are no fields of a WriteRequest message present there

Actually it does: the `repeated updates` field is a `WriteRequest` field [[1]], and the textproto you linked to contains a bunch of `updates { ... }`.

This is admittedly subtle, because the text proto format doesn't include an outer wrapper for the message that is being encoded. Instead, it just shows the fields of the message at the top-level.

That being said, the text format is ambiguous, this could be a dump of *any* proto message with a `repeated Update` field. And for the binary format, this distinction matters. So we shouldn't blindly claim this is a `WriteRequest`. Do you happen to know where in the p4c source code this output gets created? From there it will be straightforward to tell the correct message type.

[1]: http://google3/third_party/p4lang_p4runtime/proto/p4/v1/p4runtime.proto;l=73;rcl=534134847

proto/p4/v1/p4runtime.proto
smolkaj: Turns out I had some catching up to do on "default entries" and `is_default_action`. All makes sense now, thanks, and sorry if I didn't make sense.

> What happens if you send a ReadRequest to one of your switches with is_default_action true?

Our switches currently only support wild card reads. And I just double checked that wildcard reads do not include the default entry, which is probably why I wasn't super aware they were a thing thus far. So all is good.

smolkaj:
This LGTM! There seem to be 3 remaining threads that would be nice to resolve, 2 of them being nits. Other than that, from my side this is ready to be merged.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Here is the p4c code that writes the entries file: https://github.com/p4lang/p4c/blob/main/control-plane/p4RuntimeSerializer.cpp#L973-L976

in case the line numbers go askew, the method name is `P4RuntimeEntriesConverter`

proto/p4/v1/p4runtime.proto
jafingerhut: Version x.y.z of what project?  Stratum?  p4lang/PI?  The P4Runtime API specification?

docs/v1/P4Runtime-Spec.mdk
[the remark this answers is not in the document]
jafingerhut: > This also SGTM. Just to check my understanding, the main difference is that we now have "constant tables" with and without `has_initial_entries`.

Yes, and the text I have in this PR now mentions both of these possibilities explicitly, and calls both variants "constant tables".

> Orthogonal to this PR, but perhaps we should ask p4c to disallow this? One weird corner case less?

I suppose we could, but it doesn't seem that bad to me if `entries = { }` has the same meaning as a table with no `entries` table property declared in the source code at all.

proto/p4/v1/p4runtime.proto
jafingerhut: Note: I have changed it to `is_const` last Friday, so resolving this comment.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: If you would like to examing the binary and/or JSON format of the entries file generated by current open source p4c, in addition to the text format, they are all in the zip file attached to this comment.
[tmp.zip](https://github.com/p4lang/p4runtime/files/12086082/tmp.zip)

smolkaj: I didn't find the place where things get serialized, but I would consider the following a good-enough hint that the message is a `p4.v1.WriteRequest`: https://github.com/p4lang/p4c/blob/main/control-plane/p4RuntimeSerializer.cpp#L1327-L1328

I also opened https://github.com/p4lang/p4c/issues/4070 so people don't have to guess in the future.

jafingerhut: Is there an easy way for you to parse the contents of the binary format entries file in the file `tmp.zip` attached in my previous comment, to determine whether it is a valid binary `WriteRequest` protobuf message?  I suspect it probably is, but do not know any quick way to check.

proto/p4/v1/p4runtime.proto
jafingerhut: I added two "Added in 1.4.0" comments with my latest commit to this PR, one for `is_const` in `TableEntry`, and the other for `has_initial_entries` in `Table`.

docs/v1/P4Runtime-Spec.mdk
smolkaj: Confirmed with a Google-internal CLI tool that entries.bin can be parsed as a `p4.v1.WriteRequest`.

smolkaj (approved):
LGTM modulo the remaining open comment in the appendix regarding the protobuf schema.

jafingerhut:
> LGTM modulo the remaining open comment in the appendix regarding the protobuf schema.

With commit 22 on this PR, I have attempted to clarify the description of the protobuf schema of an entries file.

smolkaj:
Looks great to me.

jafingerhut:
@saynb @jamescchoi  This has been through careful review by several others already, so hopefully should also meet with your approval, but in case you want to take a look, this is something we are going to be asked to implement by some folks, so good to be aware of the details.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3293,10 +3309,73 @@ direct resources that are being queried.  Idle timeouts are not supported for
...
+&ie; only `MODIFY` operations on direct resources are allowed.
+
+Unlike a table with `is_const_table = true`, a client may insert
+entries into a table with `has_initial_entries = true` and
+`is_const_table = false`, subject to capacity constraints on the
+number of entries supported by the target for the table.
+
+The contents of preinitialized tables can be queried by the client
+through a `ReadRequest`.  The server fills in the same fields in the
+response as it does for csontant tables, as described in Section
```
antoninbas, line 3365: typo: s/csontant/constant

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3043,6 +3049,10 @@ limitation. It is recommended that, for the sake of portability, P4Runtime
 clients do not try to insert additional entries once the size indicated in
 P4Info has been reached.
 
+The `is_const` field must be `false` in any `INSERT`, `MODIFY`, or
```
antoninbas, line 3052: There is a comprehensive list of fields for the `TableEntry` message earlier in this section (or at least it should be a comprehensive list...). I have noticed that `is_const` is not listed there.

Look for:
> P4Runtime supports inserting, modifying, deleting and reading table entries with
the `TableEntry` entity, which has the following fields:

docs/v1/P4Runtime-Spec.mdk
jafingerhut: fixed

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Added is_const field to that list of fields.

antoninbas (approved):

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3293,10 +3312,73 @@ direct resources that are being queried.  Idle timeouts are not supported for
...
+`is_const_table` is false.  For tables declared with `const entries`,
+see Section [#sec-constant-tables].
+
+Every P4 table falls into one of three categories:
+
+* *Normal table*: Neither `entries` nor `const entries` are declared
+  in the source code, and thus `is_const_table` and
+  `has_initial_entries` will both be false.  A corner case is that if
+  it has `entries = { }` with no `const` before `entries`, &ie; an
+  empty list of entries, that is also a normal table.
```
chrispsommers, line 3341: It occurs to me that handling this corner-case gracefully may be useful for cases where source code is dynamically-generated based on e.g. macros or other techniques. In such cases, certain "build profiles" may result in an empty `entries` list.

chrispsommers (approved):

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6598,4 +6680,53 @@ clients are aware that the server is using a larger maximum receive message
...
+The format of the entries file is a single `WriteRequest` message
+containing one `Update` sub-message per entry in the P4 source program
+defined via an `entries` table property.  All `Update` sub-messages
+have `type` equal to `INSERT`, and `entity` is a `TableEntry` message
+containing the data for one table entry.
+
+Note that for a table with `const entries`, or with `entries` having
+one or more individual entries declared `const`, it would be an error
+for a P4Runtime client to send a `WriteRequest` to a P4Runtime server
+with the contents of the entries file, because the server should
```
jamescchoi, line 6728: It may be good to further clarify that error applies to the entries with const only, not the whole contents of the entries file.

jamescchoi (approved):
Looks to be well thought out and pretty complete.
The comment is optional.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: With the latest commit on this PR as of writing this comment, I have replaced the last paragraph of the new appendix with the following, in an attempt to clarify:
```
Note that if a P4Runtime client attempted to send a `WriteRequest` to
a P4Runtime server with the contents of the entries file, the server
must return an error for each entry that has `is_const` true, as
described in Section [#sec-table-entry].
```

jafingerhut:
@smolkaj @chrispsommers Happy to discuss this at the next P4 API work group meeting on 2023-Aug-11, in hopes of seeing if it has the right approvals for merging.

smolkaj:
We have 4 approvals, so I think we are good to go. 
Let's wait until the meeting before we hit merge, just in case.

PR #435, opened 2023-06-16
Generic Table First draft

saynb:
* Phase 1: 
Objective : Phase 1 aims to provide support for a set of simple target-specific externs through GenericTables. Current APIs should work on the tables as intended. Read-write symmetry should be maintained wherever possible. Have a structure in place in p4info with which new type system and properties at different granularities can be specified and maintained.

  * Not extend any of the current RPC updates. Use INSERT, MODIFY, DELETE and Read RPC as is
  * Define Table categories/ properties
  * Define Entry and Field properties
  * Define Type system
  * Achieve Write/Read for simple PNA specific externs like ExactValueMatchLookupTable
  * Work on generic-table-dev
* Phase 2
Extend API list and other features.

  * Entry Handles
  * Table attributes (symmetric vs asymmetric)
  * New APIs
    * reset
    * add_or_mod
    * sync
    * usage_get
    * mod_inc
  * Table custom operations extensions. A way to define new operations via json modeling. For example, sync
* Phase 3
  * String representation of fields. Like IPv4 IPv6
  * Table notifications. Model notifications in a generic way


docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+
+* `other_properties`, an `Any` Protobuf message [@ProtoAny] to embed
+  architecture-specific or target-specific table properties [@P4TableProperties]
+  that the target wants to convey.
+
+## `GenericData` { #sec-p4data-generic-data}
+
+`GenericData` is an addition to p4data.proto- which aims at adding support for
+data types which aren't covered by P4Data. This incorporates all the data types
+used as primitive data types like bytes, float, strings. It contains P
```
chrispsommers, line 6350: Dangling phrase? "It contains P"

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
 
+# GenericTable { #sec-generic-table}
+
+## GenericTable p4info { #sec-p4info-generic-table}
+
+`GenericTable` messages are used to map non-PSA P4-externs or non-P4 target-
+specific "fixed" features to their implementation in a generic way. The same can be achieved
+via `Extern`, but GenericTable provides a structured way in which every feature is
+represented as a set of match-fields and data-fields. The data consists of unions
+wihch are similar to actions. One use of GenericTable in a server backend is
```
chrispsommers, line 6290: s/wihch/which/

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+with TDI [Table Driven Interface](https://github.com/p4lang/tdi), but targets
+can use Generic table to map to their own specific feature implementations as
+well.
+It defines the following fields
+
+* `generic_table_type_id`, a 8-bit unsigned integer which uniquely identifies the
+  generic_table_type in the context of the architecture.
+  This value is in the range of `[0x00, 0xff]`. 
+  The ID in the preamble is created as `0x18` `generic_table_type_id` `16 bits of
+  genereted ID`.
```
chrispsommers, line 6300: s/genereted/generated/

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  that the target wants to convey.
+
+## `GenericData` { #sec-p4data-generic-data}
+
+`GenericData` is an addition to p4data.proto- which aims at adding support for
+data types which aren't covered by P4Data. This incorporates all the data types
+used as primitive data types like bytes, float, strings. It contains P
+
+  * bitstring : Bytes type used for simple unsigned integers or bytes
+  * float : Floating type values
```
chrispsommers, line 6353: Many questions about float came up in meetings and review of former PR417, e.g. https://github.com/p4lang/p4runtime/pull/419#discussion_r1232429264

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  * bool : Boolean type values
+  * generic_struct : Used to define a structure that can contain
+    one or more of the other members as named. One structure can contain
+    different members of different types in GenericData.
+  * generic_bag : Unordered collection of members of same type. Duplicates
```
chrispsommers, line 6358: Are there any real-world examples where this is needed? We might want to hold off on types which are "academic" for not to reduce the task of implementing/testing implementations.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+
+GenericTableEntry can be used to program non-PSA externs or non-P4 target-specific
+"fixed" features to their implementation in a generic way. This provides an alternative
+to `Extern` in a structured way and within p4runtime guidelines and semantics.
+The idea has been borrowed from TDI [Table Driven Interface](https://github.com/p4lang/tdi),
+that every state can be representated as a table or multiple tables. A table here is any
+data structure with one or more entries and each entry can be represented as a set of
+key fields and data fields.
+Actions in P4 tables behave like unions and that has been borrowed here in GenericTable.
+This can be potentially used targets to map to either TDI based backends, or even non-TDI
```
chrispsommers, line 6389: ...used by targets

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+P4Info has been reached.
+
+## GenericTable Categories
+
+All tables fall into certain "categories" which define what APIs are
+applicable on the table and the APIs.
+
+  * Regular tables: Read, INSERT, MODIFY, DELETE work as intended. Entries in
+    a table can be inserted, modified and deleted.
+    * `INSERT` : The union field must be set for every `INSERT` update. If
```
chrispsommers, line 6459: It could be beneficial to allow for omitting one or all union fields and let the target provide default value(s). This could be specified by an annotation. I'm thinking about how some DASH P4 model code tables are mapped to SAI tables, where action params are mapped to SAI attributes, Some of those attributes have defaults or can be omitted and it's hard to map the P4Runtime INSERT operation to a variable-attribute SAI call.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+      If `default_entry` is true, then the default entry is reset to initial
+      values. Const union rules apply.
+      Wildcard doesn't apply to default_entry.
+    * `READ` : The Read RPC. All entries can be read. If no entities are
+      provided in request, then all entries are read, AKA wildcard read.
+      If `default_entry` is true, then the default entry is returned. Match
+      fields should not be set in this case.
+      Wildcard read does NOT return default entry.
+  
+  * Read-only tables: Entries can only be read. Write RPC doesn't work.
```
chrispsommers, line 6509: Which error is returned if illegal operations are attempted?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,314 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+      fields should not be set in this case.
+      Wildcard read does NOT return default entry.
+  
+  * Calculation tables: Entries can only be read. However different from
+    Read-only in the sense that there is no defined size of the table since
+    the possible range of matchfields can be immense. For example, hash
+    calculation tables where matchfields are the field values and the union
+    data contains the hash value. There is no default_entry for such tables.
+    * `READ` : The Read RPC. Wildcard read doesn't exist for this table.
+      Match fields should always be set
```
chrispsommers, line 6541: Which error is returned if match fields are missing?
Which error is returned if an illegal operation is attempted?

chrispsommers (commented):
Miscellaneous comments and small fixes.

docs/v1/P4Runtime-Spec.mdk
saynb: Added a bit of text regarding expectation that HW can returned ceiled lower precision

docs/v1/P4Runtime-Spec.mdk
saynb: Added. Please check

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6347,10 +6347,13 @@ It defines the following fields
...
 data types which aren't covered by P4Data. This incorporates all the data types
-used as primitive data types like bytes, float, strings. It contains P
+used as primitive data types like bytes, float, strings. It contains p4type as
+well. P4type is to be used for any data type present in P4. If there is a
+conflict, like bool, then the P4 data type should be used.
 
   * bitstring : Bytes type used for simple unsigned integers or bytes
-  * float : Floating type values
+  * float : Floating type values. Target might return a lower-precision
+    ceiled value even if field is non-volatile.
```
chrispsommers, line 6356: How do we reconcile R/W symmetry?

saynb: How about a precision length field in p4info? The controller will know about length and will have expectations set as to what will the reading value be

chrispsommers: This probably requires some group discussion, this is s tricky issue. Precision length might not capture all the details (round-up/round-down, etc.)

docs/v1/P4Runtime-Spec.mdk
saynb: removed

docs/v1/P4Runtime-Spec.mdk
saynb: Added a flow chart diagram. please check

chrispsommers: > Added a flow chart diagram. please check
Thanks.
Is the .png editable (for posterity)?

proto/p4/GenericTable.md
```diff
@@ -8,6 +8,23 @@ defining the data fields has been presented. Only 1 of them is required.
 generic_tables {
   generic_table_type_id : 145
   generic_table_type_name : "MulticastGroup"
+  generic_table_category : "Regular-In"
```
chrispsommers, line 11: Remove? Want one source of truth, the properties.

chrispsommers (changes requested):
See comments

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,356 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+specific "fixed" features to their implementation in a generic way. The same can be achieved
+via `Extern`, but GenericTable provides a structured way in which every feature is
+represented as a set of match-fields and data-fields. The data consists of unions
+which are similar to actions. One use of GenericTable in a server backend is
+with TDI [Table Driven Interface](https://github.com/p4lang/tdi), but targets
+can use Generic table to map to their own specific feature implementations as
+well.
+It defines the following fields
+
+* `generic_table_type_id`, a 8-bit unsigned integer which uniquely identifies the
```
jafingerhut, line 6296: a -> an

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,356 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  generic_table_type in the context of the architecture.
+  This value is in the range of `[0x00, 0xff]`. 
+  The ID in the preamble is created as `0x18` `generic_table_type_id` `16 bits of
+  generated ID`.
```
jafingerhut, line 6300: I don't understand the sentence above.  What are the 3 quoted things `0x18` `generic_table_type_id` `16 bits of generated ID` ?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,356 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+via `Extern`, but GenericTable provides a structured way in which every feature is
+represented as a set of match-fields and data-fields. The data consists of unions
+which are similar to actions. One use of GenericTable in a server backend is
+with TDI [Table Driven Interface](https://github.com/p4lang/tdi), but targets
+can use Generic table to map to their own specific feature implementations as
+well.
+It defines the following fields
+
+* `generic_table_type_id`, a 8-bit unsigned integer which uniquely identifies the
+  generic_table_type in the context of the architecture.
```
jafingerhut, line 6297: I believe this is the first time that generic_table_type is mentioned in the spec.  It seems like a good place to either explain this new concept with an example, or to link to a place where such an explanation is given elsewhere in the spec.

For example, when would an architecture use the same generic_table_type for two or more different objects?  When would it use different generic_table_type's?  Why?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,356 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  message and a single architecture.
+
+* `generic_table_type_name`, which specifies the fully-qualified P4 name of the
+  generic_table in case of P4 entities. In cases of non-P4 entities, it is a
+  name which uniquely identifies this entity in the entire space including P4
+  objects
+
+* `generic_table_category_type`, please check section [GenericTable categories](#sec-generic-table-categories) for more details
+
+
```
jafingerhut, line 6313: Is it the plan to have a separate GenericTable and GenericTableInstance message?  If so, as the current P4Info proto file definition in this PR says, then this documentation should say that, and separate the list of fields above as fields of GenericTable, from the list of fields below as fields of GenericTableInstance.

Either that, or leave this as is and get rid of the idea of grouping GenericTableInstance's with the same type into one GenericTable message, and just have every GenericTable describe one instance, and get rid of message GenericTableInstance.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,356 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+* `preamble`, a `Preamble` message with the ID, name, and alias of this 
+  GenericTable.
+
+* `match_fields`, a repeated field of type `MatchField` representing the data to
```
jafingerhut, line 6317: MatchField -> GenericMatchField (assuming the proto definition is correct)

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6269,6 +6279,356 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  be used to construct the lookup key matched in this table. For information check
+  the match_fields info in the [Table section](#sec-table)
```
jafingerhut, line 6319: Also, if GenericMatchField is correct, is this the correct cross-reference?

docs/v1/P4Runtime-Spec.mdk
saynb: replacing with svg

docs/v1/P4Runtime-Spec.mdk
saynb: It took me a while to understand what I had meant by that too. Replacing it with
`0x18 + generic_table_type_id + 16 bits of generated ID by compiler`

docs/v1/P4Runtime-Spec.mdk
saynb: Added some more text. Please check

docs/v1/P4Runtime-Spec.mdk
saynb: Done. Added text in the spec

docs/v1/P4Runtime-Spec.mdk
saynb: corrected the reference

saynb:
@chrispsommers  @jafingerhut  Have taken care of the comments and marked all items as resolved. Please feel free to unresolve any comment where you feel needs any more clarification/work

chrispsommers (approved):

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3010,6 +3018,14 @@ the `TableEntry` entity, which has the following fields:
   implement idle-timeout support for the table, if applicable. See
   [Idle-timeout](#sec-idle-timeout) section for more information.
 
+* `resources`, a `repeated GenericTableEntry` to contain a list of
+  target-specific direct resources. For example, if a target supports special
+  direct resources like x-meter and y-counter, then the p4info will contain
+  GenericTable representations of the same and the list here can contain one
+  member for each of the direct resources. If the generic match fields are set
+  in these entries, it is an error. The IDs will be mentioned in the
```
jafingerhut, line 3026: Most other places in the spec I believe say something like "If any match fields are set in these `GenericTableEntry` messages, the server must return a InvalidSomethingType error.", i.e. they specify the particular error status the server must return in this situation.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -3010,6 +3018,14 @@ the `TableEntry` entity, which has the following fields:
...
+  `direct_resource_ids` in the p4info.
```
jafingerhut, line 3027: Suggest `P4Info` instead of `p4info` to be consistent with its capitalization in the rest of the spec.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4801,6 +4817,10 @@ message should be defined in a separate architecture-specific Protobuf file. See
 section on [Extending P4Runtime for non-PSA
 Architectures](#sec-extending-p4runtime) for more information.
 
+## `GenericTableEntry`
+See section [GenericTableEntry p4runtime](#sec-generic-table-entry) for more
+info
```
jafingerhut, line 4822: please add period at end of sentence

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
 `p4.v1.TableEntry` message based on the value of architecture-specific table
 properties, but we may include on in future versions of the API.
 
+# GenericTable { #sec-generic-table}
+
+## GenericTable p4info { #sec-p4info-generic-table}
```
jafingerhut, line 6329: Suggest using capitalization `P4Info` to be consistent with other occurrences in the spec.  I won't mention it for every occurrence in comments -- please do a global search and replace check on your PR.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+
+`GenericTable` messages are used to map non-PSA P4-externs or non-P4 target-
```
jafingerhut, line 6331: Recommend "P4 externs" with no hyphen

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+specific "fixed" features to their implementation in a generic way. The same can
+be achieved via `Extern`, but GenericTable provides a structured way in which
+every feature is represented as a set of match-fields and data-fields. The data
+consists of unions which are similar to actions. One use of GenericTable in a
```
jafingerhut, line 6335: I believe that the spec fairly consistently uses markup like `GenericTable` for all message names.  Please do search throughout your PR for message names to add these.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+server backend is with TDI [Table Driven
+Interface](https://github.com/p4lang/tdi), but targets can use Generic table to
+map to their own specific feature implementations as well.  It defines the
+following fields
+
+* `generic_table_type_id`, an 8-bit unsigned integer which uniquely identifies
+  the generic_table_type in the context of the architecture.
+  This value is in the range of `[0x00, 0xff]`.
+  The ID in the preamble is created as `0x18 + generic_table_type_id + 16 bits
```
jafingerhut, line 6344: If you mean something like the following, I suggest replacing the formula you wrote with the description below, since you seem to use `+` in two different meanings in the same formula, which can be very confusing.

The ID in the preamble is calculated as `(0x18 << 24) + (generic_table_type_id << 16) + (16-bit ID generated by the P4 compiler)`

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  of generated ID by compiler`. The `generic_table_type` depicts a new table
+  "type" which would have similar properties. The organization of the table
+  type space is left up to the target. Some optional guidelines
+  * For P4 externs, any new extern to have a new table type.
+  * For non-P4 externs, options are
```
jafingerhut, line 6349: If by "non-P4 externs" you mean "target-specific fixed function blocks" I would recommend using the latter phrase, to be consistent with earlier usage.

Also I personally find the phrase "non-P4 extern" to be just about a contradiction in terms.  `extern` is a P4 language construct.

jafingerhut: The list of detailed recommendations on how an implementer might manage the generic_table_type id space seems to interrupt the flow here.  Consider putting that in a separate small section focused on that purpose, perhaps?

jafingerhut: Note: If you create a separate small section focused on managing generic_table_type id, it would also be reasonable to put one or two examples of the different approaches you describe there, to make them easier to understand.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+      of the 8 bit type ID space soon.
+  Note that this value does not need
+  to be unique across all architectures from all organizations, since at any
+  given time every device managed by a P4Runtime server maps to a single P4Info
+  message and a single architecture.
+
+* `generic_table_type_name`, which specifies the fully-qualified P4 name of the
+  generic_table in case of P4 entities. In cases of non-P4 entities, it is a
+  name which uniquely identifies this entity in the entire space including P4
+  objects
```
jafingerhut, line 6363: period at the end of this and later sentences in this section.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+* `preamble`, a `Preamble` message with the ID, name, and alias of this
+  GenericTable.
+
+* `generic_match_fields`, a repeated field of type `GenericMatchField`
+  representing the data to be used to construct the lookup key matched in this
+  table. For information check the generic_match_fields info in the [Generic
+  Table section](#sec-generic-table-entry)
+
+* `union_refs`, a repeated `UnionRef` field representing the set of possible
+  unions for this table. Functionally, it behaves same as that of ActionRefs.
```
jafingerhut, line 6382: "unions" -> "union members" ?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  union's arguments.
+
+* `size`, an `int64` describing the desired number of table entries that the
+  target should support for the table.  See the "Size" subsection within the
+  "Table Properties" section of the P4~16~ language specification for details
+  [@P4TableProperties].
+
+* `other_properties`, an `Any` Protobuf message [@ProtoAny] to embed
+  architecture-specific or target-specific table properties [@P4TableProperties]
+  that the target wants to convey.
```
jafingerhut, line 6401: Out of curiosity, do you already have a use case in mind for other_properties?  Or just there for possible future growth?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+
+## `GenericData` { #sec-p4data-generic-data}
+
+`GenericData` is an addition to p4data.proto- which aims at adding support for
```
jafingerhut, line 6405: "which aims at adding support" -> "which adds support".  If it isn't hitting the target, we should update it so it does :-)

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+data types which aren't covered by P4Data. This incorporates all the data types
```
jafingerhut, line 6406: It feels a bit informal to me to use contractions in a specification document.  Please replace them with the full words, e.g. "are not" in this case.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+conflict, like bool, then the P4 data type should be used.
+
+  * bitstring : Bytes type used for simple unsigned integers or bytes
+  * float : Floating type values. Target might return a lower-precision
+    ceiled value even if field is non-volatile. Target might choose to show a
+    field as volatile if it allows user to insert a certain value but does its
+    own modifications to the value when writing to hardware.
+  * bool : Boolean type values
+  * generic_struct : Used to define a structure that can contain
+    one or more of the other members as named. One structure can contain
```
jafingerhut, line 6418: "as named" -> "as named fields" ?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+    different members of different types in GenericData.
+  * generic_list : Used to define a list of same types. Only allows for a
```
jafingerhut, line 6420: "a list of same types" -> "a list of values, all with the same type as each other"

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6324,380 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  architecture-specific or target-specific table properties [@P4TableProperties]
+  that the target wants to convey.
+
+## `GenericData` { #sec-p4data-generic-data}
+
+`GenericData` is an addition to p4data.proto- which aims at adding support for
+data types which aren't covered by P4Data. This incorporates all the data types
+used as primitive data types like bytes, float, strings. It contains p4type as
+well. P4type is to be used for any data type present in P4. If there is a
+conflict, like bool, then the P4 data type should be used.
```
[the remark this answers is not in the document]
saynb: The original idea was to use only `GenericData` for any non-P4 features. So if any fixed feature uses bool, we would want to use it from `GenericData` and not `P4Data`

jamescchoi (commented):
Requested some change.

docs/v1/P4Runtime-Spec.mdk
saynb: Left it there just for future growth

docs/v1/P4Runtime-Spec.mdk
saynb: Added a new section with some examples

proto/p4/config/v1/p4types.proto
```diff
@@ -292,4 +292,199 @@ message P4NewTypeSpec {
...
+    GenericNamedType string = 7;
+    GenericNamedType enum = 8;
+    GenericNamedType serializable_enum = 9;
+    GenericNamedType new_type = 10;
+    // P4 data type spec can help achieve parity with types being used
+    // in P4 externs. For example, if Register is using a type in P4
+    // Data type spec, then it will make it easier to map here when using
+    // GenericTables
+    P4DataTypeSpec p4_type = 11;
+  }
```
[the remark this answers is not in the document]
saynb: These can be done through struct and lists

proto/p4/config/v1/p4info.proto
```diff
@@ -397,3 +416,89 @@ message Digest {
...
+
+// All Tables of one type will be grouped in one message.
+message GenericTable {
+  uint32 generic_table_type_id = 1;
+  string generic_table_type_name = 2;
+  repeated string generic_table_properties = 3;
+  repeated GenericTableInstance instances = 4;
+}
+
+message GenericTableInstance {
```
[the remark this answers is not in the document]
saynb: Structured this like Externs instead of P4Tables

docs/v1/P4Runtime-Spec.mdk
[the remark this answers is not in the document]
saynb: Removed P4DataTypeSpec completely as discussed

proto/p4/config/v1/p4types.proto
```diff
@@ -292,4 +292,199 @@ message P4NewTypeSpec {
...
+  repeated string annotations = 4;
+  // Optional. If present, the location of `annotations[i]` is given by
+ // `annotation_locations[i]`.
+  repeated SourceLocation annotation_locations = 5;
+  repeated StructuredAnnotation structured_annotations = 6;
+}
+
+message GenericBitTypeSpec {
+  int32 bitwidth = 1;
+  int32 default_value = 2;
```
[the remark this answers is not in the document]
saynb: Made them as bytes

docs/v1/P4Runtime-Spec.mdk
jafingerhut: If by `+` you mean concatenation of bit strings, I would recommend using P4_16 syntax `++` for that, since all readers should be familiar with it.

jafingerhut: Never mind my previous comment.  I see you have used precise C/C++/P4_16 syntax that all readers should be familiar with.  Thanks.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+
+* `type_name`, which specifies the "type" of the `GenericTable`. For example,
+  if the target has a meter called `SpecialMeter`, then this would contain
+  "SpecialMeter". In cases of non-P4 entities, it is a "type" which helps
+  identify this entity with a feature for example "MulticastGroup". For non-
+  P4 entities, sometimes the instance name and type name are the same
+
+* `repeated string properties`, please check generic-table
+  properties under section [GenericTable categories]
+  (#sec-generic-table-categories) for more details.
```
jafingerhut, line 6401: I am not sure, but perhaps Madoko markup syntax rules require [GenericTable categories](#sec-generic-table-categories) to have no white space after `]` and before `(`.

You should try generating PDF and HTML output from the Madoko file to check not just this, but the entire new version of the document to see if there are any other Madoko markup issues.  I don't see precise instructions for installing all required software on a Linux system to build the P4Runtime API spec from source code yet, so I will create a PR to add some instructions.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+given time every device managed by a P4Runtime server maps to a single P4Info
+message and a single architecture.
+
+
+## GenericTable P4info { #sec-p4info-generic-table}
+
+`GenericTable` message in P4info defines a Generic Table. It contains several
+instances in `GenericTableInstance` rather than duplicating it every time for a
+new instance just like in `Extern`.
+It defines the following fields
```
jafingerhut, line 6387: The wording of the paragraph above confuses me.  Something like the following would be clearer to me (if it is correct):

A `GenericTable` message in P4info defines one type of generic table, but not a specific instance of a generic table.  Each instance of a generic table is defined by a `GenericTableInstance` message.  This separation allows field values that are common for all instances with the same generic table type to be defined in one place, the `GenericTable` message.  A `GenericTable` message contains the following fields:

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+
+* `type_id`, an 8-bit unsigned integer which uniquely identifies
+  the generic_table_type in the context of the architecture. More details
+  in the [Generic Table ID space](#sec-generic-table-id-space) section
```
jafingerhut, line 6391: Sorry for nitpicking, but the last sentence above is not a complete English sentence, but a sentence fragment.  It is perfectly understandable, but it is nice if we can use full sentences in a specification.  Perhaps replace it with something like the following:

See Section [Generic Table](#sec-generic-table-id-space) for more details.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+
+  * bitstring : Bytes type used for simple unsigned integers or bytes. Default
+    value applicable. If not present, it is 0.
+  * bool : Boolean type values. Default value applicable. If not present, it
+    is `false`.
+  * generic_struct : Used to define a structure that can contain
+    one or more of the other members as named fields. One structure can contain
+    different members of different types in GenericData.
+  * generic_list : Used to define a list of values with the same type. It can
+    also be a list of generic_structs or generic_list.
```
jafingerhut, line 6458: Can it be a list of elements, each of which is a generic_unordered_set?  If so, why only mention generic_struct or generic_list here?  If the element type can be any GenericData type, it seems best to simply say that.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+    Members are ordered and duplicates are allowed.
+  * generic_unordered_set : Used to define a set of same types. It can also be
+    a set of generic_structs. Members are unordered and duplicates are not
+    allowed.
```
jafingerhut, line 6462: Can members be any arbitrary GenericData type?  If so, why not say that explicitly, rather than mentioning generic_struct but no other type names?

Note: If duplicates are not allowed, then for every type allowed, you must define what equality means between values of those types somewhere, explicitly.  e.g. you must define equality between two arbitrary generic_struct values, between two arbitrary generic_unordered_set values, etc.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  * string : Used to define control plane strings. The max string length is
+    defined by the size. Default value applicable. If not present, it is empty
+    string: "".
+  * enum : String representation with which safe enums are realized. Default
+    value applicable. If not present, it is empty string: "".
+  * enum_val : bytes value with which unsafe/serializable enums are defined.
```
jafingerhut, line 6468: It seems better to follow P4_16 and call these simply "serializable enums".  Why call them "unsafe"?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+    also be a list of generic_structs or generic_list.
+    Members are ordered and duplicates are allowed.
+  * generic_unordered_set : Used to define a set of same types. It can also be
+    a set of generic_structs. Members are unordered and duplicates are not
+    allowed.
+  * string : Used to define control plane strings. The max string length is
+    defined by the size. Default value applicable. If not present, it is empty
+    string: "".
+  * enum : String representation with which safe enums are realized. Default
+    value applicable. If not present, it is empty string: "".
```
jafingerhut, line 6467: How is type enum different than string?

docs/v1/P4Runtime-Spec.mdk
saynb: p4runtime spec already uses both interchangeably https://p4.org/p4-spec/p4runtime/main/P4Runtime-Spec.html#sec-enum-serializable-enum-and-error
But I don't have strong objections to removing unsafe

docs/v1/P4Runtime-Spec.mdk
saynb: enums are strings with a list of valid choices

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+guidelines and semantics.  The idea has been borrowed from TDI [Table Driven
+Interface](https://github.com/p4lang/tdi), that every state can be representated
+as a table or multiple tables. A table here is any data structure with one or
+more entries and each entry can be represented as a set of key fields and data
+fields.  Actions in P4 tables behave like unions and that has been borrowed here
+in `GenericTable`.  This can be potentially used targets to map to either TDI
+based targets, or even non-TDI based targets. If using TDI, then potentially,
+for every new feature to be added, only TDI target support and remote client
+code will be required to make use of that function.  All other middle layers
+should theoretically remain unchanged.
```
jafingerhut, line 6485: I am not sure if I understand all of the paragraph above, or why it is here, especially starting at the sentence "This can be potentially used ...".  What do you think of replace everything from that sentence onward with a new paragraph like the one below?

Aside: Note that because of the close similarities between GenericTable and the TDI API, it is intended that relatively simple software could implement translation of GenericTable configuration messages into local TDI API calls.  Details of this are left for other future documents to specify.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+
+P4Runtime supports inserting, modifying, deleting and reading `GenericTable`
+entries with the `GenericTa bleEntry` entity, which has the following fields:
```
jafingerhut, line 6488: typo: incorrect space in middle of `GenericTableEntry`

jafingerhut: suggested replacement: "entity" -> "message" ?  It is not a big deal either way, but doesn't Protobuf use "message" pretty consistently for things like `GenericTableEntry`?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+deterministic in the data plane which entry a packet will match.  If a client
+wishes to make the matching behavior deterministic, it must use different
+priority values for any pair of table entries that the same packet matches.
+
+The `match` and `priority` fields are used to uniquely identify an entry within
+a table. Therefore, these fields cannot be modified after the entry has been
+inserted and must be provided for `MODIFY` and `DELETE` updates. When deleting
+an entry, these key fields (along with `is_default_action`) are the only fields
+considered by the server. All other fields must be ignored, even if they have
+nonsensical values (such as an invalid action field). In the case of a *keyless*
```
jafingerhut, line 6528: This looks like a copy & paste bug: there is no "action" field in GenericTableEntry.  Replace with `union`?

docs/v1/P4Runtime-Spec.mdk
jafingerhut: Where is the list of valid choices defined in this proposal?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+## GenericTable Categories {#sec-generic-table-categories}
+
+All tables fall into certain "categories" which define what APIs are
+applicable on the table and the APIs.
+
+  * Regular tables: Read, INSERT, MODIFY, DELETE work as intended. Entries in
+    a table can be inserted, modified and deleted.
+    * `INSERT` : If another entry with same match key exists, then
+      `ALREADY_EXISTS` is returned. If no union has been defined, then the
+      union from the default entry of the table is used. Some tables might
```
jafingerhut, line 6553: "If no union has been defined, then the union from the default entry of the table is used."

"no union has been defined" means that there is no `union` member in the GenericTableEntry message within the WriteRequest?  If so, probably best to say that explicitly.

The number of cases here for the proposed behavior of `INSERT` seems to cry out for some kind of pseudocode description to me.

I have no idea what "Const default union rules apply" can possibly be intended to mean.  I would recommend expanding on that to explain it better, or giving a cross-reference to some section that explains it more fully, if this idea is best explained in detail in one place and not repeated in multiple places.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+~
+[generic-tables-insert-flow]: \
+build/generic-tables-insert-flow.[svg,png] \
+{ height: 7cm; page-align: forcehere }
+
+    * `MODIFY` : Union field must be set since all data fields are under a
+      union. Modifies the union spec to new union or new field values if
+      same union.
+      If `default_entry` is true, then the default entry is modified. Match
+      fields and priority should not be set. Const default union rules apply.
```
jafingerhut, line 6570: MODIFY should mention `NOT_FOUND` is returned if the entry is not found, similar to how `DELETE` does, if that is the intent.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+    * `DELETE` : Only Match fields must be set. `NOT_FOUND` is returned if
+      the entry isn't found. If no Match fields are given, then all entries
+      are deleted (wildcard delete). If the table contained init entries, they
```
jafingerhut, line 6573: "If no Match fields are given, then all entries are deleted (wildcard delete)."

What if you earlier inserted an entry where all match fields were wildcard, so none of the Match fields were present (following P4Runtime TableEntry rules, which I am guessing also apply for GenericTableEntry?).  Then the client wishes to remove THAT ONE RULE.  It is not possible to do so with this specification, is it?

jafingerhut: Does the sentence "If the table contained init entries, they are restored to initial state with their original union and union field values." apply only for the case of wildcard delete?  If so, better to somehow make that more explicit, e.g. describing wildcard delete in a separate paragraph from the "normal" delete case.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+      union. Modifies the union spec to new union or new field values if
+      same union.
+      If `default_entry` is true, then the default entry is modified. Match
+      fields and priority should not be set. Const default union rules apply.
+    * `DELETE` : Instead of deleting, this operation resets the entry with
+      default values for every field. The union field values are changed
+      but the union itself isn't changed.
+      Only Match fields must be set. `NOT_FOUND` is returned if
+      the entry isn't found. If no Match fields are given, then all entries
+      are reset (wildcard reset). If the table contained init entries, they
```
jafingerhut, line 6599: Same comment here for wildcard reset as for wildcard delete on normal tables, above.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+      fields should not be set in this case.
+      Wildcard read does NOT return default entry.
+
+  * Read-only tables: Entries can only be read. Write RPC doesn't work and
+    `NOT_SUPPORTED` is returned.
+    * `READ` : The Read RPC. All entries can be read. If no entities are
+      provided in request, then all entries are read, AKA wildcard read.
+      If `default_entry` is true, then the default entry is returned. Match
+      fields should not be set in this case.
+      Wildcard read does NOT return default entry.
```
jafingerhut, line 6617: If `READ` behaves identically on all table types, I would recommend defining the behavior precisely IN ONE PLACE, and referring to it from each table type, rather than copying and pasting.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,409 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+  * Reset-only tables :
+    * `INSERT` : Not supported. Returns `NOT_SUPPORTED` error
+    * `MODIFY` : Not supported. Returns `NOT_SUPPORTED` error
+    * `DELETE` : Instead of deleting, this operation resets the entry with
+      default values for every field.
+      Only Match fields must be set. `NOT_FOUND` is returned if
+      the entry isn't found. If no Match fields are given, then all entries
+      are reset (wildcard reset). If the table contained init entries, they
+      are restored to initial state with their original union and union field
+      values.
```
jafingerhut, line 6628: Same comment for wildcard reset as for wildcard operations on other table types mentioned earlier.

docs/v1/P4Runtime-Spec.mdk
saynb: > What if you earlier inserted an entry where all match fields were wildcard, so none of the Match fields were present (following P4Runtime TableEntry rules, which I am guessing also apply for GenericTableEntry?). Then the client wishes to remove THAT ONE RULE. It is not possible to do so with this specification, is it?

I am not aware of any wildcard inserts in p4runtime spec. Can you please help point me to the right section?

docs/v1/P4Runtime-Spec.mdk
saynb: Added a section

docs/v1/P4Runtime-Spec.mdk
saynb: removed

docs/v1/P4Runtime-Spec.mdk
saynb: Have added list of valid choices

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: This is the relevant section: p4.org/p4-spec/docs/p4runtime-spec-working-draft-html-version.html#sec-match-format

Basically, any non-exact match can be wildcard. So if you don't have any `exact` keys, you can have a fully wildcard match.

jafingerhut: Clarification based on some discussion in most recent P4 API work group meeting.  Corrections welcome if I have made any mistakes here.

In the P4Runtime API specification, "wildcard" seems to be used _only_ to refer to the capability of a single small WriteRequest or ReadRequest message to operate on potentially many entries of a table.  For regular match-action tables, today P4Runtime API spec only supports wildcard reads of all entries of a table.  It defines no way to read a subset of entries that match some specified criteria on its key and/or action.  You do this by completely omitting the "match" sub-message in the Entry message of a ReadRequest.

I will use "don't-care" to refer to the ability in the P4Runtime API specification to specify a match criteria for a field in a table entry that matches all possible values for that field.  Normally you might think that you should be able to do so by specifying a mask of 0 for a ternary field, a prefix length of 0 for an lpm field, min=0 and max=`2^W-1` for a W-bit wide range field, or use the boolean flag for an optional field that indicates its value can be anything.  However, the P4Runtime API specification today mandates that such a field be completely omitted from a protobuf message.  I have written explicit `if` conditions in Python code for creating P4Runtime API request messages that had to check for this "don't-care" condition, and avoid included the field in the request message, or else the open source BMv2 P4Runtime API server rejected the request with an error, so enforcement of this rule is implemented today, I would guess in the p4lang/PI repo code.

Thus if you want to define wildcard operations for read, modify, or delete that operate on _some_ entries, but not _all_ of them, based upon some conditions specified in the request message, you will either have to create _different_ rules for forming GenericTable request messages than are used for Table (described in previous paragraph), or you will need to use some other way to specify that a request message is a wildcard request.  I would personally recommend that you consider adding a separate boolean flag to wildcard request messages.

jonathan-dilorenzo: > In the P4Runtime API specification, "wildcard" seems to be used only to refer to the capability of a single small WriteRequest or ReadRequest message to operate on potentially many entries of a table. For regular match-action tables, today P4Runtime API spec only supports wildcard reads of all entries of a table. It defines no way to read a subset of entries that match some specified criteria on its key and/or action. You do this by completely omitting the "match" sub-message in the Entry message of a ReadRequest.

I don't think that's quite right Andy. There's a slew of other possibilities in https://p4.org/p4-spec/docs/p4runtime-spec-working-draft-html-version.html#sec-table-wildcard-reads. There are also single entry reads (if the key is fully specified).  Or, at least the second is specified [here](https://p4.org/p4-spec/docs/p4runtime-spec-working-draft-html-version.html#sec-read-rpc), but technically if the entry contains any wildcard matches, then it isn't possible to fully specify it.

jafingerhut: Ah, sorry for my too-hasty message.  Thanks for the pointer to the correct answer.

I notice that it does say that for key fields within `match`, you can either target exactly one set of field values, if you specify any of them, or you can leave out the `match` sub-message entirely, and the targeted entries will not be constrained by the match fields at all.

I believe the following sentence in the description of `match` is perhaps wrong, or at least easily misleading: "The match will be exact, which means at most one entry will be returned."

Potential counterexample: You give a wildcard ReadRequest for a table with all `ternary` fields, and specify a match criteria for at least one of the fields, and a `priority` of 0.  If earlier you inserted that exact set of field values, but multiple separate entries with different priority values (which I believe is allowed by the P4Runtime spec), the wildcard ReadRequest seems like it should return all entries with that value of `match`, and all of the different priorities.

saynb: As permy understanding, unfortunately there is no way that you can have the match subfield in the `TableEntry` message but no fields specified, since it is a repeated field directly on the message. So there is no way to differentiate between wildcard read vs reading of the single "don't care" entry in case if there is only ternary field on the table and that table has an entry with mask == 0.
I propose that we table the discussion for the wildcard read for generic tables for a later time when we can have a bit more discussion on this

jonathan-dilorenzo: At the P4RT API WG, we discussed that Sayan should open up a general issue for this, and we'll leave this draft as it is for now (and merge it) until that has been settled.

saynb: Have created an issue here https://github.com/p4lang/p4runtime/issues/462

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -6304,6 +6328,425 @@ field [@ProtoAny]. At the moment, there is not any mechanism to extend the
...
+
+Properties exist at 3 different levels - Table, entry and field (match or union)
+
+### Table properties
+
+Table properties are at table level. A combination of these define a certain
+table category. By default, if absent, they are all always false
+
+* Read-only : No Add/Del/Mod work Modify-only : Table entries can only be
+* modifed Reset-only : Table can only be reset. Either entire table or
```
saynb, line 6677: needs correction

docs/v1/P4Runtime-Spec.mdk
saynb: It is not the same for all. It is different for Calculation tables

jamescchoi (approved):
All comments addressed.

chrispsommers:
@jafingerhut @smolkaj @jonathan-dilorenzo Last call for comments/approval. it;d be nice to merge this into the dev-branch which in any case is DRAFT.

@saynb Is this synced with main?

saynb:
> @jafingerhut @smolkaj @jonathan-dilorenzo Last call for comments/approval. it;d be nice to merge this into the dev-branch which in any case is DRAFT.
> 
> @saynb Is this synced with main?

The generic-table-dev branch needs to be synced with main. I can then git merge it in this private branch. Can you please help with that?

chrispsommers:
@saynb Sure, just make the PR and tag me, thanks.

jafingerhut (approved):
LGTM

Issue #439, opened 2023-07-06
Update any-branch-uploads.yml to avoid dependabot-caused CI failures


chrispsommers:
CI workflow in [any-branch-uploads.yml](https://github.com/p4lang/p4runtime/blob/main/.github/workflows/any-branch-uploads.yml) fails if a dependabot initiated the PR. See example failure at https://github.com/p4lang/p4runtime/actions/runs/5468855168/jobs/9957044088#step:5:13.

Per https://github.com/p4lang/p4runtime/pull/437#issuecomment-1622580483, need to add a conditional to the workflow, e.g. `if: ${{ github.actor != 'dependabot[bot]' }}`

PR #440, opened 2023-07-06
Fixes https://github.com/p4lang/p4runtime/issues/439

chrispsommers:
Change CI  workflow to skip publishing if PR spawned by dependabot

### DIFF
--- a/.github/workflows/any-branch-uploads.yml
+++ b/.github/workflows/any-branch-uploads.yml
@@ -16,6 +16,7 @@ jobs:
         docker run -v `pwd`/docs/v1:/usr/src/p4-spec p4lang/p4rt-madoko:latest make
         ls docs/v1/build
     - name: Upload spec to S3 if needed
+      if: ${{ github.actor != 'dependabot[bot]' }}
       uses: jakejarvis/s3-sync-action@v0.5.1
       with:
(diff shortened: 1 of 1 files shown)


.github/workflows/any-branch-uploads.yml
```diff
@@ -16,6 +16,7 @@ jobs:
         docker run -v `pwd`/docs/v1:/usr/src/p4-spec p4lang/p4rt-madoko:latest make
         ls docs/v1/build
     - name: Upload spec to S3 if needed
+      if: ${{ github.actor != 'dependabot[bot]' }}
```
antoninbas, line 19: I think you could condition the whole job instead of just this step:
```
if: ${{ github.repository == 'p4lang/p4runtime' && github.actor != 'dependabot[bot]' }}
```

chrispsommers: I thought of that, but then decided it wouldn't hurt to do the build part and just skip the publishing, to get more "coverage" (although it might be redundant with other workflows).  What do you think?

antoninbas: I don't think strongly about it. Dependabot PRs should have no impact on spec generation, but it's not like we have a high volume of such PRs anyway.

antoninbas (approved):

PR #445, opened 2023-08-09
WIP: Reconfiguration APIs to allow multiple P4 pipelines to be chained.

loalan:
The goal of these changes are:
- provide the primitive operations to allow the user to create a P4 service chain
- support adding, removing and reconfiguring of multiple P4 pipelines to a single device
- allow for dynamic pipeline changes at runtime (update)

These additions should be backwards compatible and include add, remove, enable and disable pipeline operations. Each pipeline has a unique pipeline ID. Entities are referenced per pipeline. SetForwardingPipelineConfig allows the user to specify the placement of a new pipeline relative to the reference/default pipeline in the ForwardingPipelineConfig

See overview at: https://drive.google.com/file/d/1V9mt1BKciCoMlHfIl5IZ65PaKA9jhaev/view?usp=sharing

Background paper: https://www.usenix.org/system/files/nsdi22-paper-xing.pdf


saynb:
Somewhat related to https://github.com/p4lang/p4runtime/issues/316 ?

chrispsommers:
My initial impression is that this is a useful, elegant and backwards compatible (!) change worthy of consideration. Thanks for submitting. I enabled a CI run out of habit, and noticed the generated code check failed. You can address this when you're ready by following the simple procedure described in https://github.com/p4lang/p4runtime#detailed-processes, see the paragraph beginning with "When updating the Protobuf files...".

chrispsommers:
> Somewhat related to #316 ?

Good point. It also raises another question in my mind - when specifying the `id` field in` PipelineId`, what assumptions if any are made about how this is mapped to a particular architecture's physical pipelines, e.g. if an ASIC has four physical pipes, each of which can accept a single `ForwardingPipelineConfig`, how would this map? And, imagining we carried through with https://github.com/p4lang/p4runtime/issues/316, would a conflict arise with this PR?

loalan:
> Somewhat related to #316 ?

Yes, it seems somewhat related at the protobuf level. However, the use case is quite different. I'll take a look and see if they can be unified somewhat.

smolkaj:
@loalan we noticed at the WG meeting that the slides are all flipped by 90 degrees. Any chance you could upload a version where things are oriented the usual way?

PR #459, opened 2023-09-08
Remove 4 P4 language spec compatibility issues from the list

jafingerhut:
During 2023-Sep-08 P4.org API work group meeting, it was agreed that there are no changes required to the P4Runtime API specification to be compatible with these updates in the language spec.


chrispsommers (approved):
LGTM

PR #468, opened 2023-12-07
Updating generic-table-dev branch with p4lang:p4runtime:main

saynb:

PR #471, opened 2024-01-27
Fix regression by pinning Bazel version 

smolkaj:
We had a regression on the main branch recently due to the Bazel version changing from 6.4.0 to 7.0, see https://github.com/p4lang/p4runtime/pull/468#issuecomment-1912872372 and follow up comments.

This PR fixes this regression by pinning the Bazel version to 6.4.0.


antoninbas (approved):
I thought we had done that already, but it must have been for p4lang/PI

smolkaj:
The `.bazelversion` file was in the root folder. That's typically the right thing to do, but in our case prevented it from being picked up because we use the `proto` subfolder as the root of our Bazel workspace as a workaround for github.com/grpc/grpc/issues/20675, see https://github.com/p4lang/p4runtime/blob/main/proto/BUILD.bazel#L25-L26.


:face_with_spiral_eyes: 

It is hacks and workarounds all the way down :)

PR #472, opened 2024-02-08
Support multicast resource information.

jonathan-dilorenzo:
On real hardware, multicast resources are presumably finite. While we have a way of capturing that for other tables (with the `size` parameter) and for action profiles (with the `max_group_size` parameter), there is currently no way to capture this information for multicast (or for the clone session table for that matter).

This PR is a first stab at adding such a mechanism. In my mind, these fields would be populated by users in the P4 program through the `@pkginfo` annotation (much like the rest of pkginfo).

Obviously, this is currently a WIP and I'd want to add something to the spec too, but I wanted to see if the idea even made sense first.


chrispsommers (commented):
Hi John, I think the idea of getting platform capabilities is a good one but I'm not necessarily in favor of making it part of "PkgInfo," which to me provides program-specific metadata, not platform-specific. Perhaps we need a generic platform capabilities query API returning KV pairs?

jonathan-dilorenzo:
Yeah, PkgInfo feels a bit weird.

But I don't think we quite want to query the platform for it... Rather, we want to tell the platform our desired capabilities and then it is up to the platform to reject them if it can't fulfill our desires or accept them otherwise. Like with all other tables really. So I think we'd want to fit it into the P4Info somehow... 

But good to discuss options in the meeting!

jonathan-dilorenzo:
Based on discussions in the P4 API WG with @chrispsommers, @smolkaj, and @jafingerhut, there seems to be tentative support for adding something like this (with a few changes I'm about to make) as a short-term solution to modeling multicast resources (and giving the switch a directive to reject a P4Info if it can't).

In the long-term, we want to look into treating multicast like any other P4 table, essentially specifying that a table is a multicast table by attaching something much like an ActionProfile except that it replicates packets to all actions instead of choosing one non-deterministically (or deterministically based on a hash function). 

A first approach to the long-term solution might be to disallow applying multicast tables in the P4 program, and then they would instead be applied where ever the architecture uses a multicast table. 

We also discussed that these would be useful to provide as capabilities, so I'm gonna try to add that in a follow-up PR. The idea will be that, if the numbers are given in the P4Info, then the capability numbers must be >= to those in the P4Info (or the P4Info should otherwise be rejected). If they are not given, then the switch can still be queried to see how many multicast entries and replicas it supports.

proto/p4/config/v1/p4info.proto
```diff
@@ -48,6 +48,19 @@ message Documentation {
   string description = 2;
 }
 
+// Used to describe the required properties of the underlying platform.
+message PlatformProperties {
```
jonathan-dilorenzo, line 52: @jafingerhut, @chrispsommers, and @smolkaj, I wrapped this into a PlatformProperties message as desired.

I'm thinking best to add it to the capabilities in a separate PR because it will need a totally separate description anyway. WDYT?

chrispsommers:
This seems fine as a working idea, but if I were doing this, I'd probably make changes in p4c to vet the idea a bit further.

jonathan-dilorenzo:
@chrispsommers, to be clear, the "changes in p4c" you're referring to is the annotation interpretation?

chrispsommers:
> @chrispsommers, to be clear, the "changes in p4c" you're referring to is the annotation interpretation?

Correct, the front-end which parses annotations and renders P4Info. You might find a desire to tweak the schema after trying to implement the enhancement.

jonathan-dilorenzo:
Added support for this PR in https://github.com/p4lang/p4c/pull/4611 (though this would need to go in first, so that the P4Info fields exist to be set).

jonathan-dilorenzo:
AI: Change `multicast_table` to `multicast_group_table` and then send out again for review and we'll try to get this in before next meeting.

jonathan-dilorenzo:
I think this is now ready for a final review as discussed @chrispsommers, @smolkaj, and @jafingerhut.

proto/p4/config/v1/p4info.proto
```diff
@@ -73,6 +86,10 @@ message PkgInfo {
   string url = 8;
   // Miscellaneous metadata, structured; a way to extend PkgInfo
   repeated StructuredAnnotation structured_annotations = 9;
+  // If set, specifies the properties that the underlying platform should have.
+  // If the platform does not conform to these properties, the server should
+  // reject the P4Info when used with a SetForwardingPipelineConfigRequest.
+  PlatformProperties platform_properties = 11;
```
chrispsommers, line 92: Are we missing `repeated SourceLocation annotation_locations = 10;` ?

jonathan-dilorenzo: Looks like it's above, under annotations.

chrispsommers: Ahh, you're right, they're not in order, I forgot to expand the diff. Thanks!

chrispsommers (approved):
LGTM

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1289,12 +1292,32 @@ message PkgInfo {
   string url = 8;
   // Miscellaneous metadata, structured; a way to extend PkgInfo
   repeated StructuredAnnotation structured_annotations = 9;
+  // If set, specifies the properties that the underlying platform should have.
+  // If the platform does conform to these properties, the server should reject
```
smolkaj, line 1296: does -> doesn't?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1289,12 +1292,32 @@ message PkgInfo {
...
+  // the P4Info when used with a SetForwardingPipelineConfigRequest.
```
smolkaj, line 1297: Nit: use backticks

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1289,12 +1292,32 @@ message PkgInfo {
...
+  PlatformProperties platform_properties = 11;
+}
+~ End Proto
+
+where the `PlatformProperties` message looks as follows:
+~ Begin Proto
+// Used to describe the required properties of the underlying platform.
+message PlatformProperties {
+  // The maximum number of multicast entries (i.e. multicast groups) that the
```
smolkaj, line 1306: maximum -> minimum
expected -> required
?

smolkaj: Same a few more times.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1327,6 +1352,17 @@ Above we see several different types of annotations:
 * `@description` - This will populate the `PkgInfo.doc.description` message
   field
 
+* `@platform_property` - This is used to populate a specific field within the
+  `PlatformProperty` message in `PkgInfo.platform_property`. Multiple
+  `@platform_property` annotations are allowed. For compactness,
+  multiple key-value pairs can appear in a single `@platform_property`
+  annotation, separated by commas. Each key must only appear once and the
+  compiler must reject the program if one appears multiple times. The `key`s
```
smolkaj, line 1360: Nit: Is "key" referring to code (a variable / proto field) here? If not, let's not put it in backticks.

smolkaj (changes requested):
A few nits

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: Err, I think I don't know where you mean to use backticks here? I'm not referring to any variable, which is it is used in the rest of the proto e.g.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: Good points. The `max_replicas_per_entry` is a bit tricky because, for the action profile equivalent, the server is required to reject anything greater than the `max_group_size`. I think this new wording works though.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: It is, yeah. It's referring to the code example above.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: Nice catch! Didn't copy from the proto when I fixed it :'(.

docs/v1/P4Runtime-Spec.mdk
smolkaj: `SetForwardingPipelineConfigRequest`

proto/p4/config/v1/p4info.proto
```diff
@@ -48,6 +48,20 @@ message Documentation {
   string description = 2;
 }
 
+// Used to describe the required properties of the underlying platform.
+message PlatformProperties {
+  // The minimum number of multicast entries (i.e. multicast groups) that the
+  // platform is required to support. If 0, there are no requirements.
+  int64 multicast_group_table_size = 1;
```
smolkaj, line 55: Any particular reason for using `int64` over `int32` here (and below)?

smolkaj (approved):
One more comment. LGTM otherwise. Approving early.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: That seems inconsistent with the general style here. Code is backticked, but message names are not. See e.g. https://github.com/p4lang/p4runtime/blob/main/proto/p4/config/v1/p4types.proto#L51-L52.

proto/p4/config/v1/p4info.proto
jonathan-dilorenzo: Good question! Simply to be symmetric to tables and action profiles where `size` is an `int64` (though `max_group_size` is an `int32`, so I should at least update that). Probably. I actually don't totally remember.

That said, [`multicast_group_id` is `uint32`](https://github.com/p4lang/p4runtime/blob/main/proto/p4/v1/p4runtime.proto#L452), so it's a bit questionable to have the size be `int64`... 

I'm convinced! Changed it all to `int32`.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: Definitely not totally obvious. In my mind, I suppose it's the difference between a type and an instantiation basically. For a type, I don't use the backticks and for an instantiation of it (like a variable) I do. That said, if I'm quoting a literal code segment rather than talking about the concept, I would also use backticks I suppose. 

That said, if you wanted to go change how it's done elsewhere, I'd be amenable to fixing it here too, even though I find it slightly worse, but I definitely prefer consistency over adding it as this weird one-off.

smolkaj: My preference is for consistently putting all code in backticks to keep things simple.
I agree it is not currently done consistently, so if that bothers you, fine to keep as is.

smolkaj: My preference is to consistently put all code in back ticks for simplicity.
Agreed that this is not currently done consistently. If that bothers you, fine to keep as is.

Issue #441, opened 2023-07-18
New protobuf entities should have a comment of the form "Added in x.y.z"


smolkaj:
Context: https://github.com/p4lang/p4runtime/pull/432#discussion_r1212101247

I propose that going forward, when adding new fields/messages/services to our P4Runtime protobuf schemas, we always include comments such as
```
// Added in 1.4.0.
```
This allows developers to reason about forward/backward compatibility of various sever/client combinations when the server and client may be using different versions of the protobuf schemas.

@jafingerhut @chrispsommers WDYT?

chrispsommers:
Seems like a good ideato me!

jafingerhut:
Agreed

PR #473, opened 2024-02-08
[Deprecated] Add support for different meter types.

jonathan-dilorenzo:
Currently, the P4Runtime specification only supports trTCM (two rate two color meters) as per RFC 2698. However, in real hardware these sometimes use more hardware resources compared to single rate two color meters. 

This PR is trying to introduce a way of restricting the meters clients can program to not use the full power of trTCM if they don't need it in order to save resources. I'm imagining the type of the meter being populated by users in the P4 program through an annotation on the meter itself (e.g. @single_rate_two_color or @two_rate_three_color). 

Obviously, this is currently a WIP and I'd want to add something to the spec too, but I wanted to check the proto idea first.


jonathan-dilorenzo:
Updated the specification with the related text.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4257,6 +4279,11 @@ message MeterConfig {
 }
 ~ End Proto
 
+A MeterConfig for a *Single Rate Two Color Marker* typed meter must only be
```
jonathan-dilorenzo, line 4282: We may only want to allow one representation to make in canonical, but I'm not sure which one is nicer... 

On the one hand, it's quite nice that the semantics of the meter are the same regardless of its type if you use `pir == cir` and `pburst == cburst`. 

On the other hand, it's a little weird to require people to set values that are fundamentally unused...

@smolkaj, @jafingerhut, and @chrispsommers, thoughts? I think I prefer only allowing the equality if we want a single representation.

jonathan-dilorenzo:
Plan from April WG meeting:
1. Require `pir == cir` and `pburst == cburst`.
2. Add Single Rate Three Color Marker (RFC 2697) type.
3. Add an EBS to MeterConfig to support 2

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: Decided on only allowing equality in the meeting.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4231,10 +4245,21 @@ entity for each of the instances, specifying the `counter_id` and
...
+
+The trTCM meters an arbitrary packet stream using two configured rates ---
+the Peak Information Rate (PIR) and Committed Information Rate (CIR), and their
+associated burst sizes --- and "marks" its packets as GREEN, YELLOW or RED based
+on the observed rate.
+
+The *Single Rate Two Color Marker* meters an arbitary packet stream using a
+single configured rate --- the Committed Information Rate (CIR) and its
+associated burst size --- and "marks" its packets as GREEN or RED based on the
+observed rate.
```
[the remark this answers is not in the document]
jonathan-dilorenzo: Yupp, good idea. Coming in a PR update in  a few minutes!

jonathan-dilorenzo:
All of the AIs should now be fixed. PTAL!

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1712,14 +1712,33 @@ Both `Meter` and `DirectMeter` messages share the following fields:
...
       expressed in bytes/second.
     * `PACKETS`, for rates expressed in packets/second.
 
+  The meter type can be any of the `MeterSpec.Type` enum values:
+    * `TWO_RATE_THREE_COLOR`: This is the *Two Rate Three Color Marker* (trTCM)
+      defined in RFC 2698 [@RFC2698]. This is the standard P4Runtime meter type
+      and allows meters to use two rates to split packets into three potential
+      colors: GREEN, YELLOW, or RED. This mode is the default, but can also be
+      set explicitly in a P4 program by adding the `@two_rate_three_color`
+      annotation to the meter definition.
```
smolkaj, line 1729: Nit: can we add a code example. since it may not be obvious where to add the annotation exactly?

Probably sufficient to add it only for this case, since the other cases should be very similar.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1712,14 +1712,33 @@ Both `Meter` and `DirectMeter` messages share the following fields:
...
+    * `SINGLE_RATE_THREE_COLOR`: This is the *Single Rate Three Color Marker*
+      (srTCM) defined in RFC 2697 [@RFC2697]. This allows meters to use one rate
+      and an Excess Burst Size (EBS) to split packets into three potential
+      colors: GREEN, YELLOW, or RED. In a P4 program, this mode can be set by
+      adding the `@single_rate_three_color` annotation to the meter definition.
+    * `SINGLE_RATE_TWO_COLOR`: This is a simplified version of RFC 2697
+      [@RFC2697], which restricts meters to use only a single rate to mark
```
smolkaj, line 1736: Can we explicitly say which single rate?

smolkaj (changes requested):
This looks great! Some initial comment.

One reminder: let's say when the new fields were added following https://github.com/p4lang/p4runtime/issues/441.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: Yupp, added.

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: Good idea. Added. Not sure if it should go in the bullet or after though. Let's see how it looks.

jonathan-dilorenzo:
Looks like I somehow have to rebase this on main. Will figure it out :'(

jonathan-dilorenzo:
I don't understand why this is complaining. I thought rebasing onto main and then pushing would resolve the merge conflicts :'(. If anyone knows git better than I (a trivial task), is there something clever I can do here to resolve this locally and then push? Resolving conflict of binaries is... painful ;).

smolkaj:
I don't know where your merge conflicts are.

But can you just check out the version from main instead of merging?
```
git checkout origin/main <file>
```

smolkaj:
Unfortunately not really reviewable at the moment since it seems to contain changes from other PRs in the diff.

jonathan-dilorenzo:
Deprecating this in favor of https://github.com/p4lang/p4runtime/pull/484, since apparently, merge conflicts are beyond my capabilities ;D

PR #481, opened 2024-05-11
Clarify wildcard read of counter and meter

### LINKED ISSUE #478: What is the expect behavior of wildcard reading CounterEntry with `counter_id = 0` and `index = x`?
I'm reading the doc of reading and writing CounterEntry and getting confused about wildcard reading.

In the [doc](https://github.com/p4lang/p4runtime/blob/main/docs/v1/P4Runtime-Spec.mdk#L4260-L4264):
```
* If the `counter_id` field is set to 0 (default), the server returns the
  counter values for all indirect counter instances in the `ReadResponse`.

* If the `index` field is not set, the server returns the counter values for all
  indirect counters in the array identified by the unique id `counter_id`.
```
> Is `for all indirect counters` actually `for all indirect entries`? It seems like a typo.

In the [proto](https://github.com/p4lang/p4runtime/blob/main/proto/p4/v1/p4runtime.proto#L391-L392):
```
// - All counter cells for all counters if counter_id = 0 (default).
// - All counter cells for given counter_id if index is unset (default).
```

To my understanding, the results of different parameters are:
- `counter_id: x, index: y` => one entry of counter `x` and its index is `y`
- `counter_id: x, index: UNSET` => all entries of counter `x`
- `counter_id: 0, index: UNSET` => all entries of all counters
- `counter_id: 0, index: y` => Undefined behavior? Or all entries that index are `y`? Or does some error indicate that the parameter is illegal?

duskmoon314:
Add 'Otherwise' to clarify how `id` and `index` are used in wildcard read of counter and meter.

Close #478

### DIFF
--- a/docs/v1/P4Runtime-Spec.mdk
+++ b/docs/v1/P4Runtime-Spec.mdk
@@ -4255,13 +4255,14 @@ array have default value 0.
 A P4Runtime client may request to read the counter values of one or more
 indirect counter instances with a `ReadRequest` by including a `CounterEntry`
 entity for each of the instances, specifying the `counter_id` and
-`index`. Wildcard reads are also supported as follows.
+`index`. Wildcard reads are also supported as follows:
 
 * If the `counter_id` field is set to 0 (default), the server returns the
   counter values for all indirect counter instances in the `ReadResponse`.
 
-* If the `index` field is not set, the server returns the counter values for all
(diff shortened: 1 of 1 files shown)


antoninbas (approved):

antoninbas:
@chrispsommers @smolkaj I'll let you merge if this looks good to you

chrispsommers (approved):

PR #484, opened 2024-06-18
Add support for different meter types.

jonathan-dilorenzo:
From a new branch because I couldn't figure out how to update PR #473. See that PR for more information.


smolkaj (approved):
This looks good on my end. Can we get a second review, perhaps from @chrispsommers or @jafingerhut?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4292,9 +4333,26 @@ message MeterConfig {
...
+should return an `INVALID_ARGUMENT` error.
+
+A MeterConfig for a *Single Rate Two Color Marker* typed meter must only be
+accepted if `pir` and `pburst` are equal to `cir` and `cburst` respectively and
+`eburst` is unset. Otherwise, the server should return an `INVALID_ARGUMENT`
+error.
+
+Note: Any acceptable MeterConfig for a *Single Rate Two Color Marker* typed
+meter is also acceptable for either of the other two meter types and will mark
+packets identically in all three cases.
```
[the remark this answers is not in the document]
jonathan-dilorenzo: Done!

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -1749,14 +1749,39 @@ Both `Meter` and `DirectMeter` messages share the following fields:
...
     * `BYTES`, which signifies that this meter can be configured with rates
       expressed in bytes/second.
     * `PACKETS`, for rates expressed in packets/second.
 
+  The meter type can be any of the `MeterSpec.Type` enum values:
+    * `TWO_RATE_THREE_COLOR`: This is the *Two Rate Three Color Marker* (trTCM)
+      defined in RFC 2698 [@RFC2698]. This is the standard P4Runtime meter type
+      and allows meters to use two rates to split packets into three potential
+      colors: GREEN, YELLOW, or RED. This mode is the default, but can also be
+      set explicitly in a P4 program by adding the `@two_rate_three_color`
```
chrispsommers, line 1765: Is there a PR for the P4-16 Language Spec to reserve these new annotations? If so, it'd be nice to provide to cross-references in each PR to the other. If not, should this be initiated before merging this?

chrispsommers: Also, what are the implications for the various p4c backends? I can't recall if any platforms already support this and you're just extending P4RT to accommodate it. Thanks.

jonathan-dilorenzo: Good questions!

> Is there a PR for the P4-16 Language Spec to reserve these new annotations?

There is no such PR. We haven't been adding P4Runtime annotations to the P4 spec (see https://p4.org/p4-spec/docs/p4-16-working-draft.html#sec-p4-reserved-annotations), but maybe we should... Or maybe we should include a reference from the relevant appendix in the P4 Spec to the P4Runtime annotations and note that other P4-related standards may pre-define their own standardized annotations (which ideally shouldn't overlap)?

WDYT? Regardless, wouldn't block merging this PR on the decision.

> Also, what are the implications for the various p4c backends?

So, I think the implication would be basically precisely what the PR says. They should reject entries with invalid meter configs and create the related meters. They should also be able to say that they don't support a particular type though... Maybe I should add something to that effect to the PR?

chrispsommers: Hi Jonathan, thanks for the responses.
Re: Reeserved amnotations - thanks for refreshing my memory, we don't cite P4RT annos there, it's left as an exercise to the reader although somehow I thought we had a pointer to the P4RT repo so folks could check; guess not.

Re: Backends, I'm still unclear how you can use this change - is there a backend which can handle it?

smolkaj: > Re: Backends, I'm still unclear how you can use this change - is there a backend which can handle it?

Our "backend" handles it :)
But note that we don't actually have our own p4c backend. We just use the p4info backend and consume that.

smolkaj: Also, I really like Jonathan's suggestion to link from the P4 spec to the P4Runtime annotatons (but agree this PR should not be blocked on that).

chrispsommers: > > Re: Backends, I'm still unclear how you can use this change - is there a backend which can handle it?
> 
> Our "backend" handles it :) But note that we don't actually have our own p4c backend. We just use the p4info backend and consume that.

Ahh, OK, makes sense. Thanks!

chrispsommers (approved):
LGTM

PR #486, opened 2024-07-01
Add information about the initial default action to tables in the P4Info file.

### LINKED ISSUE #470: P4Info Table is missing information about its initial default action
I am working on a project which consumes a P4Info message but ran into a problem with tables. For a table like this
```P4
table toggle_check {
    key = {
        h.ethernet.dst_addr : exact @name("dst_eth");
    }
    actions = {
        do_something();
    }
    default action = do_something();
}
```
the corresponding P4Info message will look like this:
```Protobuf
tables {
  preamble {
    id: 40116169
    name: "toggle_check"

  }
  match_fields {
    id: 1
    name: "dst_eth"
    match_type: EXACT
  }
  action_refs {
    id: 31609776
  }
  size: 1024
}
```
The message loses information about the initial default action of the table. This is because the default action has not been marked const and thus the `const_default_action_id` field is not set. To get this information I have to look up the table in the P4 program, then identify its default action, which requires a series of convoluted steps. I can not simply rely on the P4Info file. 

Maybe we could add a second field, `initial_default_action_id`, which specifies the initial default action of the table? At least on the compiler side, it is trivial to add and this should not break the existing API.

fruffy:
Add information about the initial default action of a Table and the arguments this action is executed with. This information can be useful for static analysis or initializing a P4Runtime target. 

Corresponding PR in P4C: https://github.com/p4lang/p4c/pull/4773
Fixes #470.


proto/p4/config/v1/p4info.proto
```diff
@@ -217,6 +217,9 @@ message Table {
   repeated ActionRef action_refs = 3;
   // 0 (default value) means that the table does not have a const default action
   uint32 const_default_action_id = 4;
+  // The initial default action of the table. This can be overridden at runtime.
+  // 0 if there is no default action.
+  uint32 initial_default_action_id = 5;
```
jafingerhut, line 222: Please add a comment "Added in 1.4.0" before the new field.

proto/p4/config/v1/p4info.proto
```diff
@@ -217,6 +217,9 @@ message Table {
   repeated ActionRef action_refs = 3;
   // 0 (default value) means that the table does not have a const default action
   uint32 const_default_action_id = 4;
+  // The initial default action of the table. This can be overridden at runtime.
+  // 0 if there is no default action.
```
jafingerhut, line 221: "0 if there is no default action".  I am confused by this.  Even for P4 tables with no explicit `default_action` property, the P4 language spec says that the default_action for such a table is `NoAction`, and in p4c today this is an explicit separate action in the P4Info file with a non-0 action id.

jafingerhut: It appears this sentence may have been copied from the comments describing const_default_action_id above, but note that it is definitely possible to define a table that does not include `const default_action = foo`, whereas I do not think it is possible to define a table that has no default_action at all.

jafingerhut:
Is it planned to add to this PR modifications to the specification document describing this new field?  Or perhaps part of a separate PR?

proto/p4/config/v1/p4info.proto
fruffy: I have added the sentence because the current implementation allows for default actions to be missing from a table: https://github.com/p4lang/p4c/blob/main/control-plane/p4RuntimeSerializer.cpp#L258

jafingerhut: Is it actually possible to hit that case when executing that code, given the current front end compiler code?  I suspect the answer is "no".

fruffy: With the current front end I do not think so, but the front end is rather bloated and is likely to change: https://github.com/p4lang/p4c/issues/4634

I think it makes sense to throw an error here then.

jafingerhut: Even if the front end changes radically, this behavior of treating a table without a source-code-specified `default_action` as one with `default_action = NoAction` is codified in the language spec, so it is one aspect of the front end that I believe should _not_ change.

See Section 14.2: "The compiler must set the default_action to NoAction (and also insert it into the list of actions) for tables that do not define the default_action property.  Hence, all tables can be thought of
as having a default_action` property, either implicitly or explicitly."

fruffy:
I realized that this does not cover default arguments of the action. I think this requires adding a new message type.

fruffy:
@jafingerhut Do you know whether there are any specific restrictions on default action instantiations on tables? Otherwise, we could just create a simple P4Info message that describes an action call. This can also be reused for entries, in case we implement them.

jafingerhut:
The current file created by p4c for const entries (and initial entries) is a sequence of text-format Protobuf WriteRequest messages.

Perhaps for initial default actions, a WriteRequest that does a MODIFY (not an INSERT, which is not allowed on default actions at run time), and has the field is_default_entry=true in the TableEntry message, would be appropriate for this purpose?

To have that in the P4Info file would probably require copying, and/or somehow otherwise referencing, many of the definitions in the runtime Protobuf message definitions, which are not now part of the P4Info message definitions.

jafingerhut:
Even if such a message was not WriteRequest, then Update, then TableEntry, at least the TableEntry part would be sufficient.  I suspect even that would require a large number of other sub-message definitions to be available in the P4Info file message.

fruffy:
I was thinking many of the features of a write request are not required here. 
```
message TableAction {
  oneof type {
    Action action = 1;
    uint32 action_profile_member_id = 2;
    uint32 action_profile_group_id = 3;
    ActionProfileActionSet action_profile_action_set = 4;
  }
}

message Action {
  uint32 action_id = 1;
  message Param {
    uint32 param_id = 2;
    bytes value = 3;
  }
  repeated Param params = 4;
}
```
Something of this form might be the only required entities. The question is whether we even need the action profile specification, I have not tried whether a default action profile entry is possible.

jafingerhut:
I agree that not everything that is in the TableEntry message is required for default actions.  The one or two things that might make it worthwhile are the entry metadata.  It would be good to hear from other P4Runtime users whether they use the metadata fields for default actions as well as table entries.

jafingerhut:
@smolkaj @chrispsommers For default actions on tables, do either of you know what kinds of annotations and/or metadata someone might want to include in a P4 source file that could attach to the initial default action of the table, and would be nice to preserve in the output of the compiler describing that initial default action?

chrispsommers:
Hi Andy, I've nothing to contribute here. Any standard annotations like @brief() or @doc() would presumably be preserved in the P4info, but I think you're hinting at possibly other existing, or perhaps new, annotations to annotate default action params, etc?

> For default actions on tables, do either of you know what kinds of annotations and/or metadata someone might want to include in a P4 source file that could attach to the initial default action of the table, and would be nice to preserve in the output of the compiler describing that initial default action

jafingerhut:
@fruffy Let me know if you want more details, or would prefer if I created a separate PR to carry this torch, but here was a suggestion resulting from today's P4 API working group:

Instead of having a new field `initial_default_action_id` in the `Table` message as this PR proposes right now, do the following:

+ Copy and paste `message Action` from the file p4runtime.proto into p4info.proto.  Rename it to something that has no name conflicts in the p4info.proto file, perhaps something like `ActionSpec`.  This contains all the fields necessary to specify an action id, and values for all of its run-time parameters.

+ Add a new field `ActionSpec default_action_spec = <new_id>` to message `Table` in p4info.proto.  This would contain the action id and the action parameter values for the default_action of the table, _regardless_ of whether it was `default_action` or `const default_action`, and would still be present even if there was no `default_action` in the source code for the table definition.  In the last case, `default_action_spec` would contain an action id for a `NoAction` action, and 0 parameter values.

+ Yes, this does mean that the existing field of `Table` named `const_default_action_id` would become redundant for tables with `const default_action = ....` declared.  In the interest of backwards compatibility, it would stay in the `Table` message unchanged.  The value of `const_default_action_id` being 0 would also be the only way that a reader of the P4Info `Table` message could use to distinguish whether the `default_action` of the table was declared `const` or not.

fruffy:
Sounds good to me! I can give this a crack. I can also reuse your descriptions for the documentation.

chrispsommers:
See https://github.com/p4lang/p4runtime/issues/470

chrispsommers:
@fruffy bump - can we do this by end of month to meet 1.4.0 release goals? Thx.

fruffy:
> @fruffy bump - can we do this by end of month to meet 1.4.0 release goals? Thx.

Currently have a lot on my plate. Will do my best but can not guarantee I will finish the implementation on both the compiler and here until then. :/

smolkaj:
No worries, there will be a 1.5 if we can't make it in time.

proto/p4/config/v1/p4info.proto
```diff
@@ -217,6 +230,8 @@ message Table {
   repeated ActionRef action_refs = 3;
   // 0 (default value) means that the table does not have a const default action
   uint32 const_default_action_id = 4;
+  // The initial default action of the table. This can be overridden at runtime.
```
jafingerhut, line 233: Please add a comment "Added in v1.4.0" before new fields of messages.

proto/p4/config/v1/p4info.proto
```diff
@@ -207,6 +207,19 @@ message MatchField {
   repeated StructuredAnnotation structured_annotations = 9;
 }
 
+// A TableActionCall references a particular action id and executes the
+// action the supplied list of arguments.
```
jafingerhut, line 211: typo?  Perhaps "action with the supplied list of arguments"?

chrispsommers (commented):
@fruffy Thanks for expediting this! It looks good to me, but before I "approve" I have a question. We often like to have a p4c front-end validation (or at least a proof-of-concept) of P4info schema changes when there's something new like this. It's not mandatory, but desirable. Are you intending to implement the p4c front-end changes, and if so, is there any possibility of doing so, say, by end of the month?

fruffy:
> @fruffy Thanks for expediting this! It looks good to me, but before I "approve" I have a question. We often like to have a p4c front-end validation (or at least a proof-of-concept) of P4info schema changes when there's something new like this. It's not mandatory, but desirable. Are you intending to implement the p4c front-end changes, and if so, is there any possibility of doing so, say, by end of the month?

The corresponding PR is here https://github.com/p4lang/p4c/pull/4773

chrispsommers (approved):
LGTM, thanks!

fruffy:
> LGTM, thanks!

How's your merge process for these things? Should I just go ahead and merge or wait for @smolkaj?

chrispsommers:
@smolkaj Could you please take a look? @fruffy Let's aim to merge by Friday unless concerns are raised. You're welcome to do so.

chrispsommers:
Thanks Fabian!

PR #493, opened 2024-08-09
Update Bazel example to point to 1.4.0

smolkaj:


chrispsommers (approved):
@smolkaj @antoninbas What steps will eventually result in making the 1.4.0 tarball available? I noticed the 1.3.0 is accessible, but the 1.4.0 one isn't.

smolkaj:
We need to actually start cutting a release, see screenshot.

image

chrispsommers:
Got it, thanks!

PR #499, opened 2024-09-13
Version 1.4.0 rc updates to bazel and protobuf comments

chrispsommers:
Merge accumulated 1.4.0 updates by consensus in WG mtg:
- Add missing comments "Added in 1.4.0" and "Deprecated in 1.4.0" to protobufs
- Update script bazel comments to reflect version 1.4.0
- Regenerate protobuf client code

### DIFF
--- a/bazel/example/WORKSPACE.bazel
+++ b/bazel/example/WORKSPACE.bazel
@@ -11,8 +11,8 @@ local_repository(
 
 # http_archive(
 #     name = "com_github_p4lang_p4runtime",
-#     urls = ["https://github.com/p4lang/p4runtime/archive/v1.3.0.tar.gz"],
-#     strip_prefix = "p4runtime-1.3.0/proto",
+#     urls = ["https://github.com/p4lang/p4runtime/archive/v1.4.0.tar.gz"],
+#     strip_prefix = "p4runtime-1.4.0/proto",
 #     # sha256 = "<insert hash value here>",
 # )
 
@@ -21,7 +21,7 @@ local_repository(
 #   remote = "https://github.com/p4lang/p4runtime.git",
 #   # strip_prefix = "proto",  # https://github.com/bazelbuild/bazel/issues/10062
 #   patch_cmds = ["mv proto/* ."],  # Workaround since strip_prefix is broken.
-#   tag = "v1.3.0",
+#   tag = "v1.4.0",
 # )
 
 load("@com_github_p4lang_p4runtime//:p4runtime_deps.bzl", "p4runtime_deps")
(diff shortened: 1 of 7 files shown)


jafingerhut (approved):

smolkaj (approved):

smolkaj:
FYI, started a "prerelease" for 1.4.0 so the URLs in the Bazel rules work: https://github.com/p4lang/p4runtime/releases/tag/v1.4.0

antoninbas:
> FYI, started a "prerelease" for 1.4.0 so the URLs in the Bazel rules work: https://github.com/p4lang/p4runtime/releases/tag/v1.4.0

@smolkaj Once you create a tag (in this case v1.4.0) in the Git upstream, you are never supposed to mutate that tag (i.e., mutate the commit referenced by that tag, which is currently f50fef914398498fa97434b4b0183a6b5f74fc3f). That's pretty much a golden rule, and it can break things if you do (e.g., Go modules - it's very much a pain to repair properly and can only be done by releasing a new greater version). Did you mean to create that tag already? Marking the Github release as a "pre-release" doesn't make a difference here, because the tag is still created the same way.

If f50fef914398498fa97434b4b0183a6b5f74fc3f was the intended commit for the v1.4.0 tag all along, please ignore, but it doesn't seem to be the case, given the existence of this PR. The only way to fix it without breaking anything would be to leave the v1.4.0 tag alone now, and release v1.4.1 with all the necessary extra changes.

smolkaj:
Thanks for catching that. IIUC, you are saying that tags should always be immutable?

Would simply deleting the tag again address your concerns? Presumably no one has started using it quite yet :)

antoninbas:
> Thanks for catching that. IIUC, you are saying that tags should always be immutable?
> 
> Would simply deleting the tag again address your concerns? Presumably no one has started using it quite yet :)

No deleting the tag doesn't work.
The tag would already have been cache by https://proxy.golang.org/ (presumably right after you created it). See how it is already in the Go documentation: https://pkg.go.dev/github.com/p4lang/p4runtime?tab=versions
Mutating the tag, deleting the tag, deleting the tag in order to recreate it later (which is essentially the same as mutating it) can all create issues for downstream modules (again in the case of Golang) because of conflicts with https://sum.golang.org/
Even outside of Golang specifically, mutating a tag can create issues for software that depends on the project. Some processes which observe a tag that has changed, or release assets that have changed, can fail as it gets reported as a supply chain attack.
The only solution is to move on to v1.4.1 without changing v1.4.0.

smolkaj:
Thanks for elaborating, Antonin, I was not aware of that. 
SG regarding moving to v1.4.1.

PR #500, opened 2024-09-14
Pin the version of the rust code generation tool

duskmoon314:
In the previous commit, the versions of rust, protoc-gen-prost, and so on are not pinned. That is, the latest version is used. This will cause unexpected CI failure for commits unrelated to rust gen.

This commit pins the version in the Dockerfile:

rust: 1.81.0
protoc-gen-prost: 0.4.0
protoc-gen-prost-crate: 0.4.1
protoc-gen-tonic: 0.4.1

The dependencies are also updated:

prost: 0.13.1 -> 0.13.2
tonic: 0.12.0 -> 0.12.1

The rust code is regenerated, and the version is bumped to 1.4.0


chrispsommers (approved):

chrispsommers:
@duskmoon314, Thanks for fixing this so quickly, and thanks for contributing Rust protobufs to P4Runtime!

PR #502, opened 2024-09-24
Pin the dependencies of the Rust code gen tools

duskmoon314:
In the previous commit, only the version of `protoc-gen-prost/tonic` was pinned. However, Cargo would still ignore the lock file and use the latest version of dependencies for `protoc-gen-prost/tonic`. This meant that any new release of dependencies (such as `prost-build`) could result in differences in the generated code.

This commit addresses the issue by adding `--locked` to the `cargo install` command, ensuring Cargo respects the lock file when installing code generation tools.

The Rust code has been regenerated using this updated configuration.


duskmoon314:
There are 206 additions and 56 deletions. Changes in rust code are caused by the version of `prost` being pinned back to the lock file of `protoc-gen-prost` rather than the version of the last time we built it.

chrispsommers (approved):

chrispsommers:
It worked, thanks again @duskmoon314 for fixing this so quickly!

Issue #505, opened 2024-10-11
Protobuf build broken at HEAD


smolkaj:
The protobuf build is [broken at HEAD](https://github.com/p4lang/p4runtime/actions/runs/11296486710/job/31421454650) as of https://github.com/p4lang/p4runtime/commit/126bef23dc7692fc06f7ded44568754cbeb5ba7c. That commit should be a no-op from a code perspective, so unclear why we're seeing this regression
```
ERROR: /home/runner/.cache/bazel/_bazel_runner/1fc64d2e702c0d08ecdb88fe41750bb3/external/local_config_python/_python3/BUILD:25:8: declared output 'external/local_config_python/_python3/_python3_include/Python.h' is a dangling symbolic link
ERROR: /home/runner/.cache/bazel/_bazel_runner/1fc64d2e702c0d08ecdb88fe41750bb3/external/local_config_python/_python3/BUILD:25:8: declared output 'external/local_config_python/_python3/_python3_include/abstract.h' is a dangling symbolic link
...
```

![image](https://github.com/user-attachments/assets/20dd97f3-35a0-49c1-9f50-cd6cd322a33d)

PR #506, opened 2024-10-11
Update depdencies to latest versions.

smolkaj:
Fixes https://github.com/p4lang/p4runtime/issues/505.


smolkaj:
@chrispsommers, this doesn't do anything interesting other than updating a bunch of dependencies. This is often the easiest way to fix sudden build regressions :)

chrispsommers:
Is there any lesson or advice here beyond your comment, which might help posterity?

smolkaj:
The other lesson -- and what very likely triggered this issue -- is that our CI scripts currently use `ubuntu-latest`, which is a moving target. Specifically, in `.github/workflows/ci-build-proto.yml`, we currently have:
```
  build:
    runs-on: ubuntu-latest
```

And, sure enough, `ubuntu-latest` was moved from Ubuntu22.04 to Ubuntu 24.04 2 weeks ago: https://github.com/actions/runner-images/commit/ae99c16b0cf27c0cca7ef0b08e7bcf13c33f6cfa.

It is typical for this to cause regressions, and it is typical that they can be fixed by simply updating all dependencies to their latest versions.

If we wanted, we could use a stable label such as `ubuntu-24.04` instead of the moving label `ubuntu-latest` to avoid such regression. But I actually think this is a healthy forcing function to prevent bit rod, so would prefer to keep things as is.

smolkaj:
Shall we get this merged, @chrispsommers ?

chrispsommers (approved):

PR #507, opened 2024-10-11
Final changes for 1.4.1 release, including version updates.

smolkaj:
I used https://github.com/p4lang/p4runtime/pull/327 as an example, and followed the instructions mentioned in https://github.com/p4lang/p4runtime/pull/327.

PR #508, opened 2024-10-21
Support fallback ports in replicas for multicast

matthewtlam:
We wish to support fallback ports for multicast by adding BackupReplicas to Replicas when the primary port in the Replica becomes unavailable.


matthewtlam:
cc: @smolkaj for visibility

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4678,7 +4682,6 @@ semantics:
   service for each clone packet instance will be set to the value programmed in
   the clone session entry (`class_of_service` field). This value must be a valid
   value for the PSA `ClassOfService_t` type, which supports runtime translation
-  by default [@PSATranslation], or the server must return
```
antoninbas, line 4681: that looks like an unrelated (and incorrect) change?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4550,7 +4552,9 @@ The egress packets may be distinguished for further processing in the egress
 using the `instance` metadata. Note that a packet may not be both unicast and
 multicast; if the multicast group is set, it will override the unicast egress
 port. If the P4 `multicast_group` metadata is set to a value that is not
-programmed in the PRE, then the packet is dropped.
+programmed in the PRE, then the packet is dropped. Fallback ports for multicast
+uses one of the backup replicas when the primary multicast `port` becomes
+unavailable.
```
antoninbas, line 4557: that seems a bit too abstract for me. There are multiple "primary" multicast ports (in the example above: 5, 12, 18 and 24). So what happens when a single one fails? What happens when multiple ones fail (in particular if there are fewer backups that "primaries")?

proto/p4/v1/p4runtime.proto
```diff
@@ -471,6 +471,8 @@ message MulticastGroupEntry {
   // Arbitrary metadata from the controller that is opaque to the target.
   // Added in 1.4.0.
   bytes metadata = 3;
+  // Backup replicas used as a fallback port.
+  repeated Replica backups = 4;
```
antoninbas, line 475: I think nowadays we add a comment for new fields, to indicate which version they were introduced in. There should be examples in the file.

docs/v1/P4Runtime-Spec.mdk
matthewtlam: Fixed up the incorrect change.

proto/p4/v1/p4runtime.proto
matthewtlam: I set the version to 1.4.2. I was not too sure where that comment would go for adding a new proto message.

docs/v1/P4Runtime-Spec.mdk
matthewtlam: I updated the documentation so hopefully it is more concrete, but the idea is that each replica has an associated backup (a list of replicas). So when the primary port of the replica fails, the associated backup would choose one of it's replicas to replace the "primary" port of the old replica

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4550,7 +4566,11 @@ The egress packets may be distinguished for further processing in the egress
 using the `instance` metadata. Note that a packet may not be both unicast and
 multicast; if the multicast group is set, it will override the unicast egress
 port. If the P4 `multicast_group` metadata is set to a value that is not
-programmed in the PRE, then the packet is dropped.
+programmed in the PRE, then the packet is dropped. For each replicas(i) in
+the `multicast_group_entry`, there must be an associated backups(i) or
+`multicast_group_entry` cannot have any backups at all. Fallback ports for
+multicast uses one of the backups(i) replicas when the primary `port` of a
+replicas(i) becomes unavailable.
```
jafingerhut, line 4573: So the intent is that there must be either 0 backups in a multicast_group_entry, or there must be exactly n backups, where n is the number of repliaces in the multicast_group_entry?

Did you consider instead adding the backups message as a sub-message of the Replica message?  Then the association between the primary and backups would be more direct, and you could even make the backups optional on a per-replica basis, e.g. a mix of some replicas that have backups, and some that do not.

jafingerhut, line 4573: Is the intent that when a primary port goes down, that the corresponding backup ports must be tried in the order specified, and the system must use the first one in the list that is up?

Or do you intend that it is explicitly _unspecified_ which backup replica is used by an implementation, as long as the one selected has a port that is up?

proto/p4/v1/p4runtime.proto
```diff
@@ -471,6 +476,9 @@ message MulticastGroupEntry {
   // Arbitrary metadata from the controller that is opaque to the target.
   // Added in 1.4.0.
   bytes metadata = 3;
+  // Backup replicas used as a fallback port.
+  // Added in 1.4.2.
```
jafingerhut, line 480: My guess is that the next release version will likely be 1.5.0.  The only reason 1.4.1 was released was due to a mistake in release 1.4.0 that we wanted to quickly fix.

docs/v1/P4Runtime-Spec.mdk
matthewtlam: The intent is as you mentioned where we either have 0 backups or exactly n backups, where n is the number of replicas in multicast_group_entry.

There were considerations for other options. If we went with something like this where we made the Replica own multiple other Replicas, it provides an nesting behavior that would be undesired.
```
message Replica {
	oneof port_kind {
		uint32 egress_port = 1 [deprecated=true];
		bytes port = 3;
        }
        uint32 instance = 2;
        repeated Replica fallback_replicas = 4;
}
```

Similarly if we chose to create another message FallbackReplicas and keep it nested in Replica, it would create a potential syncing issue between FallbackReplica and Replica if changes were made.
```
message FallbackReplica {
	bytes port = 1;
	uint32 instance = 2;
}

message Replica {
	oneof port_kind {
		uint32 egress_port = 1 [deprecated=true];
		bytes port = 3;
         }
         uint32 instance = 2;
         repeated FallbackReplica fallback_replicas = 4;
}
```

matthewtlam: The intent is that when the primary port goes down, the system will use the first one in the list that is up. In this case order does indeed matter

docs/v1/P4Runtime-Spec.mdk
jonathan-dilorenzo: I think I agree with Antonin here... I would've expected that each replica has a list of backup ports basically, which it goes through in order as ports earlier in the list (and the primary port) fail. Are we sure that this list of fallback replicas can give the semantics we want? @smolkaj as well?

matthewtlam: From a few different perspectives, we want the replicas rather than just purely the fallback ports. 

In vanilla P4, the instance ID allows us to distinguish multiple packet copies in the egress pipeline. Because the P4Runtime standard requires that each (port, instance)-pair within a multicast group is unique, we need the instance ID as non-uniqueness would very likely indicate a bug (why would you want to send two identical packets out of the same port?)

For our use case, having the instance ID allows us to have 2 replicas at the same port but with different SMACs or VLANs

docs/v1/P4Runtime-Spec.mdk
matthewtlam: That's a fair analogy. I got some clarification from the ASIC team.

TLDR; allowing `MulticastGroupEntry` to own the `FallbackReplicas` is a 1:1 ASIC representation and results in fewer ASIC operations. 

However, I feel as though the increased flexibility of having Replicas owning the backups is desired. I've updated my PR accordingly. Will mark this comment as resolved since it is no longer relevant

docs/v1/P4Runtime-Spec.mdk
matthewtlam: As addressed in my other comment and offline discussion, we have replicas own the backup replicas. Closing this comment.

proto/p4/v1/p4runtime.proto
```diff
@@ -447,6 +447,13 @@ message PacketReplicationEngineEntry {
   }
 }
 
+// A backup replica used as a fallback port for a replica.
+message BackupReplica {
+  // Added in v1.5.0.
+	bytes port = 1;
+	uint32 instance = 2;
```
jafingerhut, line 454: Not a big deal, but it looks like there is some tabs vs. whitespace weirdness going on in the definition of BackupReplicate.

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4544,7 +4564,10 @@ four replicas of an ARP packet. These replicas will appear in the egress
 pipeline as independent packets with egress port set to PSA device port numbers
 corresponding to SDN port numbers 5, 12, 18 and 24. For more discussion on the
 translation between SDN ports and PSA device ports, refer to the
-[PSA Metadata Translation](#sec-translation-of-port-numbers) section.
+[PSA Metadata Translation](#sec-translation-of-port-numbers) section. Each
+replica can optionally have a list of backup replicas used as fallback ports
+for multicast. When the primary port (replica) becomes unavailable, the system
+uses the first available backup replica.
```
jafingerhut, line 4570: Suppose one of the replicas for a multicast group is for port 5, instance X, and it has a backup replicate with port 6, instance Y.

Initially port 5 is up, so packets for this replica go to port 5 with instance X.

At some point later port 5 goes down, while port 6 is up.  It seems pretty clear from what you have written already that packets for this replica should now go to port 6, with instance Y.

What if port 6 remains up, and later port 5 comes back up again?  Do you expect/require that packets for this replica go back to port 5, instance X?  Or should the choice of replica only be re-evaluated if the port for where replicas are currently going, goes down?

I suppose a similar scenario applies to watch ports in P4Runtime, and I do not recall if it says explicitly what should happen in this scenario.

Also another minor nit: would it be just as clear to you to replace "becomes unavailable" with "goes down", and "first available backup replica" with "first backup replica for which its port is up"?  The terminology of a port being up or down is used when describing watch port behavior, but I don't recall whether the terms "available" and "unavailable" are used anywhere in the spec currently.

jafingerhut: Also, it might be nice to say explicitly what you expect to happen if the primary and all backup ports are down for a particular replica.

Do you expect no such copy to be made for that replica?

Use the last replica that was being used before it went down, if any of the ports were ever up, or the first one if none of the ports were ever up since the multicast group was last modified?

I realize the answer does not matter if there are no side effects as a result of performing after-multicast-replication processing on the packet before it is dropped (because the port is down), but there often _are_ side effects in after-multicast-replication processing, e.g. counter and meter updates.

matthewtlam: In the first situation that you mentioned Andy, we will go back to the original primary port whenever it comes back up. Will also make those changes as you've requested to specify up/down vs available/unavailable.

In the case where all the primary and all the backup ports are down for a particular replica, nothing is done to recover. We do not expect a copy to be made for that replica.

matthewtlam: I've updated the P4runtime Spec accordingly with the discussion. Please take a look @jafingerhut

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4544,7 +4564,14 @@ four replicas of an ARP packet. These replicas will appear in the egress
...
 translation between SDN ports and PSA device ports, refer to the
-[PSA Metadata Translation](#sec-translation-of-port-numbers) section.
+[PSA Metadata Translation](#sec-translation-of-port-numbers) section. Each
+replica can optionally have a list of backup replicas used as fallback ports
+for multicast. When the primary port (replica) goes down, the system
+uses the first backup replica for which its port is up. Whenever the primary
+port goes back up, the system will use the primary port again. In the case
+when the primary replica's ports and the backup replicas' ports are down,
+nothing is done for recovery until the primary replica or the backup replicas's
+ports go back up.
```
jafingerhut, line 4574: Thanks for those added details.

There are still (at least) two possibilities for what you expect to happen when the primary and all backups are down for a particular replica, which due to packet processing side effects are different from each other.  Do you prefer one behavior over the other?

+ no replica made, and thus no side effects from after-multicast-replication processing should occur for this replica

vs

+ send one replica to the primary port, which will be dropped, but perform any side effects of after-multicast-replication processing

jafingerhut: A third option: Leave it as an implementation-specific behavior whether it does the first thing, or the second thing.  You have no preference.

matthewtlam: What do you think about option 4:

The behavior is the same it was prior to introducing fallback ports

jafingerhut: That's a good one!  I like it.

jafingerhut: Although it might raise the question in the minds of some readers: "What _was_ the behavior prior to introducing fallback ports" ?

matthewtlam: Specifying the behavior prior to how fallback ports was introduce might not be the best place.  I think that it might be up to the P4 architecture to define this behavior

jafingerhut: A probably-common implementation of this feature is one similar to the watch port feature.  For the watch port feature on action selectors, a typical implementation is that some driver software running locally on the switch monitors port up/down status, and edits the action selector contents in hardware as it detects ports going up and down, to remove any group elements that are associated with ports that are down.

For this feature, the similar implementation would be for driver software that is monitoring up/down status to edit multicast replication lists programmed into the hardware, as you specify here.  That is just one possible implementation, of course.

Such an implementation _could_ completely remove any entry for the replica in the hardware state when the primary and all backup ports are down for a particular replica.  It _could_ also leave some replica in the hardware state in this situation.

If you say you are willing to leave it up to the P4 architecture to specify this, then perhaps you do not have any preference between these choices?

matthewtlam: Yes that's right, I believe that we leave it as an implementation-specific behavior then, with no strong preference

jafingerhut: I think it might be nice in the PR to at least mention these two possible behaviors, and recommend that implementors use one of them.

matthewtlam: Updated the documentation to specify the two possible behaviors as suggested. Marking this as resolved.

jafingerhut (approved):
Looks reasonable to me.

matthewtlam:
@chrispsommers Could you please take a look?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4544,7 +4564,21 @@ four replicas of an ARP packet. These replicas will appear in the egress
...
-[PSA Metadata Translation](#sec-translation-of-port-numbers) section.
+[PSA Metadata Translation](#sec-translation-of-port-numbers) section. Each
+replica can optionally have a list of backup replicas used as fallback ports
+for multicast. When the primary port (replica) goes down, the system
+uses the first backup replica for which its port is up. Whenever the primary
+port goes back up, the system will use the primary port again. In the case
+when the primary replica's ports and the backup replicas' ports are down,
+nothing is done for recovery until the primary replica or the backup replicas's
+ports go back up. When the primary and all the backups are down for a
+particular replica, the packet processing side effects can result in one of the
```
chrispsommers, line 4575: What do you mean by side-effects? (I can make some guesses.). It might be helpful to mention an example. I don't think side-effects has been mentioned in the spec before.

rust/src/p4.v1.rs
```diff
@@ -739,12 +739,26 @@ pub mod packet_replication_engine_entry {
         CloneSessionEntry(super::CloneSessionEntry),
     }
 }
+/// A backup replica used as a fallback port for a replica.
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, ::prost::Message)]
+pub struct BackupReplica {
+    /// Added in v1.5.0.
```
chrispsommers, line 746: I think we still need to increment the working draft to 1.5.0. @smolkaj Did you see https://github.com/p4lang/p4runtime/pull/507#issuecomment-2432816297? Would you prefer we do a simple PR to bump the version, before accepting this PR?

chrispsommers (changes requested):
LGTM, but can you provide a brief example of "side-effect?" Also I want to address https://github.com/p4lang/p4runtime/pull/507#issuecomment-2432816297.

docs/v1/P4Runtime-Spec.mdk
jafingerhut: I agree it would be good to be explicit in the spec if it is not clear.  Side effects like counter updates, meter updates, and/or P4 register array writes in the P4 code after multicast replication is done.

matthewtlam: I have specified the packet processing side effects that Andy pointed out. Marking this comment as resolved.

rust/src/p4.v1.rs
matthewtlam: As discussed offline with @smolkaj, Steffen said that he prefers to increment the working draft to 1.5.0

proto/p4/v1/p4runtime.proto
```diff
@@ -457,6 +464,9 @@ message Replica {
     bytes port = 3;
   }
   uint32 instance = 2;
+  // Backup replicas used as a fallback port for the replica.
```
smolkaj, line 467: Nit: "List of backup replicas used as a fallback when the primary replica port goes down, in order of preference."

proto/p4/v1/p4runtime.proto
```diff
@@ -447,6 +447,13 @@ message PacketReplicationEngineEntry {
   }
 }
 
+// A backup replica used as a fallback port for a replica.
+message BackupReplica {
+  // Added in v1.5.0.
```
smolkaj, line 452: Please move the "Added in" comment to the right place, i.e. to the top of the message.

smolkaj (changes requested):
Presumably we want to make support for backup replicas optional? @chrispsommers @jafingerhut?

If so, Matthew, can we call that out explicitly and require targets to either support them or return UNIMPLEMENTED?

jafingerhut:
> @jafingerhut @smolkaj @chrispsommers I was wondering if it was worth specifying if we wish to make it a property that no replica port can be used again in any of it's backup replica ports?

If you consider that an error for a client to try to specify that, and for a server to be required to return an error in that situation, it seems worth specifying the conditions under which the server must return an error, and which error code should be returned.

Another condition that you might want an error for:

(a) among all backup replicas for a single port, the set of (port,instance) values should contain no duplicates
(b) among all replicas and their backups for an entire multicast group, the set of (port,instance) values should contain no duplicates.

(b) is stronger than (a), of course.  If you intend (b), I'd write down (b) in the spec.

smolkaj:
> Presumably we want to make support for backup replicas optional? @chrispsommers @jafingerhut?
> 
> If so, Matthew, can we call that out explicitly and require targets to either support them or return UNIMPLEMENTED?

@chrispsommers @jafingerhut gentle ping in case you want to chime in on this. My recommendation is to make backup replica support optional, but require servers to communicate if they don't support them by returning an UNIMPLEMENTED error if a client tries to program a backup replica. Thoughts?

chrispsommers:
> > Presumably we want to make support for backup replicas optional? @chrispsommers @jafingerhut?
> > If so, Matthew, can we call that out explicitly and require targets to either support them or return UNIMPLEMENTED?
> 
> @chrispsommers @jafingerhut gentle ping in case you want to chime in on this. My recommendation is to make backup replica support optional, but require servers to communicate if they don't support them by returning an UNIMPLEMENTED error if a client tries to program a backup replica. Thoughts?

Agreed, good call!

jafingerhut:
> > > Presumably we want to make support for backup replicas optional? @chrispsommers @jafingerhut?
> > > If so, Matthew, can we call that out explicitly and require targets to either support them or return UNIMPLEMENTED?
> > 
> > 
> > @chrispsommers @jafingerhut gentle ping in case you want to chime in on this. My recommendation is to make backup replica support optional, but require servers to communicate if they don't support them by returning an UNIMPLEMENTED error if a client tries to program a backup replica. Thoughts?
> 
> Making support for this optional on the server side, with an UNIMPLEMENTED return error status if the server does not support it, seems reasonable to me, especially if you expect to encounter devices that do not support the feature. Early on, at least, that seems likely.

A question for you: suppose a client tries to use the feature, and a server not only does not implement it, but is using the current version of the P4Runtime protobuf definitions to parse messages.  Such a server will not even "see" the backup replicas in the data structures it parses from incoming requests, true?  And thus I suspect they would quietly implement everything except the backup replicas functionality, with no error, yes?

As long as a client realizes that, and/or you design some other mechanism for a client to know whether the server supports this feature, seems reasonable to me.

smolkaj:
Yes, agreed, @jafingerhut. 
But I think this is okay and no different from when we add other features in a non-breaking way, no?
In general, a client cannot rely on behavior/a feature added in version `x.y.z` unless it ensures that the server has version `x.y'.z'` with `y' >= y`.

smolkaj:
> > @jafingerhut @smolkaj @chrispsommers I was wondering if it was worth specifying if we wish to make it a property that no replica port can be used again in any of it's backup replica ports?
> 
> If you consider that an error for a client to try to specify that, and for a server to be required to return an error in that situation, it seems worth specifying the conditions under which the server must return an error, and which error code should be returned.
> 
> Another condition that you might want an error for:
> 
> (a) among all backup replicas for a single port, the set of (port,instance) values should contain no duplicates (b) among all replicas and their backups for an entire multicast group, the set of (port,instance) values should contain no duplicates.
> 
> (b) is stronger than (a), of course. If you intend (b), I'd write down (b) in the spec.

Since we're already disallowing duplicate *primary* replicas within a group today, it seems natural to generalize this to *backup* replicas to require uniqueness ignoring the primary/backup distinction.  IIUC, that's property (b) proposed above.

jafingerhut:
> Since we're already disallowing duplicate _primary_ replicas within a group today, it seems natural to generalize this to _backup_ replicas and require uniqueness across _all_ replicas (whether primary or backup) within a group. IIUC, that's property (b) proposed above.

Yes, your restatement of what I wrote as option (b) looks equivalent to me.

matthewtlam:
@smolkaj @jafingerhut @chrispsommers I have updated the P4runtime Spec with the ideas we've discussed on support for backup replicas being optional. As well as the properties of replicas and backup replicas and the associated error messages for breaking the properties

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4544,7 +4562,23 @@ four replicas of an ARP packet. These replicas will appear in the egress
 pipeline as independent packets with egress port set to PSA device port numbers
 corresponding to SDN port numbers 5, 12, 18 and 24. For more discussion on the
 translation between SDN ports and PSA device ports, refer to the
-[PSA Metadata Translation](#sec-translation-of-port-numbers) section.
+[PSA Metadata Translation](#sec-translation-of-port-numbers) section. Each
+replica can optionally have a list of backup replicas used as fallback ports
+for multicast. When the primary port (replica) goes down, the system uses the
+first backup replica whose port is up. Whenever the primary port goes back up,
```
smolkaj, line 4568: "Whenever the primary port goes back up, the system will use the primary port again."

A similar thing happens when we're currently using back up 2, but then back up 1's port comes back up, right? Can we generalize this a bit to capture that as well?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4544,7 +4562,23 @@ four replicas of an ARP packet. These replicas will appear in the egress
...
+the system will use the primary port again. In the case when the primary
+replica's ports and the backup replicas' ports are down, nothing is done for
+recovery until the primary replica's or backup replicas' ports go back up.
+When the primary and all the backups are down for a particular replica, the
+packet processing side effects can result in one of the two main expected
```
smolkaj, line 4573: "two main expected behaviors that the target is recommended to implement."

Nit: Can we make it clearer to explain that this is a forward reference to what you are about to say?
For example, wow about ending this sentence in a colon and then using a numbered list for the two options?

docs/v1/P4Runtime-Spec.mdk
```diff
@@ -4577,6 +4613,8 @@ semantics.
   metadata in the data plane set to the deleted `multicast_group_id` will be
   dropped.
 
+The support for back replicas is optional by the target and if the target does
```
smolkaj, line 4616: Please move this under INSERT, so we have the full specification will all error codes in one place.

proto/p4/v1/p4runtime.proto
```diff
@@ -457,6 +464,10 @@ message Replica {
     bytes port = 3;
   }
   uint32 instance = 2;
+  // List of backup replicas used as a fallback when the primary replica port
```
smolkaj, line 467: "when the primary replica port goes down" -> "when the primary replica port (and all higher-preference backup replica ports) are down"

smolkaj (approved):
Final nits.
Approving early.

matthewtlam:
I've added the changes to the final nits that Steffen requested

matthewtlam:
@chrispsommers was wondering if you could please take another look?

chrispsommers (approved):

smolkaj:
Approved at the working group meeting.

PR #512, opened 2024-12-12
Unimportant whitespace change in README to see if CI tests pass

jafingerhut:


smolkaj:
I suspect the regression could be due to https://github.com/actions/runner-images/issues/10636

jafingerhut:
> I suspect the regression could be due to [actions/runner-images#10636](https://github.com/actions/runner-images/issues/10636)

If that is the cause, perhaps we should consider making a quick change that causes that CI job to use whatever version of Ubuntu it was before (maybe 22.04?), until a longer-term fix can be devised.

PR #513, opened 2024-12-12
Temporarily pin Ubuntu CI version to 22.04, since 24.04 causes regression. 

smolkaj:
@jafingerhut  https://github.com/p4lang/p4runtime/pull/512#issuecomment-2537572636


jafingerhut (approved):
LGTM

PR #521, opened 2024-12-18
CI: Test on several Bazel and Ubuntu versions.

smolkaj:
Also makes some improvements to the CI script taken from https://github.com/google/netkat.
They will allow us to make better use of the cache while avoiding hitting GitHub's cache limits.


.github/workflows/ci-build-proto.yml
```diff
@@ -14,30 +14,92 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-22.04
+
+    strategy:
+      matrix:
+        # We only test on the oldest version we want to support and latest.
+        # We trust that things also work for versions in the middle.
```
jonathan-dilorenzo, line 21: That's brave ;)

smolkaj: A lot less brave than what we have today, which is to only test a single version.

smolkaj:
FYI: Since this is changing the name of some checks, we need to bypass branch protections to submit and/or update our branch protections rules (probably afterwards).

chrispsommers (approved):
LGTM

PR #532, opened 2025-01-19
Whitespace-only README test change to cause CI to run

jafingerhut:


jafingerhut:
Even though madoko-lint is mentioned _nowhere_ in any file of this repository that I can find, it is still in the list of tests to pass on the Github page of this PR.  I do not understand why, but it would be nice to change it.

jafingerhut:
One of the CI tests is "Build spec / asciidoc-lint (pull_request)", so that is new with a recent PR to switch to AsciiDoc.

Could it be that `madoko-lint` comes from some configuration via Github web page UI, not from any file in the `.github` directory files?

PR #510, opened 2024-11-16
Switching from Madoko to AsciiDoc for P4Runtime specification

Dscano:
I am currently in the process of transitioning the P4Runtime specification from Madoko to AsciiDoc. Below is an overview of the remaining tasks:

- I need to fix the layout of two tables in the P4Runtime-spec.adoc file.
- Finalize the reorganization and changes necessary for folder structure to support the switch from Madoko to AsciiDoc.

PR #539, opened 2025-01-26
Fixed CI configuration: adapted folder structure

Dscano:
Moved the folder from docs/v1/build to docs/v1 and separately copied the generated files.
@chrispsommers

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 # Emacs
 *~
-docs/v1/
 .DS_Store
 build
 dist
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 # Emacs
 *~
-docs/v1/build/
+docs/v1/
 .DS_Store
 build
 dist
--- a/.github/workflows/any-branch-uploads.yml
+++ b/.github/workflows/any-branch-uploads.yml
@@ -15,7 +15,8 @@ jobs:
       run: |
         make -C docs/tools/
         docker run -v `pwd`/docs/v1:/usr/src/p4-spec p4lang/p4rt-asciidoc make
-        ls docs/v1/build
+        ls docs/v1/P4Runtime-Spec.pdf
+        ls docs/v1/P4Runtime-Spec.html
     - name: Upload spec to S3 if needed
       if: ${{ github.actor != 'dependabot[bot]' }}
       uses: jakejarvis/s3-sync-action@v0.5.1
--- a/.github/workflows/any-branch-uploads.yml
+++ b/.github/workflows/any-branch-uploads.yml
@@ -27,5 +27,5 @@ jobs:
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         AWS_REGION: 'us-west-2'
-        SOURCE_DIR: 'docs/v1/build'
+        SOURCE_DIR: 'docs/v1'
         DEST_DIR: ci/${{ github.ref_name }} # only runs for push events
--- a/.github/workflows/main-branch-uploads.yml
+++ b/.github/workflows/main-branch-uploads.yml
@@ -29,14 +29,15 @@ jobs:
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         AWS_REGION: 'us-west-2'
-        SOURCE_DIR: 'docs/v1/build'
+        SOURCE_DIR: 'docs/v1'
         DEST_DIR: docs/main
     - name: Prepare spec for upload to gh-pages
       run: |
         git checkout gh-pages
         mkdir -p spec
         rm -rf spec/main
-        cp -r docs/v1/build spec/main
+        cp docs/v1/P4Runtime-Spec.html spec/main
+        cp docs/v1/P4Runtime-Spec.pdf spec/main
     - name: Upload spec to gh-pages
       uses: EndBug/add-and-commit@v9
       with:
--- a/.github/workflows/main-branch-uploads.yml
+++ b/.github/workflows/main-branch-uploads.yml
@@ -18,7 +18,8 @@ jobs:
       run: |
         make -C docs/tools/
         docker run -v `pwd`/docs/v1:/usr/src/p4-spec p4lang/p4rt-asciidoc make
-        ls docs/v1/build
+        ls docs/v1/P4Runtime-Spec.pdf
+        ls docs/v1/P4Runtime-Spec.html
     - name: Upload spec to S3
       uses: jakejarvis/s3-sync-action@v0.5.1
       with:
--- a/.github/workflows/tag-uploads.yml
+++ b/.github/workflows/tag-uploads.yml
@@ -31,14 +31,15 @@ jobs:
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
         AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         AWS_REGION: 'us-west-2'
-        SOURCE_DIR: 'docs/v1/build'
+        SOURCE_DIR: 'docs/v1/'
         DEST_DIR: docs/${{ github.ref_name }}
     - name: Prepare spec for upload to gh-pages
       run: |
         git checkout gh-pages
         mkdir -p spec
         rm -rf spec/$TAG
-        cp -r docs/v1/build spec/$TAG
+        cp docs/v1/P4Runtime-Spec.html spec/$TAG
+        cp docs/v1/P4Runtime-Spec.pdf spec/$TAG
     - name: Upload spec to gh-pages
       uses: EndBug/add-and-commit@v9
       with:
--- a/.github/workflows/tag-uploads.yml
+++ b/.github/workflows/tag-uploads.yml
@@ -20,7 +20,8 @@ jobs:
       run: |
         make -C docs/tools/
         docker run -v `pwd`/docs/v1:/usr/src/p4-spec p4lang/p4rt-asciidoc make
-        ls docs/v1/build
+        ls docs/v1/P4Runtime-Spec.pdf
+        ls docs/v1/P4Runtime-Spec.html
     - name: Upload spec to S3
       uses: jakejarvis/s3-sync-action@v0.5.1
       with:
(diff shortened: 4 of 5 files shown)


.gitignore
```diff
@@ -1,6 +1,6 @@
 # Emacs
 *~
-docs/v1/build/
+docs/v1/
```
jafingerhut, line 3: I don't think we want docs/v1/ directory in .gitignore, do we?  If I understand the git behavior correctly, making this change will cause _no_ files in docs/v1 or its subdirectories to show up in the output of commands like `git status`, and I think it might also make it impossible to `git commit` files in that directory, unless you first modify the `.gitignore` file.

Dscano: Sorry, you're right

jafingerhut (approved):
LGTM

chrispsommers:
@jafingerhut @Dscano The build finally went through, until it encountered upload auth issues. Does anybody know the solution?

Dscano:
> @jafingerhut @Dscano The build finally went through, until it encountered upload auth issues. Does anybody know the solution?

I compared the logs from [main-branch-uploads-success](https://github.com/p4lang/p4runtime/actions/runs/12288434986/job/34292223762#logs) and [main-branch-uploads-fails](https://github.com/p4lang/p4runtime/actions/runs/12982970840/job/36203537920#step:5:1). It seems that in the main-branch-uploads-fails run, the deletion of old files is performed, but we are unable to do this due to permission issues. These permissions are related to the AWS_S3 user that runs the command.

Based on the arguments and this [guide](https://github.com/jakejarvis/s3-sync-action/blob/master/README.md), It should be fine, but further investigation is needed.

jafingerhut:
I don't know the solution, but perhaps this CI task does not have access to the secret credentials required to update that S3 storage account?

chrispsommers:
@antoninbas Any ideas here? Thanks.

chrispsommers:
@fruffy we're having S3 access issues, do you have any advice?

antoninbas:
I have added the missing permission so it should work now.

Also, please modify the job definition so that only the required files are uploaded (build outputs), not all the source files. cc @Dscano 

On a side note, this is my personal AWS account. I am happy to keep it going the way it is, but feel free to switch to a p4.org account if such a thing exists.

Finally, it seems that the build outputs are no longer "copied" to the gh-pages branch, which means the official p4.org links are no longer up-to-date.

fruffy:
Can check this over the weekend if it isn't resolved by then.

chrispsommers:
Thanks @antoninbas. @fruffy we should probably have p4.org account for this and not depend upon personal accounts, although Antonin has graciously extended his use of his.

chrispsommers:
@Dscano Could you please take a look at @antoninbas's comments? Thanks!

Dscano:
Hi @chrispsommers, I’ve addressed Antoinnin’s comment in this PR https://github.com/p4lang/p4runtime/pull/541#issue-2824393055

fruffy:
@AdarshRawat1 You did a great job with automated documentation and CI for P4C. We have similar problem here with P4Runtime. Do you think you could set up a similar structure for p4runtime? We do not need Doxygen etc, html and pdf previews for a branch are enough. If you are interested we can discuss with @chrispsommers, @Dscano, and @jafingerhut. I do not think it will be a big time investment.  

@antoninbas Trying to understand the current setup. Why do we need AWS, to upload the html and PDF files. Is github pages not enough?

AdarshRawat1:
> @AdarshRawat1 You did a great job with automated documentation and CI for P4C. We have similar problem here with P4Runtime. Do you think you could set up a similar structure for p4runtime? 

I'll be happy to help! I'm looking into the current setup and will update you soon.

> @antoninbas Trying to understand the current setup. Why do we need AWS, to upload the html and PDF files. Is github pages not enough?

+1 on this.

AdarshRawat1:
> Finally, it seems that the build outputs are no longer "copied" to the gh-pages branch, which means the official p4.org links are no longer up-to-date.

@antoninbas  Could you help me understand why the [latest GitHub Pages deployment ](https://p4lang.github.io/p4runtime/)is pointing to this?

image

Even the [June 7, 2024 deployment](http://p4.org/p4runtime/) leads to an empty page. Has it been like this from the start, or did something change recently?
image

AdarshRawat1:
> Do you think you could set up a similar structure for p4runtime? We do not need Doxygen etc, html and pdf previews for a branch are enough. 

Based on [the docs](https://github.com/p4lang/p4runtime/tree/main/docs/v1#markup-version), I'm assuming that we'll be working with [AsciiDoc](https://docs.asciidoctor.org/) for rendering the output on GitHub Pages. Is this what you had in mind?

fruffy:
> > Do you think you could set up a similar structure for p4runtime? We do not need Doxygen etc, html and pdf previews for a branch are enough.
> 
> Based on [the docs](https://github.com/p4lang/p4runtime/tree/main/docs/v1#markup-version), I'm assuming that we'll be working with [AsciiDoc](https://docs.asciidoctor.org/) for rendering the output on GitHub Pages. Is this what you have in mind?

Davide set this up in https://github.com/p4lang/p4runtime/pull/510. Afaik this is already working but I have not kept up with the developments in this repository.

Dscano:
> > Do you think you could set up a similar structure for p4runtime? We do not need Doxygen etc, html and pdf previews for a branch are enough.
> 
> Based on [the docs](https://github.com/p4lang/p4runtime/tree/main/docs/v1#markup-version), I'm assuming that we'll be working with [AsciiDoc](https://docs.asciidoctor.org/) for rendering the output on GitHub Pages. Is this what you have in mind?

Yes, so basically, AsciiDoc is the text processor used to generate the `.pdf `and `.html` files. You can generate these files via the  `make` command within the `p4runtime/docs/v1/`.

AdarshRawat1:
@Dscano @fruffy @jafingerhut @chrispsommers @antoninbas Are you looking for something [like this?](https://adarshrawat1.github.io/p4runtime-doc-testing/)

You may review the [Workflow file here.](https://github.com/AdarshRawat1/p4runtime-doc-testing/blob/main/.github/workflows/deploy-docs.yml) & the demo [deployment Repository here](https://github.com/AdarshRawat1/p4runtime-doc-testing).

Suggestion: We could bypass Amazon S3 to avoid billing for @antoninbas  and instead use GitHub Pages for domain redirection. Let me know if I’m overlooking anything—would love to hear your thoughts!

jafingerhut:
@AdarshRawat1 I do not know the details, but I believe there are Github actions for this repository: https://github.com/p4lang/p4-spec

such that on every PR merged into the main branch, the PDF and HTML of the P4-16-spec.adoc AsciiDoc source file are generated, and they are put into a place, I think into a file in the gh-pages branch of the repo, where they are soon visible via links on this page: https://p4.org/specs/

In quickly looking at the .github/workflows files in the p4-spec repo, I notice it mentions S3 there as well, but I have not tried to follow its full behavior to understand it.  From what I explain below, it is possible that the S3 version copied by the p4-spec scripts is not used by the p4.org web site.

On this page https://p4.org/specs/ there are 2 links for "working draft" versions of the P4_16 language specification.  The HTML URL link on that page is to https://p4.org/p4-spec/docs/p4-16-working-draft.html

If you do `wget https://p4.org/p4-spec/docs/p4-16-working-draft.html` on a system with `wget` installed, I get this HTML text:
```
<!DOCTYPE html>
<html>
<head>
<script>
const url = "https://raw.githubusercontent.com/p4lang/p4-spec/gh-pages/docs/P4-16-working-spec.html";  
const req = new XMLHttpRequest();
req.open("GET", url, true);
req.onload = (_) => {
    const doc = document.open("text/html", "replace");
    doc.write( req.response );
    doc.close();
};
req.send(null)
</script>
</head>
<body></body>
</html>
```
I do not know JavaScript, but Nate Foster wrote the above, and the basic idea is to get the HTML from the URL https://raw.githubusercontent.com/p4lang/p4-spec/gh-pages/docs/P4-16-working-spec.html and then load that into the browser and display that.  This was a short quick hack to display the HTML at that other URL into the user's browser, because simply linking directly to https://raw.githubusercontent.com/p4lang/p4-spec/gh-pages/docs/P4-16-working-spec.html shows raw HTML, not rendered HTML.

There is a similar JavaScript script for the language spec PDF link.

And also similar ones for the PSA specification, PNA specification, and P4Runtime specification.  The other specifications on that page have not been updated in a while, so do not use this technique for their working draft links, if they have them, which is fine -- they don't need to be updated.

What would be nice for all of the active specifications is to have a similar workflow, such that on a merge of a PR to their main/master branch, causes the PDF and HTML to be generated, and the latest version to be stored in some place that is visible from the p4.org specifications page.  This has been working for over a year now, but perhaps with different Gitub workflow scripts for different P4 specification repositories.  It would be nice if they could be made more similar to each other.

One thing that was an issue for over a year in the p4-spec repo is that the way it committed new versions of generated PDF and HTML to its repo was that it kept a full history of all PDF and HTML versions, growing the repository history and size by a _lot_ on every merge.  That might be improved now, I am not sure.  I believe that the p4runtime repo used different options on git commands committing to its gh-pages branch, maybe the `-f` option?, that caused it _not_ to grow by a lot on every new PDF and HTML version.

antoninbas:
Davide was able to fix the CI by restoring the previous folder structure. So now the gh-pages branch is updated as expected and the latest ("draft") version of the spec is accessible through p4.org.

I don't know if any additional change is required wrt gh-pages now that this has been addressed.

On the workflow that pushes build artifacts to an AWS S3 bucket: this was the legacy way of uploading the spec, until we switch to gh-pages. We kept it around because it would upload the build outputs (.html and .pdf files) not just for the main branch and for tags, but also for any branch in this repo. This was a convenient way to review spec changes for frequent contributors who have write access to this repo and can therefore open PRs from branches in this repo instead of from a personal fork. When review the PR, reviewers can access a "well-known" link (which includes the branch name) to make sure there are not obvious formatting issues. This is still documented here: https://github.com/p4lang/p4runtime/tree/main/docs#ci-upload-of-built-documents. IIRC, there is a rule in the AWS bucket that delete the spec documents for the development branch after a couple of months. I will let the current maintainers / WG chairs decide what they want to do:
1) keep things as they are
2) upload the build outputs for dev branches to gh-pages instead of AWS S3 - I don't think there is any technical issue with this, it's just a bit more work to set up some job to delete stale documents (assuming one doesn't want things to accumulate in gh-pages)
3) **upload the build outputs as workflow artifacts** - this is the best solution because it should also work for PRs from forks, and artifacts can be deleted automatically after 30 or 90 days
4) remove that capabilty altogether

The 3rd solution is the best one IMO. Someone could take a look and check that the uploaded artifacts -  and in particular the .html file - are easy to view.

chrispsommers:
@antoninbas Thanks for the refresher. I too like your proposed 3rd approach. We should harmonize with other P4 WG's if possible. I'm not sure what effort is needed to achieve this - would the P4 LDWG need to make corresponding changes etc.

fruffy:
@AdarshRawat1 has set up something like this for P4C's documentation. We could have a similar approach here. Example: 
https://github.com/p4lang/p4c/pull/5113#issuecomment-2615831349

We can move the discussion in a separate issue to track this.

PR #546, opened 2025-02-11
Update actions/cache@v2, which is soon to be retired by Github actions

jafingerhut:


jafingerhut:
According to a log message in a failing test for this recently created PR that modifies one of the `.proto` files, the Github action actions/cache@v2 used by `ci-build-proto.yml` is soon to be retired (2025-Mar-01), and is being failed intermittently before then as a warning to repos still using it:
```
Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v2`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down
```

chrispsommers (approved):
LGTM, let's have @smolkaj take a look.

smolkaj (approved):
Thanks!

PR #548, opened 2025-02-27
Specify read-write symmetry exception for backup replicas in P4Runtime Spec

matthewtlam:


chrispsommers (approved):
LGtM

matthewtlam:
Screenshot 2025-03-04 at 11 31 49 AM

@jafingerhut @chrispsommers I was wondering if you knew what these checks were and why they are waiting for the status to be reported?

chrispsommers:
@matthewtlam I'm not sure, perhaps because uploads are not done on PRs, only merges to main? It looks like we maintainers can bypass the warning but I'd rather understand this first. @jafingerhut or @smolkaj any ideas? 
image

jafingerhut:
Those names `main-branch-uploads` and `any-branch-uploads` definitely correspond to names of workflows in this directory: https://github.com/p4lang/p4runtime/tree/main/.github/workflows

I can tell from the Settings on this Github repo that they are listed as required things for the main branch.

I do not know the history of those CI tasks, or why they are required.

jafingerhut:
From git commit log, main-branch-workflows.yml was created in 2022 by Antonin Bas, when this repo transitioned from using Travis CI to using Github Actions, then it was tweaked slightly in 2025 by Davide Scano related to the Madoko -> AsciiDoc transition.

jafingerhut:
Given that an inconsequential change to the spec and creating a PR from it has the same effect (see https://github.com/p4lang/p4runtime/pull/549), that is strong evidence that the hung tests have nothing to do with this PR's changes.

I am not sure, but perhaps someone marked these 2 CI tests as requirements for PRs, but they should not have?  I do not know of any commit log for Github settings, unfortunately, so I have no easy way to check this other than asking folks who might have changed it.

@antoninbas @fruffy @Dscano Do any of you know if these 2 CI tests that are "Waiting for status to be reported", perhaps forever, should be changed in Github settings so they are not required for a PR to be merged?

smolkaj:
I did have to update the "required checks" as part of https://github.com/p4lang/p4runtime/pull/521, since that changed the name of the bazel CI runs. I did not intend to make changes to the upload checks, but I wouldn't rule out that I did make such changes accidentally.

smolkaj:
+1 on figuring out if the "upload" CI checks are intended to be required or not.

fruffy:
> +1 on figuring out if the "upload" CI checks are intended to be required or not.

You should be able to set the required CI checks in Settings->Branches->Main->Required Status Checks. The available checks are determined by recent PRs that run. Outdated checks have to be removed/replaced.

jafingerhut:
> > +1 on figuring out if the "upload" CI checks are intended to be required or not.
> 
> You should be able to set the required CI checks in Settings->Branches->Main->Required Status Checks. The available checks are determined by recent PRs that run. Outdated checks have to be removed/replaced.
> 
> Edit: Oh I see the issue. These workflows are not being triggered properly. It is possible that the trigger setting [here](https://github.com/p4lang/p4runtime/blob/main/.github/workflows/any-branch-uploads.yml#L3) is not activated because @matthewtlam is pushing to a branch on a fork. What is missing is the pull_request triggered like [here](https://github.com/p4lang/p4runtime/blob/main/.github/workflows/codegen.yml#L3). But this is just guesswork.

Pushing to a branch on a fork is a common workflow on Github, one that we should consider normal, I think.

Pushing to a branch in the repo where you want to create the PR is also fairly common, but less common, since you have to have permission to create branches in the repo, which we don't give everyone permission to do.

In short, @chrispsommers, if you think this PR's changes are good, I'd say hit the check box to override the warning and merge it, if you approve and are ready to merge it.  This kerfuffle with weird required CI tasks that should not be blocking the issue are a temporary solvable thing that we will solve, but probably not today.  And given that I have already reproduced this particular behavior on another simple PR that was also created from a branch on a fork, we know how to reproduce this scenario pretty easily, and do not need this particular PR to wait in order to diagnose the issue.

chrispsommers:
@jafingerhut @fruffy @smolkaj thanks for weighing in. Andy I agree. I am inclined to merge. Any objections?

PR #556, opened 2025-04-29
Add device_id field to capabilites request

verios-google:
Pretty self-explanatory. A controller can speak to more than one switch, best to know which switch the request is for.


proto/p4/v1/p4runtime.proto
```diff
@@ -834,7 +834,10 @@ enum SdnPort {
 }
 
 //------------------------------------------------------------------------------
-message CapabilitiesRequest {}
+message CapabilitiesRequest {
+  // Added in v1.5.0
+  uint64 device_id = 1;
```
antoninbas, line 839: Is there a way to have this be optional?
The `p4runtime_api_version` would typically be the same for all devices controlled by a P4Runtime server, so if a client just wants to query that information, now it has to provide a `device_id` for no reason.
The notion of capabilities is (intentionally?) pretty broad, but I expect some would be device-independent, and some may be device-specific. Could `device_id` be defined as an optional filter to limit the capabilities in the response to the ones applicable to that specific device? If no `device_id` is provided, the response should only include capabilities which are applicable across all devices, or should include a mapping from device ID to device-specific capabilities in the `experimental` field.

jonathan-dilorenzo: That seems like a reasonable approach to me. Let's say "If unset (or 0), should return only capabilities common to all devices controlled by the server. Otherwise, should return any capabilities applicable to the given device." or something like that?

proto/p4/v1/p4runtime.proto
```diff
@@ -834,7 +834,12 @@ enum SdnPort {
 }
 
 //------------------------------------------------------------------------------
-message CapabilitiesRequest {}
+message CapabilitiesRequest {
+  // Optional: Response should include the device-specific capabilities of the
+  // device identified by `device_id` in addition to server-wide capabilities.
```
LimeHat, line 839: I believe `CapabilitiesResponse` message should be updated accordingly as well

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -6032,10 +6032,14 @@ error {
...
 
-The `CapabilitiesRequest` message contains a single field `device_id`.
-
-The `device_id` uniquely identifies the target P4 device. A `NOT_FOUND` error is
-returned if the `device_id` is not recognized by the P4Runtime server.
+The `CapabilitiesRequest` message contains a single optional field `device_id`.
+
+`device_id` is an optional filter to limit the capabilities in the response to those
+applicable to the target P4 device. A `NOT_FOUND` error is returned if the `device_id`
+is not recognized by the P4Runtime server. If `device_id` is unset (or 0), the response
```
chrispsommers, line 6039: This implies to me that a real device cannot have an id of 0? Excuse me if this is well-known.

verios-google: Well darn. That's a good observation. I don't believe there's anything outlawing `device_id = 0`.

This means I would have to add the `optional` keyword to the field

This would make migration to proto `editions` a little messier. Not necessarily complicated in anyway just a few more toggles to make sure semantics are identical (or we don't care about exact semantic matching)

chrispsommers: I don't think that adding `optional` changes the semantics of the message. The reader still cannot tell in proto3 if the field was set to 0 or unset, correct? The way this is sometimes handled is to wrap the field in a xxValue message, e.g. [Uint32Value](https://protobuf.dev/reference/java/api-docs/com/google/protobuf/UInt32Value.html)

verios-google: IIRC, adding the `optional` keyword to field X introduces the function `has_X()`

Messages have this function by default (which is why wrapping works), but enums, numericals, and strings only gain this function if the optional keyword is used. https://protobuf.dev/programming-guides/field_presence/#presence-in-proto3-apis

jonathan-dilorenzo: @chrispsommers, not clear if it's well known, but indeed implied by the specification. See https://p4.org/p4-spec/docs/p4runtime-spec-working-draft-html-version.html#_set_unset_scalar_fields. That said, this could perhaps be made clearer for each of the fields where you're not allowed to use 0... But I guess it's nice to not have to mention it every time.

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -6030,10 +6030,19 @@ error {
 == `Capabilities` RPC
 
 The `Capabilities` RPC offers a mechanism through which a P4Runtime client can
-discover the capabilities of the P4Runtime server implementation. At the moment,
-the `CapabilitiesRequest` message is empty and the `CapabilitiesResponse`
-message includes the `p4runtime_api_version` string field and the `experimental`
-field, which has type `Any` Protobuf message cite:[ProtoAny].
+discover the capabilities of the P4Runtime server implementation.
```
jonathan-dilorenzo, line 6033: "... and the devices the server controls." ?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -6030,10 +6030,19 @@ error {
...
+
+The `CapabilitiesRequest` message contains a single optional field `device_id`.
+
+`device_id` is an optional filter to limit the capabilities in the response to those
+applicable to the target P4 device. A `NOT_FOUND` error is returned if the `device_id`
+is not recognized by the P4Runtime server. If `device_id` is unset (or 0), the response
```
jonathan-dilorenzo, line 6039: Unset and 0 are also the same, so maybe say "(i.e. 0)" instead just to be clearer?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -6030,10 +6030,19 @@ error {
...
+should only include capabilities that are server-wide (those which are applicable
+across all devices). Note that if `device_id` is set to a valid id, both server-wide and
+device-specific capabilities will be returned.
+
+The `CapabilitiesResponse` message includes the `p4runtime_api_version` string field
+and the `experimental` field, which has type `Any` Protobuf message cite:[ProtoAny].
 
 The `p4runtime_api_version` field must be set to the full semantic version string cite:[SemVer]
```
jonathan-dilorenzo, line 6047: Worth specifying that this is server-wide? Or e.g. "On success, this should be returned regardless of device ID."?

I'm not married to this though btw.

docs/v1/P4Runtime-Spec.adoc
chrispsommers: OK, I see. Somehow I thought `optional` was _ignored_ in proto3. I've been toiling in ignorance for years.

That being said, per [8.1.1. Set / Unset Scalar Fields](https://p4.org/p4-spec/docs/p4runtime-spec-working-draft-html-version.html#_set_unset_scalar_fields):
> In Protobuf version 3 (proto3), the default value of scalar fields is 0 for numeric types such as int32, and the empty string "" for string types (string and bytes). An application, such as the P4Runtime client or server, is unable to distinguish between an unset scalar field and a scalar field set to its default value. Therefore, we usually reserve the default values 0 and "" of scalar fields to mean "unset".

I wonder why we didn't extend this topic and articulate that one should add `optional` to a scalar field to allow the reader to distinguish an unset value from a default/0 value?

It seems like in the past we did some backflips dealing with the limitations of default/unset fields lacking `optional`, and I wonder why we didn't solve it by adding `optional` (perhaps due to backwards-compatibility)?

jonathan-dilorenzo: So, somehow, I'm recalling that Optional is strongly discouraged (maybe because it's forgotten so often, so you should just not use the default value for scalars anyway) and that's why we avoid using it. However, brief googling suggests that we externally recommend always using it: https://protobuf.dev/programming-guides/proto3/#field-labels

So no idea what's going on there :'(. 

Proto3 also didn't have optional for some time, so that may have been why we didn't use it initially at least. That said, I think disallowing default values to be used to refer to specific objects is generally a good idea, so not sure we suffered much from not doing this ;).

chrispsommers: >Proto3 also didn't have optional for some time

I didn't get the memo when it was added, ROTFL, which probalby explains my confusion. Thanks!

docs/v1/P4Runtime-Spec.adoc
verios-google: Done

docs/v1/P4Runtime-Spec.adoc
verios-google: Done

docs/v1/P4Runtime-Spec.adoc
verios-google: Done

docs/v1/P4Runtime-Spec.adoc
verios-google: I updated the documentation to say "unset (i.e. 0)" to specify that they are one in the same. Good enough?

verios-google:
@chrispsommers and @jonathan-dilorenzo, any other open concerns for the PR, otherwise can I get approval and merger

jonathan-dilorenzo (approved):
LGTM!

jonathan-dilorenzo:
So, if we think there's anything to discuss here, @antoninbas and @chrispsommers, then we could chat in meeting on Friday. O.w. I'd be comfortable with this getting merged as it looks now (or with whatever additional comments you have)?

docs/v1/P4Runtime-Spec.adoc
chrispsommers: So, are we OK with the implication that a concrete `device_id` of zero is unsupported, or at least we can't filter for such a `device_id`? I guess we can accept this limitation since this is experimental and will be deprecated.

jonathan-dilorenzo: Yes, I think we are both OK with that and that it is explicitly unsupported by the P4Runtime spec already. That said, I would also support an extra clarification (in a different PR), maybe in https://p4.org/p4-spec/docs/p4runtime-spec-working-draft-html-version.html#sec-client-arbitration-and-controller-replication stating explicitly that device ID 0 is an invalid argument... Though I also see that it's technically redundant with the rest of the spec...

chrispsommers (approved):
LGTM

chrispsommers:
> So, if we think there's anything to discuss here, @antoninbas and @chrispsommers, then we could chat in meeting on Friday. O.w. I'd be comfortable with this getting merged as it looks now (or with whatever additional comments you have)?

I'm OK with as-is.

antoninbas (approved):
LGTM, thanks for making the suggested change

I'd recommend making it explicit that 0 is not a valid device_id in a follow-up change. When I looked at the spec the other day, I couldn't find any mention of this, even though we do have a general statement that the zero value for scalar fields is usually not valid.

verios-google:
Sounds good, I'll take a stab at it when I get the time. @chrispsommers can you merge this PR

chrispsommers:
> Sounds good, I'll take a stab at it when I get the time. @chrispsommers can you merge this PR

If you're going to add a sentence about device_id 0 isn't valid, how about do that before we merge? It sounds trivial and will avoid yet another PR. We can merge in the meeting today, one way or the other.

PR #557, opened 2025-05-09
Clarify in spec that device id can never be 0

verios-google:


verios-google:
@chrispsommers, @antoninbas, and @jonathan-dilorenzo follow-up PR as requested

chrispsommers (approved):
LGTM, thanks.

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -601,14 +601,19 @@ their write requests. gRPC provides authentication methods cite:[gRPCAuth] that
...
-=== Default Role
+=== Default Role and Default Device ID
 
 A controller can omit the role message in `MasterArbitrationUpdate`. This
 implies the "default role", which corresponds to "full pipeline access".
 This also implies that a default role has a `role_id` of `""` (default).
 If using a default role, all RPCs from the controller (e.g. `Write`) must
 leave the `role` unset.
 
+The "default device_id" corresponds to `0`. As mentioned [previously](#sec-default-valued-fields),
```
jonathan-dilorenzo, line 612: nit: This is technically mentioned in a later section. Maybe just "As per section X, ..."?

jonathan-dilorenzo: But skip this and instead do the thing below imo.

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -601,14 +601,19 @@ their write requests. gRPC provides authentication methods cite:[gRPCAuth] that
...
+an application is unable to distinguish between unset vs default valued scalar
+fields. Some requests (e.g. [Cabilities](#sec-capabilities-rpc)) uniquely handle an unset `device_id`.
```
jonathan-dilorenzo, line 614: Instead of "As mentioned [previously](#sec-default-valued-fields),
an application is unable to distinguish between unset vs default valued scalar
fields.  Some requests (e.g. [Cabilities](#sec-capabilities-rpc)) uniquely handle an unset `device_id`.
Therefore, a device must NEVER be assigned a `device_id` of 0."

How about: "As per section X, default values are invalid for all scalar P4 object IDs. This includes `device_id`, so the `0` ID may not be assigned to a device. "

jonathan-dilorenzo (changes requested):

docs/v1/P4Runtime-Spec.adoc
verios-google: You know, at this point I'm more of a tool then a person :)

jafingerhut: For good or ill, lots of changes to P4 specification PRs are like this.  I woudn't be surprised if the same were true for other specs maintained by groups.

jonathan-dilorenzo: To be clear, the X was supposed to be filled in using a macro ;).

Would remove the last 2 paragraphs. Maybe you can add "Therefore, the default device ID (which is indistinguishable from 0) may never be used to refer to a specific device." ?

jonathan-dilorenzo (approved):
Looks great! Thanks!

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -601,14 +601,18 @@ their write requests. gRPC provides authentication methods cite:[gRPCAuth] that
...
 
 A controller can omit the role message in `MasterArbitrationUpdate`. This
 implies the "default role", which corresponds to "full pipeline access".
 This also implies that a default role has a `role_id` of `""` (default).
 If using a default role, all RPCs from the controller (e.g. `Write`) must
 leave the `role` unset.
 
+As per section [Default-Valued Fields](#sec-default-valued-fields), default values are invalid for all
+scalar P4 object IDs. This includes `device_id`, so the `0` ID must not be
+assigned to a device.
```
smolkaj: Optional nit: maybe we can make this a bit more concise and precise by replacing: ", so the `0` ID must not be assigned to a device." -> "which must be nonzero."

smolkaj (approved):
Thanks!

PR #559, opened 2025-07-09
Clarify sending ActionProfile members with duplicate actions.

jonathan-dilorenzo:
Specifically, clarify that P4Runtime clients are only discouraged from sending ActionProfile members with duplicate actions when using the `sum_of_weights` group size calculation, since the same may literally not be achievable using the `weight` field otherwise.

### DIFF
--- a/docs/v1/P4Runtime-Spec.adoc
+++ b/docs/v1/P4Runtime-Spec.adoc
@@ -3992,7 +3992,7 @@ It is possible to include several `ActionProfileAction` messages with the same
 exact `action` specification in one `ActionProfileActionSet` message. For
 `sum_of_members`, this allows clients to specify weights greater than the
 `max_member_weight` set. E.g. modifying the example above, we can weight
-nexthop 1 to be 2x100 (if `max_member_weight` is 100 as follows:
+nexthop 1 to be 2x100 (if `max_member_weight` is 100) as follows:
 [source,protobuf]
 ----
 table_entry {


smolkaj:
>  since the same may literally not be achievable using the weight field otherwise.

Could you provide a concrete example? I may be slow to understand, but I'm not seeing it.

smolkaj:
Did you by any chance `sum_of_members` rather than `sum_of_weights`?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3990,7 +3990,8 @@ inserted. Therefore, 3 distinct `WriteRequest` batches are required.
 
 It is possible to include several `ActionProfileAction` messages with the same
 exact `action` specification in one `ActionProfileActionSet` message. However,
```
smolkaj: From P4 API WG: Let's discuss the two cases (sum of members and sum of weights separtely) to make this clearer, and ideally include an example for each.

Suggested phrasing:
* For `sum_of_members`, this can be useful \<reason\>  
   \<example\>
* For `sum_of_weights`, this is allowed but discouraged, because such ... can always be collapsed into a single message, like so:  
  \<example\>

chrispsommers (approved):
LGTM

jonathan-dilorenzo:
@smolkaj and @chrispsommers, are we ready to merge this? Is it blocked on something?

PR #560, opened 2025-08-07
Add ActionSelectionMode and SizeSemantics to ActionProfileActionSet in P4Runtime.

angelazhang8:
# Motivation 
Our switches currently support per-group hash modes and resource usage modes. We are currently not leveraging this capability. This P4 API change allows for more granular control of the WCMP implementation in GPINS.

# Summary of changes
Add `BucketSelectionMode` (with modes `DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR`, `HASH`, and `RANDOM`) and `SizeSemantics` (with modes `DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR`, `SUM_OF_WEIGHTS`, and `SUM_OF_MEMBERS`) to `ActionProfileActionSet` in p4runtime.proto.


chrispsommers:
@angelazhang8 Thanks for your submission. Could you please add a description, including summary of changes and a motivation for this? Will you be able to present this to us in the WG meeting on Friday 8/8/25?

matthewtlam:
Please update the docs/v1/P4Runtime-Spec.adoc accordingly with the restrictions, expected behavior, etc. You can run the lint tool using `./tools/asciidoclint.py` to format the spec.

Also please use the script `./codegen/update.sh` to update the GoLang, Python and Rust versions

proto/p4/v1/p4runtime.proto
```diff
@@ -284,6 +284,18 @@ message Action {
 
 message ActionProfileActionSet {
   repeated ActionProfileAction action_profile_actions = 1;
+  enum BucketSelectionMode {
```
matthewtlam: Add comment 
`// Added in v1.5.0.`

matthewtlam: Please add comments explaining the use case of this. Assumably, it's to support WCMP

angelazhang8:
@jafingerhut Can you also PTAL? Thanks!

proto/p4/v1/p4runtime.proto
```diff
@@ -284,6 +284,19 @@ message Action {
 
 message ActionProfileActionSet {
   repeated ActionProfileAction action_profile_actions = 1;
+  // Support per-group hash modes and resouce usage modes. Added in v1.5.0.
```
matthewtlam: NIT:  `Added in v1.5.0.` Add to new line.

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3872,10 +3872,15 @@ does not need to keep track of group ids and member ids.
...
 One shots are programmed by choosing the `ActionProfileActionSet` message as the
 `TableAction`. The `ActionProfileActionSet` message consists of a set of
-`ActionProfileAction` messages. This set should have cardinality no greater than
-`max_group_size` (if `max_group_size` is nonzero) if `selector_size_semantics`
-is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
-`ActionProfileAction` message has the following fields:
+`ActionProfileAction` messages, a `BucketSelectionMode` message, and a
+`SizeSemantics` message. The `BucketSelectionMode` enum consists of
+`DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR`, `HASH`, and `RANDOM`.
+The SizeSemantics`SizeSemantics` enum consists of
```
matthewtlam: Used the word `SizeSemantics` twice in a row

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3872,10 +3872,15 @@ does not need to keep track of group ids and member ids.
 
 One shots are programmed by choosing the `ActionProfileActionSet` message as the
 `TableAction`. The `ActionProfileActionSet` message consists of a set of
-`ActionProfileAction` messages. This set should have cardinality no greater than
-`max_group_size` (if `max_group_size` is nonzero) if `selector_size_semantics`
-is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
-`ActionProfileAction` message has the following fields:
+`ActionProfileAction` messages, a `BucketSelectionMode` message, and a
+`SizeSemantics` message. The `BucketSelectionMode` enum consists of
+`DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR`, `HASH`, and `RANDOM`.
```
matthewtlam: May wish to expand more upon the semantics of these enums.

`SizeSemantics` used for resource accounting.

proto/p4/v1/p4runtime.proto
jafingerhut: Do you have target devices in mind that support independent values of bucket_selection_mode and size_semantics for each ActionProfileActionSet?

That is, suppose table t1 has an action selector implementation, and for its corresponding action selector implementation (call it as1), you create 5 groups.  Do you expect to be able the values of bucket_selection_mode and size_semantics independently for each of those 5 groups?

docs/v1/P4Runtime-Spec.adoc
verios-google: Expanding on this comment.

 `BucketSelectionMode` and `SizeSemantics` enums are sufficiently complicated that they should be explained similarly to how `ActionProfileAction` is explained. For example:
 
 The `BucketSelectionMode` enum is used for X. It can be the following values:
   - DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR: This means X (used for backward compatibility)
   - HASH: This means X
   - RANDOM: This means X
   
  This way the flow of this portion will become:
  - ActionProfileSet has 3 components
  - ActionProfileActions (existing stuff)
  - BucketSelectionMode
  - SizeSemantics

proto/p4/v1/p4runtime.proto
```diff
@@ -284,6 +284,19 @@ message Action {
...
+    DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR = 0;
+    HASH = 1;
+    RANDOM = 2;
+  }
+  enum SizeSemantics {
+    DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR = 0;
+    SUM_OF_WEIGHTS = 1;
+    SUM_OF_MEMBERS = 2;
+  }
+  BucketSelectionMode bucket_selection_mode = 2;
```
verios-google: Each of the new fields should have `Added in v1.5.0`. It may feel a little redundant, but the creation of a feature and the use of a feature are separate concepts, best to be explicit instead of having people assume the feature was used when it was first created.

proto/p4/v1/p4runtime.proto
verios-google: This should be split into two comments. One for each new enum since they are separate concepts

jonathan-dilorenzo: Yes, that's exactly right. If I understand my colleagues correctly, this seems to be a per-group attribute in SAI generally, so any SAI-compliant switch (arguably) should have this. (I could be wrong and could try to find a source if helpful).

jonathan-dilorenzo: nit: resource*

jonathan-dilorenzo:
@jafingerhut and @chrispsommers, we internally don't currently have a use-case, but we would be totally amenable to adding these same knobs to ActionProfileGroups for feature parity (since otherwise the selector seems to become more expressive which is a bit odd imo).

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3872,10 +3872,15 @@ does not need to keep track of group ids and member ids.
 
 One shots are programmed by choosing the `ActionProfileActionSet` message as the
 `TableAction`. The `ActionProfileActionSet` message consists of a set of
-`ActionProfileAction` messages. This set should have cardinality no greater than
-`max_group_size` (if `max_group_size` is nonzero) if `selector_size_semantics`
-is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
-`ActionProfileAction` message has the following fields:
+`ActionProfileAction` messages, a `BucketSelectionMode` message, and a
```
jonathan-dilorenzo: nit: message -> enum twice

docs/v1/P4Runtime-Spec.adoc
jonathan-dilorenzo: +1 Would break it up exactly like this. I.e. put the 2 enums in their own 2 paragraphs below the ActionProfileActions explaining their semantics.

docs/v1/P4Runtime-Spec.adoc
jonathan-dilorenzo, line 1: After describing the two enums, we will want to say something about how servers are allowed to treat them. Probably something like:
"
All servers and targets must support the `DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR` and `DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR` values for `BucketSelectionMode` and `SizeSemantics` respectively. Servers must also support the specific value combination that corresponds to that default. Otherwise, if the target does not support per-group hash settings, the server should return `UNIMPLEMENTED`. If the target supports per-group hash settings, but not the specific combination given, the server should return `INVALID_ARGUMENT`.  
"

jonathan-dilorenzo: In comment from meeting, would be good to also say that these modes override the action selector settings.

proto/p4/v1/p4runtime.proto
```diff
@@ -284,6 +284,20 @@ message Action {
 
 message ActionProfileActionSet {
   repeated ActionProfileAction action_profile_actions = 1;
+  // Support per-group hash modes and resouce usage modes.
+  // Added in v1.5.0.
+  enum BucketSelectionMode {
```
jonathan-dilorenzo: Actually, wdyt about ActionSelectionMode? Makes more sense in P4Runtime since we never talk about buckets...

+ Ripple effects obvs.

jonathan-dilorenzo:
In SAI, there's a [per-group hash algorithm attribute](https://github.com/opencomputeproject/SAI/blob/master/inc/sainexthopgroup.h#L291), which one can set to [SAI_HASH_ALGORITHM_RANDOM](https://github.com/opencomputeproject/SAI/blob/b925fc0f0fc8c620eea7d9dead7fc99e33b7b44e/inc/saiswitch.h#L197) to get the random behavior specified here. I'm not aware of switches that allow you to otherwise set arbitrary hash algorithms per group (though they may exist), but we use NONE to pick the default per-switch hash algorithm.

Still looking at the SUM_OF_WEIGHTS vs SUM_OF_MEMBERS, but I suspect it comes from [these SAI group types](https://github.com/opencomputeproject/SAI/blob/master/inc/sainexthopgroup.h#L39). Perhaps the [as-of-yet-undefined WCMP type](https://github.com/opencomputeproject/SAI/blob/master/inc/sainexthopgroup.h#L68) is the one we use for SUM_OF_MEMBERS and have not yet upstreamed.

proto/p4/v1/p4runtime.proto
angelazhang8: Done.

Additionally, there seems to be some more talk of changing
```
enum SizeSemantics {
    DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR = 0;
    SUM_OF_WEIGHTS = 1;
    SUM_OF_MEMBERS = 2;
}
```
to
```
enum SizeSemantics {
    DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR = 0;
    COUNT_OF_MEMBERS = 1;
    SUM_OF_MEMBER_WEIGHTS = 2;
}
```

What do you think of that?

jonathan-dilorenzo: These names are already in use, so wouldn't make sense.

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3872,10 +3872,40 @@ does not need to keep track of group ids and member ids.
...
+* `DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR`: uses the size specified by the
+action selector, ensuring backward compatibility.
+
+* `SUM_OF_WEIGHTS`: the group's size is the sum of the individual weights of
+all its members.
+
+* `SUM_OF_MEMBERS`: the group's size is the total count of its members.
+
+All servers and targets must support the
+DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR and
```
matthewtlam: NIT: In ``. Similarly for all other key variables

proto/p4/v1/p4runtime.proto
```diff
@@ -284,6 +284,23 @@ message Action {
...
+    HASH = 1;
+    RANDOM = 2;
+  }
+  // Added in v1.5.0.
+  enum SizeSemantics {
+    DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR = 0;
+    SUM_OF_WEIGHTS = 1;
+    SUM_OF_MEMBERS = 2;
+  }
+  // Added in v1.5.0.
```
matthewtlam: NIT: Remove `Added in v1.5.0.` for both action_selection_mode and size_semantics because it is depends on SizeSemantics and ActionSelectionMode which is supported in version 1.5.0. I think it's a bit too verbose, but can ignore if everyone else has a strong preference to be verbose

jafingerhut:
> @jafingerhut, @verios-google, @matthewtlam, and @chrispsommers, can we lock in the current proto (possibly with some additional comments)? Or any concerns with the naming?
> 
> Edit: So, I just realized that we need the proto submitted here before we can move forward internally too (since we have external repositories depending on this). We can try to make quick progress on the prose if you'd like? Or maybe we can even do something dumb and submit the proto without prose (or with imperfect prose) saying that the prose is coming by next meeting? If so, we would commit to making all the necessary improvements.

On the .proto file changes alone, I have no objections to the current names.  The most important thing is that they are explained clearly in some section of the specification before the next spec release, preferably with exactly what names they correspond to in the SAI APIs, with links to relevant SAI documentation.  I think it is fine for those details NOT to be in the .proto file, but only in the P4Runtime specification document.

matthewtlam (approved):
Besides my small NITs, everything LGTM. Approving early, but should fix the issue with one of the BUILDs failing though

proto/p4/v1/p4runtime.proto
jonathan-dilorenzo: +1

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3872,10 +3872,40 @@ does not need to keep track of group ids and member ids.
...
+Each `ActionSelectionMode` enum can take on the following values:
+* `DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR`: uses the action specified by the
+action selector, ensuring backward compatibility.
+
+* `HASH`: selects an action based on the hash mode from the action selector.
+
+* `RANDOM`: selects an action at random.
+
+Each `SizeSemantics` enum can take on the following values:
+* `DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR`: uses the size specified by the
```
jonathan-dilorenzo: nit: uses the "size semantics" specified ...

proto/p4/v1/p4runtime.proto
```diff
@@ -284,6 +284,23 @@ message Action {
 
 message ActionProfileActionSet {
   repeated ActionProfileAction action_profile_actions = 1;
+  // Support per-group hash modes and resource usage modes.
+  // Added in v1.5.0.
+  enum ActionSelectionMode {
+    DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR = 0;
+    HASH = 1;
```
jonathan-dilorenzo, line 293: Would be great to add some short comments about what these mean. Tbh, probably the same thing you wrote as bullet points in the spec is fine.

proto/p4/v1/p4runtime.proto
```diff
@@ -284,6 +284,23 @@ message Action {
 
 message ActionProfileActionSet {
   repeated ActionProfileAction action_profile_actions = 1;
+  // Support per-group hash modes and resource usage modes.
```
jonathan-dilorenzo: This doesn't really make sense as a comment here. Can you reframe this to a comment each on ActionSelectionMode and SizeSemantics saying what they're both for?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3872,10 +3872,40 @@ does not need to keep track of group ids and member ids.
...
-`ActionProfileAction` messages. This set should have cardinality no greater than
+`ActionProfileAction` messages, a `ActionSelectionMode` enum, and a
+`SizeSemantics` enum. The `ActionSelectionMode` and `SizeSemantics` enums
+override the existing action selector settings.
+
+Each `ActionSelectionMode` enum can take on the following values:
+* `DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR`: uses the action specified by the
+action selector, ensuring backward compatibility.
+
+* `HASH`: selects an action based on the hash mode from the action selector.
```
jonathan-dilorenzo: Can you add what these correspond to in SAI with the source file? E.g. "In a SAI-based switch, this would correspond to setting `SAI_NEXT_HOP_GROUP_ATTR_HASH_ALGORITHM` to `SAI_HASH_ALGORITHM_NONE`, using the default action selector mode."

Etc.

Reminder of my SAI comments below:
In SAI, there's a [per-group hash algorithm attribute](https://github.com/opencomputeproject/SAI/blob/master/inc/sainexthopgroup.h#L291), which one can set to [SAI_HASH_ALGORITHM_RANDOM](https://github.com/opencomputeproject/SAI/blob/b925fc0f0fc8c620eea7d9dead7fc99e33b7b44e/inc/saiswitch.h#L197) to get the random behavior specified here. I'm not aware of switches that allow you to otherwise set arbitrary hash algorithms per group (though they may exist), but we use NONE to pick the default per-switch hash algorithm.

jonathan-dilorenzo:
@jafingerhut and @angelazhang8, wdyt about separating out the proto PR from the spec PR so that we can keep leaving comments on the spec part without blocking on it?

chrispsommers (approved):
I am OK with approving the .proto now and even leaving the spec a little bit WIP, but let's create an issue to update the spec in a timely manner. We can bump it all to 1.5.1 then? I believe in progress.

jafingerhut:
SInce most of the comments here are about the spec, not the `.proto` file, perhaps leave this issue as it is now, and create a new one for the `.proto` file change by itself that can proceed faster?

proto/p4/v1/p4runtime.proto
verios-google: I prefer things to be verbose since the creation of a concept and it's use are not the same thing and may have different properties (i.e. time of intro). But I'll defer to Jonathan on this

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3872,10 +3872,40 @@ does not need to keep track of group ids and member ids.
...
+All servers and targets must support the
+DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR and
+DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR values for BucketSelectionMode and
+SizeSemantics respectively. Servers must also support the specific value
+combination that corresponds to that default. Otherwise, if the target does not
+support per-group hash settings, the server should return UNIMPLEMENTED. If the
+target supports per-group hash settings, but not the specific combination given,
+the server should return INVALID_ARGUMENT.
+
+The `ActionProfileAction` set should have cardinality no greater than
```
verios-google, line 3879: Nit: Move the existing section above the new section to match the declaration in the proto (as well as the declaration in the earlier paragraph)

proto/p4/v1/p4runtime.proto
verios-google: +1

proto/p4/v1/p4runtime.proto
jonathan-dilorenzo: This is also reasonable to be clear. I could literally see it both ways ;).

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3897,6 +3901,38 @@ is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
   on the `watch_port` field, which also apply for one shot action selector
   programming.
 
+Each `ActionSelectionMode` enum can take on the following values:
+
+* `DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR`: uses the action specified by the
+  action selector, ensuring backward compatibility. In a SAI-based switch, this
+  would correspond to setting `SAI_NEXT_HOP_GROUP_ATTR_HASH_ALGORITHM` to
```
jonathan-dilorenzo: nit: Can we change to "setting the per-group `...` attribute to ..."?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3897,6 +3901,38 @@ is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
...
+  `SAI_HASH_ALGORITHM_NONE`, using the default action selector mode.
+
+* `HASH`: selects an action based on the result of the hashing algorithm chosen
```
jonathan-dilorenzo: nit: "default hashing algorithm used by the switch." maybe?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3897,6 +3901,38 @@ is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
...
+  by the switch.
+
+* `RANDOM`: selects an action at random. In a SAI-based switch, this would
+  correspond to setting `SAI_NEXT_HOP_GROUP_ATTR_HASH_ALGORITHM` to
```
jonathan-dilorenzo: Same comment above with "per-group `...` attribute".

docs/v1/P4Runtime-Spec.adoc
jonathan-dilorenzo: Maybe also add a citation linking to the SAI attribute you're talking about?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3897,6 +3901,38 @@ is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
...
+
+Each `SizeSemantics` enum can take on the following values:
+
+* `DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR`: uses the size semantics
+  specified by the action selector, ensuring backward compatibility.
+
+* `SUM_OF_WEIGHTS`: the group's size is the sum of the individual weights of
+  all its members.
+
+* `SUM_OF_MEMBERS`: the group's size is the total count of its members.
```
jonathan-dilorenzo, line 3931: @jafingerhut, do you think it would be useful to have a SAI correspondence here too? The SUM_OF_MEMBERS one hasn't been upstreamed yet (though they're apparently working on it), but there are a few that correspond to `SUM_OF_WEIGHTS` I believe.

jonathan-dilorenzo (approved):
This is great! Thanks Angela! Mostly small comments, so marking approved.

docs/v1/P4Runtime-Spec.adoc
jafingerhut: Any correspondences / links to SAI documentation of these features seems useful to me.

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3897,6 +3901,38 @@ is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
...
+
+All servers and targets must support the
+`DEFAULT_MODE_DETERMINED_BY_ACTION_SELECTOR` and
+`DEFAULT_SIZE_DETERMINED_BY_ACTION_SELECTOR` values for `ActionSelectionMode`
+and `SizeSemantics` respectively. Servers must also support the specific value
+combination that corresponds to that default. Otherwise, if the target does not
+support per-group hash settings, the server should return `UNIMPLEMENTED`. If
```
jonathan-dilorenzo: nit: Let's change to "per-group hash settings **for this action profile**, the server..."

docs/v1/P4Runtime-Spec.adoc
angelazhang8: Added documentation link for `SUM_OF_WEIGHTS`.

PR #563, opened 2025-09-11
Support `weights_disallowed` field in P4Info

matthewtlam:
The P4 program generally specifies a hash algorithm for use with action selectors. Previously, these did not show up in the P4Info since the choice of algorithm did not affect the control plane. However, we are now running into switches that support a new kind of algorithm called Dynamic Load Balancing (DLB) or Adaptive Routing (AR). See e.g. https://sonic.software/sai/group__SAIARS.html for some of the SAI APIs.

These algorithms have the interesting property that, instead of the controller defining the relative weights in the group, they are dynamically changing based on various measures of port quality. That means that this kind of algorithm **does** affect the control plane, since weights no longer make sense.

We wish to introduce this new field ,`weights_disallowed`, to the P4Info to denote to the control plane that programming will have no effects on the weights (and thus, to avoid confusion, should be rejected).

cc: @jonathan-dilorenzo, @smolkaj  for vis

### DIFF
--- a/docs/v1/P4Runtime-Spec.adoc
+++ b/docs/v1/P4Runtime-Spec.adoc
@@ -3901,7 +3901,7 @@ is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
   message must not exceed the `max_group_size` specified in the P4Info
   (if greater than 0), or the server must return `INVALID_ARGUMENT`.
   If `selector_size_semantics` is `sum_of_members`, the individual weight of
-  each member must not exceed`max_member_weight` (if greater than 0), or the
+  each member must not exceed `max_member_weight` (if greater than 0), or the
   server must return `INVALID_ARGUMENT`.
 
 * `watch_port` is the controller-defined port that the action's liveness depends


proto/p4/config/v1/p4info.proto
```diff
@@ -323,27 +323,33 @@ message ActionProfile {
...
+    // Added in v1.4.0.
     SumOfWeights sum_of_weights = 6;
     // group size is the sum of the group's members.
-    // Added in v1.4.0
+    // Added in v1.4.0.
     SumOfMembers sum_of_members = 7;
   }
+
+  // dictates whether the controller can specify weights for groups programmed
+  // in this ActionProfile. If `weights_determined_by_switch`, then all weights
```
jonathan-dilorenzo: ... And will instead be determined internally by the switch.

proto/p4/config/v1/p4info.proto
```diff
@@ -323,27 +323,33 @@ message ActionProfile {
...
+  // must be absent.
+  // Added in v1.5.0.
+  bool weights_determine_by_switch = 8;
```
antoninbas: s/weights_determine_by_switch/weights_determined_by_switch

you have the correct version in the comment

proto/p4/config/v1/p4info.proto
chrispsommers: "If `weights_determined_by_switch`, seems like an incomplete phrase."

How about "If `weights_determined_by_switch` is true"

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -1504,6 +1504,10 @@ The `ActionProfile` message includes the following fields:
       maximum weight of each individual member. If unset, any 32-bit integer is
       allowed for weight.
 
+* `weights_disallowed`, a boolean flag indicating whether the controller
+   programs the weights or the switch determines the weights based on dynamic
```
jonathan-dilorenzo: How about "... programs the weights (false) or the switch determines them through some other mechanism (true). One such mechanism is dynamic load balancing or adaptive routing (See e.g. SAI API)."

And then link to SAI API:https://sonic.software/sai/group__SAIARS.html through the standard mechanism

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -1504,6 +1504,10 @@ The `ActionProfile` message includes the following fields:
       maximum weight of each individual member. If unset, any 32-bit integer is
       allowed for weight.
 
+* `weights_disallowed`, a boolean flag indicating whether the controller
```
jonathan-dilorenzo: I think I'd put this below the "PSA programs ..." paragraph, and also add "The @weights_disallowed annotation can be added to action selectors in P4 programs to set this value to true." Or something like that.

docs/v1/P4Runtime-Spec.adoc
jonathan-dilorenzo, line 1: Also update this section to say weight 0 is valid iff `weights_disallowed` is set: http://screen/BmLLHEkZJiXWPMi (I know this only works internally). And that any other weight is disallowed if that is set.

proto/p4/config/v1/p4info.proto
```diff
@@ -323,27 +323,33 @@ message ActionProfile {
...
     SumOfWeights sum_of_weights = 6;
     // group size is the sum of the group's members.
-    // Added in v1.4.0
+    // Added in v1.4.0.
     SumOfMembers sum_of_members = 7;
   }
+
+  // dictates whether the controller can specify weights for groups programmed
+  // in this ActionProfile. If `weights_disallowed` is true, then all weights
+  // must be absent.
```
jonathan-dilorenzo: Add: "Unset (false) in action profiles."

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3880,16 +3891,17 @@ is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
...
-  `max_group_size` specified in the P4Info (if greater than 0), or the server
-  must return `INVALID_ARGUMENT`. If `selector_size_semantics` is
-  `sum_of_members`, the individual weight of each member must not exceed
-  `max_member_weight` (if greater than 0), or the server must return
-  `INVALID_ARGUMENT`.
+* `weight` specifying the probability of the action's selection at runtime.
+  If `weights_disallowed` is set to true, then 0 (unset) is the only valid
+  `weight`. If `weights_disallowed` is false, then 0 is not a valid `weight`.
+  The server must return `INVALID_ARGUMENT` if the client attempts to use invalid
+  `weight`s. If `selector_size_semantics` is `sum_of_weights`, then the sum of
```
jonathan-dilorenzo: nit: If it's possible to put the "If `selector_size_semantics`  ... " one line down, I think that's a bit more readable and may have a smaller apparent diff.

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -3880,16 +3891,18 @@ is `sum_of_members`, or else the server must return `INVALID_ARGUMENT`. Each
...
+  If `weights_disallowed` is set to true, then 0 (unset) is the only valid
+  `weight`. If `weights_disallowed` is false, then 0 is not a valid `weight`.
+  The server must return `INVALID_ARGUMENT` if the client attempts to use invalid
+  `weight`s.
+  If `selector_size_semantics` is `sum_of_weights`, then the sum of all weights
+  across all `ActionProfileAction` messages for that `ActionProfileActionSet`
+  message must not exceed the `max_group_size` specified in the P4Info
+  (if greater than 0), or the server must return `INVALID_ARGUMENT`.
+  If `selector_size_semantics` is `sum_of_members`, the individual weight of
+  each member must not exceed`max_member_weight` (if greater than 0), or the
```
jafingerhut: Typo: I think there should be a space after "exceed" and before `max_member_weight`.

smolkaj: @matthewtlam there is an open comment here that blocks merging.

PR #594, opened 2026-02-23
Support `group_action` field in ActionProfileActionSet

matthewtlam:
# Objective
Support per group action for ActionProfileActionSets in P4.

# Motivation
In the P4Runtime API, the ActionProfileActionSet is used to program actions associated with an action profile, which is typically used for modeling the forwarding behaviors for ECMP/WCMP. Each ActionProfileAction represents a weighted path with an action that needs to be executed (e.g. set nexthop or encap) for a WCMP member. P4 currently does not have a native way to provide a per group action that applies on all paths of an ActionProfileActionSet. A common scenario that would require this is setting a metadata for any path taken by the ActionProfile (e.g. setting a route_metadata).

### Option 1: Action Composition
Define new actions in the P4 program that combine the desired "group" logic with each specific member action. For example, adding a group parameter of route_metadata to every single action.

Pros:
- Applies actions sequentially to get desired forwarding behavior

Cons:
- Will require additional checks to make sure that the group parameter is consistent across all actions in the action set, which is not current expressible in P4Runtime
- Redundancy of defining unnecessary actions or additional parameters

### Option 2: Two P4 Tables
Use a primary table with an action profile behaves “normally” and use another table for lookup before or after the primary table using the action profile to apply the group logic.

Pros:
- Applies actions sequentially to get desired forwarding behavior

Cons:
- Cannot atomically update 2 tables in lock step
- May take up additional resources (depending on hardware)
- Changes the pipeline structure

# Design Options

## Option 1: P4Runtime Native Support (Implementation reflected in this PR)
Extend the P4Runtime API to support per group action in action sets. The proposed `group_action` in ActionProfileActionSet adds a native mechanism for group parameters. The `group_action` should behave as a "base" action that is logically composed with the selected member action. Semantically, it functions as a pre-action to the existing member-specific logic, effectively providing the shared context needed for the pipeline to process any selected ActionProfileAction. 

Pros
- Reduced redundancy and clearer intent by not specifying the same common action parameters for every member in the group
- Save space since we do not need to use another table to represent this and instead have one action rather than sending multiple requests
- Clearly defined P4 types for the group parameters

Cons
- Only applies to ActionSelectors with the oneshot annotation

### P4 Runtime Changes  
```
// p4runtime.proto
message TableAction {
  oneof type {
    Action action = 1;
    uint32 action_profile_member_id = 2;
    uint32 action_profile_group_id = 3;
    ActionProfileActionSet action_profile_action_set = 4;
  }
}

message Action {
  uint32 action_id = 1;
  message Param {
    uint32 param_id = 2;
    bytes value = 3;
  }
  repeated Param params = 4;
}

message ActionProfileActionSet {
  repeated ActionProfileAction action_profile_actions = 1;
    Action group_action = 5;             <-------------- Added line
  ...
}
```

### P4Info Changes
Ideally there should be a way to specify that an action defined for a table can only be used as a per group action. To specify this in P4 code, we propose to add an annotation (e.g., @pergroupaction) on the actions list in the table declaration.  This allows the target to enforce that the designated action can be used only as a per group action in the ActionProfileActionSets.  

Example of using the annotation: 
```
  ActionSelector(...) wcmp_group_selector;
  table wcmp_group_table {
    key = {
      group_id : exact;
      selector_input : selector;
    }
    actions = {
      set_nexthop_id;
      @pergrouponly group_action;           <------------------ Using the new annotation
      @defaultonly NoAction;
    }
    const default_action = NoAction;
    implementation = wcmp_group_selector;
  }
```

Ideally, this can be reflected in Scope used in ActionRef by the P4 compiler:
```
// p4info.proto
message ActionRef {
  uint32 id = 1;
  enum Scope {
    TABLE_AND_DEFAULT = 0;
    TABLE_ONLY = 1;
    DEFAULT_ONLY = 2;
    PER_GROUP_ONLY = 3;         <-------------------Added line
  }
  Scope scope = 3;
  repeated string annotations = 2;
  // Optional. If present, the location of `annotations[i]` is given by
  // `annotation_locations[i]`.
  repeated SourceLocation annotation_locations = 5;
  repeated StructuredAnnotation structured_annotations = 4;
}
```

Alternatively or as a short term solution, the annotation can be embedded in annotations repeated field instead of first class support in p4info. 

## Option 2: Metadata
Embed the per group metadata setting in the table entry’s metadata. 

Pros:
- Metadata is free-form so the per group data can be manipulated and stored

Cons:
- Strictly not P4Runtime standard compliant because 
- Switch and controller need to agree on schema of metadata
- Type/encoding is not captured by P4Runtime specification

cc: @kheradmand @smolkaj @jonathan-dilorenzo


matthewtlam:
This PR focus mostly on the API changes. A follow-up PR will be made to update the spec accordingly to handle all the intricacies of adding this global action

proto/p4/v1/p4runtime.proto
```diff
@@ -284,6 +284,10 @@ message Action {
 
 message ActionProfileActionSet {
   repeated ActionProfileAction action_profile_actions = 1;
+  
+  // Added in v1.6.0.
```
smolkaj, line 286: 1.5.1

smolkaj (changes requested):

proto/p4/v1/p4runtime.proto
matthewtlam: Resolving this comment since we discussed that this should increment the version number since it's a new feature

jafingerhut:
Is it important to you that BMv2 implement this new behavior?  If so, are you thinking of identifying some people that could make the necessary changes, which are probably spread over at least these repos?
+ https://github.com/p4lang/behavioral-model
+ https://github.com/p4lang/PI
+ https://github.com/p4lang/p4c _might_ need changes - I am not sure about that.  Depends upon whether it preserves the new annotation `@globalonly` and writes it into the relevant output file(s) that BMv2 uses.

proto/p4/config/v1/p4info.proto
```diff
@@ -278,6 +278,7 @@ message ActionRef {
     TABLE_AND_DEFAULT = 0;
     TABLE_ONLY = 1;
     DEFAULT_ONLY = 2;
+    GLOBAL_ONLY = 3;
```
jonathan-dilorenzo: Add the "// Added in v1.6.0." here too.

matthewtlam:
> A little bikeshedding on the name `@global_only`. Alternate suggestion:
> 
> * `@per_group_only`
> * `@group_action_only`
> 
> In particular, something with `group` in its name seems like a good idea, to make it clear that this is specific to groups in action selectors.

Open minded to names. WDYT @smolkaj or @jonathan-dilorenzo or @chrispsommers?

matthewtlam:
@jafingerhut 

Had a discussion with @smolkaj today how we could use AI, perhaps Claude code to easily add support for global actions to BMv2 within 1-2 weeks. Similarly the same on the P4C side if that is indeed needed. 

I've also updated the documentation in the P4Runtime Spec to reflect the oneshot comment as well as this global action being a pre action to the "local" action. 

Please take another look and let me know if there are any concerns on your side

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -1350,16 +1350,20 @@ control plane. This message contains the following fields:
...
-       `TABLE_AND_DEFAULT`, `TABLE_ONLY` and `DEFAULT_ONLY`. The `scope` of the
-       action is determined by the use of the P4 standard annotations
-       `@tableonly` and `@defaultonly` cite:[P4ActionAnnotations]. `TABLE_ONLY`
-       (`@tableonly` annotation) means that the action can only appear within
-       the table, and never as the default action. `DEFAULT_ONLY`
+    ** `scope`, an enum value which can take one of four values:
+       `TABLE_AND_DEFAULT`, `TABLE_ONLY`, `GLOBAL_ONLY` and `DEFAULT_ONLY`. The
+       `scope` of the action is determined by the use of the P4 standard
+       annotations `@tableonly`, `@globalonly`, and `@defaultonly`
+       cite:[P4ActionAnnotations]. `TABLE_ONLY` (`@tableonly` annotation) means
```
jonathan-dilorenzo: Would reframe since I don't believe @globalonly is a default annotation: https://p4.org/wp-content/uploads/sites/53/2024/10/P4-16-spec-v1.2.5.html#sec-table-action-anno

You could try to add that as a default annotation, but probably don't want to block on that.

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -1350,16 +1350,20 @@ control plane. This message contains the following fields:
...
+       that the action can only appear within the table, and never as the
+       default action. `GLOBAL_ONLY` (`@globalonly` annotation) means that
+       the action can only appear within the table, but only used in the
```
jonathan-dilorenzo: Would remove "only appear within the table, but", since it's superseded by the only used part.

How about:
" `GLOBAL_ONLY` (`@globalonly` annotation) means that
       the action can only be used as a `global_action` in an `ActionProfileActionSet`."

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4104,7 +4109,42 @@ Note that to preserve read-write symmetry, the server
 must not coalesce multiple `ActionProfileAction` messages with the same `action`
 specification into one. Additionally, each `action` specification would count as
 a separate member for the purposes of e.g. the `sum_of_members` group size
-calculation for Action Selectors.
+calculation for Action Selectors
```
jonathan-dilorenzo: nit: Add back period.

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4104,7 +4109,42 @@ Note that to preserve read-write symmetry, the server
...
+
+The `global_action` is a native global metadata setting mechanism for
```
jonathan-dilorenzo: Would remove first line and start with "The `global_action` behaves ..."

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4104,7 +4109,42 @@ Note that to preserve read-write symmetry, the server
...
+`ActionProfileActionSet`s. This behaves as a "base" action that is logically
+composed with the selected member action. Semantically, this functions acts as
```
jonathan-dilorenzo: Remove "functions" or "acts"

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4104,7 +4109,42 @@ Note that to preserve read-write symmetry, the server
...
+a pre-action to the existing member-specific logic, effectively providing a
```
jonathan-dilorenzo: Would remove ", effectively providing a
shared context needed for the pipeline to process any selected
`ActionProfileAction`"

Or at least, I'm not sure what useful information that provides?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4104,7 +4109,42 @@ Note that to preserve read-write symmetry, the server
...
+shared context needed for the pipeline to process any selected
+`ActionProfileAction`. For example, the global action that sets the route
+metadata would occur first and then the nexthop is set for each
+`ActionProfileAction`. Note that this functionality can only be used with
+the one shot method.
```
jonathan-dilorenzo: programming* method (or check what it actually says in spec atm).

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4104,7 +4109,42 @@ Note that to preserve read-write symmetry, the server
...
 a separate member for the purposes of e.g. the `sum_of_members` group size
-calculation for Action Selectors.
+calculation for Action Selectors
+
+The `global_action` is a native global metadata setting mechanism for
+`ActionProfileActionSet`s. This behaves as a "base" action that is logically
+composed with the selected member action. Semantically, this functions acts as
+a pre-action to the existing member-specific logic, effectively providing a
+shared context needed for the pipeline to process any selected
+`ActionProfileAction`. For example, the global action that sets the route
```
jonathan-dilorenzo: " For example, the global action that sets the route metadata would occur first and then the nexthop is set for each `ActionProfileAction`."

This is a bit harder to contextualize outside of SAI. How about "For example, you might use it to set metadata for every packet that hits a particular group, irrespective of the member that then gets selected."

jonathan-dilorenzo (approved):

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4106,6 +4111,38 @@ specification into one. Additionally, each `action` specification would count as
...
+shot action selector programming.
+
+P4Runtime-Spec.adoc
+----
+table_entry {
+  table_id: 0x0212ab34
+  match { /* lpm match */ }
+  action {
+    action_profile_action_set {
+      global_action: {
```
kheradmandG: nit: remove ':'

docs/v1/P4Runtime-Spec.adoc
matthewtlam: Made a note that " `@globalonly` is currently not a P4 standard annotation. "

jafingerhut:
Attaching slides I presented at 2026-Mar-13 P4 API working group meeting on the topic of the name for the new annotation, and more generally the name of this enhancement proposal.
[per-group-action-proposal-naming.pptx](https://github.com/user-attachments/files/25979663/per-group-action-proposal-naming.pptx)

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4106,6 +4111,38 @@ specification into one. Additionally, each `action` specification would count as
...
+shot action selector programming.
+
+P4Runtime-Spec.adoc
+----
+table_entry {
+  table_id: 0x0212ab34
+  match { /* lpm match */ }
+  action {
+    action_profile_action_set {
+      per_group_action {
```
verios-google: *group_action

proto/p4/v1/p4runtime.proto
```diff
@@ -284,6 +284,10 @@ message Action {
...
+  Action per_group_action = 5;
```
verios-google: group_action

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4106,6 +4111,38 @@ specification into one. Additionally, each `action` specification would count as
 a separate member for the purposes of e.g. the `sum_of_members` group size
 calculation for Action Selectors.
 
+The `per_group_action` behaves as a "base" action that is logically composed with
```
verios-google: group_action

verios-google (commented):
Just update the table entry message name to group_action. Per group works in the config, but group_action should be used when programming

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -1350,16 +1350,20 @@ control plane. This message contains the following fields:
...
     ** `id`, the `uint32` identifier of the action.
-    ** `scope`, an enum value which can take one of three values:
-       `TABLE_AND_DEFAULT`, `TABLE_ONLY` and `DEFAULT_ONLY`. The `scope` of the
-       action is determined by the use of the P4 standard annotations
-       `@tableonly` and `@defaultonly` cite:[P4ActionAnnotations]. `TABLE_ONLY`
+    ** `scope`, an enum value which can take one of four values:
+       `TABLE_AND_DEFAULT`, `TABLE_ONLY`, `PER_GROUP_ONLY` and `DEFAULT_ONLY`. The
+       `scope` of the action is determined by the use of the P4 standard
+       annotations `@tableonly`, and `@defaultonly` cite:[P4ActionAnnotations].
+       `@pergrouponly` is currently not a P4 standard annotation. `TABLE_ONLY`
```
smolkaj: but planned?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -1350,16 +1350,20 @@ control plane. This message contains the following fields:
...
+       the table, and never as the default action. `PER_GROUP_ONLY`
+       (`@pergrouponly` annotation) means that the action can only be used in as
+       a `per_group_action` in an `ActionProfileActionSet`. `DEFAULT_ONLY`
        (`@defaultonly` annotation) means that the action can only be used as the
        default action. `TABLE_AND_DEFAULT` is the default value for the enum and
-       means that neither annotation was used in P4 and that the action can be
-       used both within the table and as the default action.
+       means that no annotation was used in P4 and that the action can be
+       used both within the table and as the default action, but not as a
+       per group action.
```
smolkaj: Nit: when you use "per group" as an adjective, I'd suggest per-group, as in "per-group action" (compound adjective)

docs/v1/P4Runtime-Spec.adoc
smolkaj: FWIW, I like Victor's suggestion of suing the shorter name for the message field and the longer name when you need to disambiguate

proto/p4/v1/p4runtime.proto
smolkaj: If this get's executed before the other actions, should we also put it syntactically before the other actions?

smolkaj: A one-line comment explaining what this is and does also would be nice.

smolkaj (changes requested):
LGTM modulo open comments. Nice work. Matthew!

One additional ask: can you please summarize what our plans for p4c and BMv2 support are?

docs/v1/P4Runtime-Spec.adoc
matthewtlam: Yup, but planned. Added it to the spec

matthewtlam:
> LGTM modulo open comments. Nice work. Matthew!
> 
> One additional ask: can you please summarize what our plans for p4c and BMv2 support are?

@smolkaj 

I think at the high level these are the steps that need to be taken

1. Update P4Lang/PI to use the new P4Runtime standard (we will need this for BMv2 to understand the new schema)
2. Update P4C to support the new @pergrouponly annotation
3. Update the PI library in BMv2
4. Implement the group action logic in BMv2

I think I still have to understand the lower level implementation details, but I think AI could be used to figure out most of this.

cc: @jafingerhut, @chrispsommers let me know if I am missing anything

smolkaj (approved):

chrispsommers (approved):
LGTM thanks

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -1350,16 +1350,20 @@ control plane. This message contains the following fields:
...
-       (`@tableonly` annotation) means that the action can only appear within
-       the table, and never as the default action. `DEFAULT_ONLY`
+    ** `scope`, an enum value which can take one of four values:
+       `TABLE_AND_DEFAULT`, `TABLE_ONLY`, `PER_GROUP_ONLY` and `DEFAULT_ONLY`. The
+       `scope` of the action is determined by the use of the P4 standard
+       annotations `@tableonly`, and `@defaultonly` cite:[P4ActionAnnotations].
+       `@pergrouponly` is currently not a P4 standard annotation, but planned.
+       `TABLE_ONLY` (`@tableonly` annotation) means that the action can only
+       appear within the table, and never as the default action. `PER_GROUP_ONLY`
+       (`@pergrouponly` annotation) means that the action can only be used in as
```
jafingerhut: "used in as" -> "used as" ?

docs/v1/P4Runtime-Spec.adoc
```diff
@@ -4106,6 +4111,38 @@ specification into one. Additionally, each `action` specification would count as
 a separate member for the purposes of e.g. the `sum_of_members` group size
 calculation for Action Selectors.
 
+The `group_action` behaves as a "base" action that is logically composed with
+the selected member action. Semantically, this acts as a pre-action to the
+existing member-specific logic. For example, you might use it to set metadata
+for every packet that hits a particular group, irrespective of the member that
+then gets selected. Note that this functionality can only be used with one
+shot action selector programming.
```
jafingerhut, line 4119: Just a question for my curiosity: It seems there is nothing fundamental blocking us making it possible to specify group actions when using the "not one-shot" method of configuring action selectors.  However, that is not urgent at this time, and can be done in later enhancements to the P4Runtime spec if someone wishes.  Is all of that correct?

jafingerhut (approved):
LGTM.  Would be nice to fix the minor typo I found before merging, but not a requirement as it is quite understandable what the meaning is.

docs/v1/P4Runtime-Spec.adoc
matthewtlam: I'm not too sure on the specifics implementation details of the one-shot method. My understanding was that it was originally supposed to be syntactic sugar on top of the "non-one shot" method. I was told by @kheradmandG that there's a difference now so I specified that it only applies to "one-shot only"

@smolkaj or @jonathan-dilorenzo do you have any context on this?

docs/v1/P4Runtime-Spec.adoc
matthewtlam: Resolved the typo. Thanks for pointing it out Andy

docs/v1/P4Runtime-Spec.adoc
jonathan-dilorenzo: Exactly correct @jafingerhut. This should be as easy as adding the same thing to the ActionProfileGroup proto.

We just don't have any use case (and don't actually support those at all I'm pretty sure) so we aimed to keep it simple by targeting only one piece.

PR #601, opened 2026-03-20
Remove root-BUILD-file config from proto/BUILD.bazel

smolkaj:
The config has been moved to the root BUILD.bazel, and we forgot to delete it here.


smolkaj:
Gonna merge this without review as a one-off, since it's a minor fix.

smolkaj:
Please reach out if there are any concerns or it causes issues, and we can immediately revert.

PR #604, opened 2026-03-24
Change PER_GROUP_ONLY to GROUP_ACTION and update P4Runtime spec

matthewtlam:
This name change from PER_GROUP_ONLY to ACTION_GROUP is to avoid the confusion regarding the expected semantics of the @groupaction annotation. The @tableonly and @defaultonly are defined in a way that if an action has no annotation, that means that can be used for both table and default).

But our semantics of @groupaction (previously @pergrouponly) is inconsistent with that, in the sense that if the annotation is present on an action, it can be used as a group action (and nothing else) and if the annotation is absent, the action cannot be used as a group action. 

This was also changed to maintain consistency across annotation and scope


docs/v1/P4Runtime-Spec.adoc
```diff
@@ -1351,13 +1351,13 @@ control plane. This message contains the following fields:
   specified in the same P4Info message and it includes the following fields:
     ** `id`, the `uint32` identifier of the action.
     ** `scope`, an enum value which can take one of four values:
-       `TABLE_AND_DEFAULT`, `TABLE_ONLY`, `PER_GROUP_ONLY` and `DEFAULT_ONLY`. The
+       `TABLE_AND_DEFAULT`, `TABLE_ONLY`, `PER_GROUP` and `DEFAULT_ONLY`. The
```
jonathan-dilorenzo: Why not GROUP_ACTION?

jonathan-dilorenzo: Or, it seems sorta nice that they are the same...

jafingerhut: Agreed that if `@groupaction` is the annotation, then `GROUP_ACTION` as the value of the enum would be nice to keep it similar to existing corresponding attribute/enum-value-names:

+ @defaultonly -> DEFAULT_ONLY
+ @tableonly -> TABLE_ONLY
+ @groupaction -> GROUP_ACTION

kheradmandG:
@matthewtlam could you update the description on why we are proposing this change?

This is to avoid the confusion regarding the expected semantics of the `@groupaction` annotation. 
The `@tableonly` and `@defaultonly` annotations are defined in a way that if an action has no annotation, that means the action can be used for both table and default. 
But our semantics of `@groupaction` (previously `@pergrouponly`) is inconsistent with that, in the sense that if the annotation is present on an action, it can be used as a group action (and nothing else) and if the annotation is absent, the action cannot be used as a group action. 
The drop of "only" postfix in `@groupaction` is meant to indicate that there is a difference between how `@groupaction` annotation vs. `@tableonly` and `@defaultonly` annotations work. 


Also it seems there is consensus on @groupaction / GROUP_ACTION. Could you please update the PR.

kheradmandG:
@jafingerhut @jonathan-dilorenzo @smolkaj if everyone is happy with the idea, let's get this in soon before we have developments with the older annotation.

jafingerhut:
> @jafingerhut @jonathan-dilorenzo @smolkaj if everyone happy with the idea, let's get this in soon before we have developments with the older annotation.

I'll let others speak for themselves, but for myself, I am happy to approve if we make the suggested change of annotation called `@groupaction` and enum value name changed to `GROUP_ACTION`.

matthewtlam:
@jafingerhut @chrispsommers @smolkaj @jonathan-dilorenzo @kheradmandG 

I updated it so everything is consistent with each other. Please take a look

jonathan-dilorenzo (approved):
LGTM!

jafingerhut (approved):
LGTM

smolkaj (approved):
LGTM

chrispsommers (approved):
LGTM

PR #607, opened 2026-04-11
proto: add Java proto-library targets

smolkaj:
Adds first-class `java_proto_library` Bazel targets for all four protos
(`p4runtime`, `p4data`, `p4info`, `p4types`) and sets the standard set
of Java file options on each.

## Why

- **Bazel:** Today every Kotlin/Java consumer (4ward, DVaaS, sonic-pins,
  ...) re-declares these wrappers locally. Strict-deps rule sets
  (e.g. rules_kotlin with `strict_kotlin_deps = "error"`) require a
  *direct* `java_proto_library` per referenced proto — transitive
  exposure isn't enough. Shipping them upstream removes the boilerplate.

- **Java options:** Without explicit options, generated Java packages
  fall back to defaults that vary by environment (e.g. google3 defaults
  to `com.google.protos.<package>`), so the same imports break across
  build systems. Setting them explicitly pins behavior.

## Conventions

All choices align with how protobuf's own well-known types are
configured ([`wrappers.proto`][1], [`timestamp.proto`][2], …):

- `java_package = "org.p4.v1"` / `"org.p4.config.v1"` — reverse-DNS
  prefix (`org.p4`) ensures global uniqueness and follows Java
  conventions. Decided by the working group.
- `java_outer_classname = "<File>Proto"` — standard `Proto` suffix,
  used uniformly even when there's no name collision (forward-compat:
  adding a same-named message later won't force a rename).
- `java_multiple_files = true` — top-level messages get their own
  `.java` file, so imports collapse from
  `org.p4.v1.P4RuntimeOuterClass.ForwardingPipelineConfig` to
  `org.p4.v1.ForwardingPipelineConfig`.

No existing Java consumers to break — this PR introduces the Java
proto-library targets in the first place.

[1]: https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/wrappers.proto
[2]: https://github.com/protocolbuffers/protobuf/blob/main/src/google/protobuf/timestamp.proto


smolkaj:
Would be grateful for a sanity check.
Should be routine and boring, I'm looking for rubber stamping :)

smolkaj:
@matthewtlam could you PTAL?

matthewtlam (approved):
LGTM, thanks Steffen!

smolkaj:
Discussed in working group meeting:
- Change java package name to start with "org.p4" to ensure global uniqueness and align with Java conventions

chrispsommers (approved):
Approve conditional on prepending package name with `org.` per meeting convo.

PR #611, opened 2026-05-06
Automatic releases and PR labels

chreekat:
This change is based on similar machinery recently added to ptf, bmv2, and PI.

1. PRs get labeled automatically based on which files they modify
2. Once a month, a PR gets created automatically that bumps the new VERSION file 
3. Separately, another workflow notices when the VERSION file changes on main, and pushes a corresponding tag.
4. The tag push triggers the existing "Tag uploads" job
5. The tag push also triggers a **new** workflow that creates a GitHub release. It makes use of the automatic PR labels.

As an unrelated bonus, I set up dependabot to bump GitHub Action dependencies.


chrispsommers:
I'm unclear why we would want automatic monthly releases. Is there some writeup which covers the motivation for this? I'm sure the other projects vetted this and have good rationale, but I'm not privy to the conversations which led to that. Thanks!

jafingerhut:
p4c has had automated monthly releases for 1-2 years now, and it seems reasonable there, as it is a minor version bump every month, with the 20-30 commits or so that tend to be made each month in the newest release.

I agree that for specifications like the language spec in the p4-spec repo, and the P4Runtime API spec and its associated code files in this p4runtime repo, the long-standing practice has been that the relevant working groups decide when new releases are made.  It seems reasonable to keep it that way to me, as long as we have steps for producing a new release that are reasonably well documented for each working group to follow.

@chreekat I think this particular repository is probably either  in the "we don't want automated monthly releases" category, or we should have more discussion with the relevant working groups to see if that is something they want to implement.

chreekat:
It did cross my mind that automatic releases for this repo might be inappropriate. Making the PR was easy enough that I decided it would be the way I raise the question! We can simply close it now. 

@jafingerhut are there other repos that *do* still need the automatic release treatment?

jafingerhut:
> @jafingerhut are there other repos that _do_ still need the automatic release treatment?

Good question.  The most important ones for automatic release would be:

+ p4c - automated releases have already been implemented a couple of yeras ago, and working smoothly,.
+ behavioral-model - already implemented, but being debugged now in issues and PRs
+ PI
+ p4mlir-incubator - Ask @asl if he thinks it would be useful to have automated releases.
+ ptf - For this one, we should consider whether we want the automated release process to include creating a new version of the package published on pypi.org
+ p4runtime-shell - same comment as for ptf
+ open-p4studio - This has not changed much since most files were added a while ago.  As for all other repositories, any automated release process should do nothing if no changes have been made since the previous release.
+ p4-constraints - ask @smolkaj if he thinks it would be helpful to have an automated release process for this one.
+ third-party - Not clear to me where this is still used, if at all.  If it is a core piece of infrastructure, or becomes one, then an automated release process is worth considering.

I create VM images for the tutorials repository every month, manually.  I don't know a good way to fully automate that process, but it might be nice if there was one, if it was not difficult to do so.  I can create an issue in that repository with some thoughts on that idea there.

The repos listed below are currently released manually, at the decision of the chairs of the relevant work group, and I would guess that manual process is preferred:

+ p4-spec
+ p4runtime
+ pna

The repositories below do not have any code that I am aware of, thus an automated release seems not useful:

+ gsoc
+ project-ideas
+ p4per

The repositories below are quite inactive since 2024, so no need to create an automated release process for them, unless development becomes more active on them:

+ tdi
+ p4-dpdk-target
+ target-syslibs
+ target-utils
+ p4pi
+ p4analyzer

All of the other p4lang repositories that I have not mentioned explicitly in this comment are because it has been 5 years since the repository has been useful, and it is either already archived status, or should be considered to change to archived status.

PR #612, opened 2026-05-08
Round 2 of adding copyright and license notices to files

jafingerhut:
With the current state of this PR as of commit 1, this adds copyright and license notices that are lintable via the open source REUSE program for all remaining files, except 2 that I am checking with Davide Scano whether he wrote them, or copied them from somewhere else.  I think this is reasonable to review and merge without those, and I can add info for those 2 files in a later PR if I do not find out the details soon enough for this PR.

For all files modified (or for which a <origfilename>.license file was added, if the original file does not support comments), the copyright holder is the person who first added the file to this repository, and the year is the year that happened.

### DIFF
--- a/py/REUSE.toml
+++ b/py/REUSE.toml
@@ -1,0 +1,8 @@
+version = 1
+
+[[annotations]]
+path = [
+    "p4/**"
+]
+SPDX-FileCopyrightText = "2020 The P4 Language Consortium"
+SPDX-License-Identifier = "Apache-2.0"
--- a/rust/Cargo.lock.license
+++ b/rust/Cargo.lock.license
@@ -1,0 +1,3 @@
+SPDX-FileCopyrightText: 2024 Campbell He <kp.campbell.he@duskmoon314.com>
+
+SPDX-License-Identifier: Apache-2.0
--- a/.bazelignore
+++ b/.bazelignore
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2026 Steffen Smolka
+#
+# SPDX-License-Identifier: Apache-2.0
+
 # Exclude bazel sub-projects so `bazel build //...` works.
 bazel/example
 
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,3 +1,7 @@
+# SPDX-FileCopyrightText: 2026 Steffen Smolka
+#
+# SPDX-License-Identifier: Apache-2.0
+
 # Use Bzlmod (`MODULE.bazel`) instead of `WORKSPACE.bazel`.
 common --enable_bzlmod
 common --noenable_workspace
--- a/.bazelversion.license
+++ b/.bazelversion.license
@@ -1,0 +1,3 @@
+SPDX-FileCopyrightText: 2022 Steffen Smolka
+
+SPDX-License-Identifier: Apache-2.0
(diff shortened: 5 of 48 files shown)


jafingerhut:
I found the original source of the 2 files in question, and added copyright holder, year, and license to them in commit 2 of this PR.  All files are now finished with this PR (unless new ones are added before this one is merged, which is perfectly fine as we can mark them later if needed).

jafingerhut:
@smolkaj @chrispsommers I think this is ready to review, and to merge if you approve.

There are a few source files used in generating the output from the Asciidoc source of the P4Runtime API spec that Davide Scano copied from other places, and I have tracked down the copyright holder and license for those.

The majority of files have the copyright year as the year that the file was first added to this repository, the copyright holder as the individual who first added the file to this repository, and the license as Apache-2.0.  The year the file was added to this repo I retrieved from the git commit log via a little program I wrote, so they should be trustworthy and there should be no need to double-check them all.

py/REUSE.toml
```diff
@@ -0,0 +1,8 @@
+version = 1
+
+[[annotations]]
+path = [
+    "p4/**"
+]
+SPDX-FileCopyrightText = "2020 The P4 Language Consortium"
+SPDX-License-Identifier = "Apache-2.0"
```
jafingerhut, line 8: smolkaj chrispsommers The REUSE software allows you to create these special REUSE.toml files with semantics for defining the copyright and license info for a group of files all at once, without having to annotate each one.  In this case, this REUSE.toml file defines that all files in the "p4" subdirectory beneath the "py" directory (where this REUSE.toml file is located) have the indicated copyright and license info.  My understanding is that all files in py/p4 and its subdirectories are automatically generated, so this REUSE.toml file is one way to annotate them all at once, without having to worry about ensuring that the auto-generation process inserts correct copyright and license annotations in each auto-generated file.

We could, of course, attempt to do that insertion of such annotations in every auto-generated file, if you know how, but we can also save that for later if desired.

rust/Cargo.lock.license
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: 2024 Campbell He <kp.campbell.he@duskmoon314.com>
```
jafingerhut, line 1: There are several files added that have ".license" as a suffix in their name.  The REUSE software looks at the contents of such files to check that the corresponding file without the ".license" suffix has a copyright and license annotation.

This is useful primarily for two cases:
(a) the file has no comment syntax defined, e.g. JSON data files
(b) the file is auto-generated, and modifying the auto-generation program to insert copyright and license annotations is impossible or difficult.

Case (b) is the case for this Cargo.lock file.

chrispsommers (approved):

matthewtlam (approved):

smolkaj (approved):

py/REUSE.toml
chrispsommers: Noted

smolkaj: SGTM. Could we add a brief comment giving this justification (files are auto generated)?

rust/Cargo.lock.license
smolkaj: Got it, makes sense.

PR #615, opened 2026-06-12
Add CI check to validate the presence of copyright and license annotations

jafingerhut:
in all files of the repo.  Also add a comment to py/REUSE.toml explaining why it exists.


py/REUSE.toml
```diff
@@ -1,3 +1,25 @@
+# This file is used by the `reuse` tool for copyright and license
```
jafingerhut, line 1: @smolkaj Please take a look at this comment to see if it addresses what you hoped it would.

smolkaj: Beautiful, thanks!
Honestly, far superior to what I envisioned :)

smolkaj (approved):

PR #620, opened 2026-07-06
codegen: use command -v instead of fragile which + \$? pattern

Devansh-567:
## Description
Refactored `codegen/compile_protos.sh` to use `command -v` instead of the fragile `which` + `$?` pattern. 

The previous implementation relied on checking `$?` immediately after a variable assignment (e.g., `PROTOC="$(which protoc)"`). While functional, this pattern is highly fragile as any commands accidentally inserted between the assignment and the `if` block would silently break the error handling. Additionally, `command -v` is the POSIX-compliant standard for binary verification, offering better portability than the external `which` command.

## Changes
- Replaced `which` lookups with direct short-circuit evaluation (`command -v ... || { ... }`).
- Removed fragile `$?` status checks following variable assignments.
- Reduced overall line count while preserving identical functional behavior.

### DIFF
--- a/codegen/compile_protos.sh
+++ b/codegen/compile_protos.sh
@@ -25,24 +25,17 @@ if [ "$#" -ne 1 ]; then
 fi
 
 BUILD_DIR="$1"
-PROTOC="$(which protoc)"
-if [ "$?" -ne 0 ]; then
-    echo "Could not find protoc"
-    exit 2
-fi
+
+command -v protoc >/dev/null 2>&1 || { echo "Could not find protoc"; exit 2; }
+PROTOC="$(command -v protoc)"
 
 echo "Using $PROTOC"
 
-GRPC_CPP_PLUGIN="$(which grpc_cpp_plugin)"
-if [ "$?" -ne 0 ]; then
-    echo "Could not find CPP protoc plugin"
-    exit 2
-fi
-GRPC_PY_PLUGIN="$(which grpc_python_plugin)"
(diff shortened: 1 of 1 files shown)


fruffy (approved):

chreekat (commented):
I think someone could just as easily argue for the reverse patch (e.g. "an if statement is a better indication of intent and more robust to accidentally adding more binary operators"). 

There is no bug being fixed.

So, I don't think this change passes the high bar that should be put in front of *any* change. After all, every commit may introduce bugs, may accidentally change behavior, and complicates history.

Devansh-567:
> I think someone could just as easily argue for the reverse patch (e.g. "an if statement is a better indication of intent and more robust to accidentally adding more binary operators").
> 
> There is no bug being fixed.
> 
> So, I don't think this change passes the high bar that should be put in front of _any_ change. After all, every commit may introduce bugs, may accidentally change behavior, and complicates history.

Fair, there's no active bug here, and I could have framed this better. The concrete motivation is that which isn't POSIX and its behavior/availability isn't guaranteed across environments the way command -v is; that's the main thing I'd want to fix vs. a style preference. If that's not enough of a bar for this repo, happy to close this, no strong objection either way. 😊

chreekat:
I'd like to keep the if + $? check, but  `command -v` instead of `which` would be fine :)

chrispsommers (approved):
LGTM

PR #622, opened 2026-07-07
proto: add curl fallback and validate downloaded status.proto

Devansh-567:
## Summary

Improve `proto/update_google_rpc_status.sh` by making the download process more robust.

### Changes

* Prefer `curl` when available and fall back to `wget`
* Exit with a clear error if neither download tool is installed
* Download to a temporary file before replacing the checked-in `status.proto`
* Add a basic sanity check to verify the downloaded file is a valid `.proto` before overwriting the existing file

This helps prevent invalid downloads from replacing `status.proto` and aligns the script with the download fallback pattern used elsewhere in the repository.

### DIFF
--- a/proto/update_google_rpc_status.sh
+++ b/proto/update_google_rpc_status.sh
@@ -22,9 +22,29 @@
 # The file is very stable and hasn't changed in many years, but we enforce
 # freshness of this copy using a GitHub Actions workflow just in case.
 
-set -e
+set -eu
 
 STATUS_PROTO_URL="https://raw.githubusercontent.com/googleapis/googleapis/refs/heads/master/google/rpc/status.proto"
 THIS_DIR="$(cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
+DEST_FILE="$THIS_DIR/google/rpc/status.proto"
 
-wget "$STATUS_PROTO_URL" -O "$THIS_DIR/google/rpc/status.proto"
+TMP_FILE="$(mktemp)"
+trap 'rm -f "$TMP_FILE"' EXIT
+
+if command -v curl >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


matthewtlam:
Could you explain why curl over wget is more robust?

chrispsommers:
What motivated this PR, did you experience real failures or is this just trying to incorporate best practices?

Devansh-567:
> Could you explain why curl over wget is more robust?

Thanks for the question! To clarify, the change isn't about curl being inherently more robust than wget, both are reliable tools. The intent is: curl is preferred first because it's already the more commonly available tool in minimal CI/container images (and is used elsewhere in this repo's scripts), so preferring it keeps behavior consistent across the codebase. wget is kept as a fallback so the script still works on systems where only wget is installed. If neither tool is present, the script now fails fast with a clear error instead of silently proceeding or failing with a confusing message later on.

So this is less "curl > wget" and more "support both, prefer the one used elsewhere in the repo, and fail clearly if neither exists."

Devansh-567:
> What motivated this PR, did you experience real failures or is this just trying to incorporate best practices?

@chrispsommers, this wasn't prompted by a specific failure I hit personally. It's a defensive/best-practices improvement:

The original script assumed curl was always present, which isn't guaranteed on all dev/CI environments. It also overwrote status.proto directly during download, so a partial/failed download (e.g., network hiccup, HTTP error page instead of the proto) could silently corrupt the checked-in file. This PR downloads to a temp file first and does a basic sanity check (that it looks like a valid .proto file) before replacing the real file, so a bad download fails loudly instead of quietly breaking the repo. I don't have a specific incident to point to, but this pattern (fallback + validate-before-replace) is already used elsewhere in the repo, so this brings update_google_rpc_status.sh in line with that convention. Happy to add more detail to the PR description if that's helpful for review. 😊

chrispsommers (approved):
LGTM thanks

PR #623, opened 2026-07-08
ci: standardize actions/checkout to v5 across all workflows

Devansh-567:
## Summary

This repo's GitHub Actions workflows pinned `actions/checkout` to several different major versions inconsistently:

- `@v3` - `any-branch-uploads.yml`, `main-branch-uploads.yml`, `tag-uploads.yml` (x2), `codegen.yml`, `spec.yml` (asciidoc-lint, build-spec jobs)
- `@v4` - `bazel-build.yml`
- `@v5` - `rust-build.yml`, `google-rpc-status-synced.yml`
- `@v6` - `spec.yml` (verify-licenses job)

This PR standardizes all `actions/checkout` references to a single version: **`@v5`**.

## Why v5

- Already in use and working in this repo (`rust-build.yml`, `google-rpc-status-synced.yml`), so it's a proven baseline rather than an untested jump.
- Recent and actively supported, without being the very newly released `v7` (GA'd June 17, 2026), which introduces new default behavior for `pull_request_target`/`workflow_run` checkouts. None of this repo's workflows use those triggers, so it wouldn't have caused issues either way, but `v5` avoids being an early adopter of a days-old major version in CI.

## Changes

Purely mechanical version bumps, no other changes:

- `any-branch-uploads.yml`: `v3` -> `v5`
- `bazel-build.yml`: `v4` -> `v5`
- `codegen.yml`: `v3` -> `v5`
- `main-branch-uploads.yml`: `v3` -> `v5`
- `spec.yml`: `v6` -> `v5` (verify-licenses), `v3` -> `v5` (asciidoc-lint), `v3` -> `v5` (build-spec)
- `tag-uploads.yml`: `v3` -> `v5` (x2)

`rust-build.yml` and `google-rpc-status-synced.yml` are unchanged as they already used `v5`.

### DIFF
--- a/.github/workflows/spec.yml
+++ b/.github/workflows/spec.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Validate that all files have copyright and license info
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v6
       - name: Install dev dependencies
         run: |
           ./CI/install-uv.sh
--- a/.github/workflows/spec.yml
+++ b/.github/workflows/spec.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Validate that all files have copyright and license info
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v5
       - name: Install dev dependencies
         run: |
           ./CI/install-uv.sh
@@ -30,7 +30,7 @@ jobs:
   asciidoc-lint:
     runs-on: [ubuntu-latest]
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v5
     - name: Run linter
       run: |
         ./tools/asciidoclint.py docs/v1/P4Runtime-Spec.adoc
@@ -38,7 +38,7 @@ jobs:
   build-spec:
     runs-on: [ubuntu-latest]
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v5
     - name: Build spec
       run: |
         make -C docs/tools/
--- a/.github/workflows/any-branch-uploads.yml
+++ b/.github/workflows/any-branch-uploads.yml
@@ -14,7 +14,7 @@ jobs:
     if: ${{ github.repository == 'p4lang/p4runtime' }}
     runs-on: [ubuntu-latest]
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v5
     - name: Build spec
       run: |
         make -C docs/tools/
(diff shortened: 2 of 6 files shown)


.github/workflows/spec.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Validate that all files have copyright and license info
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@v5
```
matthewtlam: Could you explain why you are downgrading this from v6 --> v5? Are there any downgrade implications? I assume it's just for consistency

Devansh-567: > Could you explain why you are downgrading this from v6 --> v5? Are there any downgrade implications? I assume it's just for consistency

Yes 😊, this is purely for consistency, not because v6 has a known issue.
On downgrade implications: I checked the actions/checkout release notes between v5 and v6, and there's no breaking change or behavior difference relevant to this workflow's verify-licenses job (it's a straightforward checkout, no special inputs like persist-credentials, sparse-checkout, or ref handling that changed between versions). So functionally this job will behave identically on v5.
The reasoning for picking v5 as the repo-wide standard rather than bumping everything to v6 (or v7):

v5 is already proven in this repo (rust-build.yml, google-rpc-status-synced.yml use it today), so standardizing on it means every workflow converges on a version that's already known to work here, rather than introducing a new version through this PR.
v6/v7 aren't broken or risky, but since none of the workflows need anything from those newer releases, there was no upside to bumping up to v6 across the board versus consolidating down to the version already validated in-repo.

matthewtlam: Can we make all the workflow versions go to v6 instead? I feel like a downgrade isn't necessary if v6 isn't risky

jafingerhut: Agreed that it seems a bit odd to intentionally downgrade the version already in use by a prominent part of the CI checks for the repo.

Devansh-567: Done

jafingerhut: So if we are going to the trouble to update some of the Github action versions, is there a reason to avoid upgrading them all to v6 while we are at it?  Maybe that will not work for some reason, but I am curious to know.

matthewtlam: +1, why are we not upgrading to v6 for all of them. Similarly, the PR description and title no longer match what is happening in this PR if one is v6

chrispsommers: @Devansh-567 gentle bump - can you reply? Thanks.

Devansh-567: I'm going to close this PR for now as I'm currently dealing with some personal matters. As soon as things settle down, I'll be back to pick this up and update everything to v6.

Thank you for your patience, understanding, and apologies for the delay!

PR #625, opened 2026-07-15
Update Python version and dependencies in tag-uploads workflow

Devansh-567:
## Summary
 
Small maintenance PR bumping stale CI and Python packaging version floors. No functional/behavioral changes, versions only.
 
### 1. `.github/workflows/tag-uploads.yml`
- `actions/setup-python@v3` = `actions/setup-python@v5`
`v3` runs on the deprecated Node16 runtime, which GitHub Actions now flags with deprecation warnings on every run. `v5` is already what the rest of the ecosystem (and most other actions in this repo) use, so this just brings the `publish-to-pypi` job in line.
 
### 2. `py/setup.cfg`
- `python_requires = >=3.6` = `>=3.8`
Python 3.6 reached end-of-life in December 2021 and isn't installable on current GitHub-hosted runner images. The declared floor was implying support that doesn't actually exist in practice. `3.8` reflects the oldest version that's realistically still testable/installable today.
 
- `protobuf >= 3.6.1` = `protobuf >= 4.21.6`
- `grpcio >= 1.17.2` = `grpcio >= 1.49.1`
Both floors were set years behind what the package is actually built and tested against. Raising them to versions that are actually exercised in CI avoids advertising compatibility that isn't real and avoids users pinning ancient, insecure versions based on the stated minimum.

### DIFF
--- a/py/setup.cfg
+++ b/py/setup.cfg
@@ -20,10 +20,10 @@ classifiers =
 [options]
 packages = find:
 platforms = any
-python_requires = >=3.6
+python_requires = >=3.8
 setup_requires =
     setuptools_scm
 install_requires =
-    protobuf >= 3.6.1
-    grpcio >= 1.17.2
+    protobuf >= 4.21.6
+    grpcio >= 1.49.1
     googleapis-common-protos >= 1.52


matthewtlam (approved):
LGTM!

jafingerhut (approved):
LGTM

PR #626, opened 2026-07-15
Derive Go Docker tag from Dockerfile's GO_VERSION in update.sh

Devansh-567:
## Summary
 
Removes a hardcoded Go version literal in `codegen/update.sh` that could silently drift out of sync with `codegen/Dockerfile`. No functional/behavioral changes when versions already match, this only changes how the version is sourced.
 
## Problem
 
Three places currently encode a Go version, independently of each other:
 
- `go.mod`: `go 1.20`
- `codegen/Dockerfile`: `ARG GO_VERSION=1.20.5`
- `codegen/update.sh`: hardcoded `golang:1.20` for the `go mod tidy` step
They agree today, but nothing ties them together. Bumping the Dockerfile's `GO_VERSION` (e.g. for a Go security patch or minor upgrade) doesn't touch `update.sh`, so the `go mod tidy` step could end up running against a different Go version than the one CI actually builds/tests with, a subtle, easy-to-miss source of drift.
 
## Change
 
`codegen/update.sh` now reads `GO_VERSION` directly out of `codegen/Dockerfile`:
 
```bash
GO_VERSION="$(sed -n 's/^ARG GO_VERSION=//p' codegen/Dockerfile)"
```
 
and uses it for the `go mod tidy` container instead of the hardcoded `golang:1.20` tag. A guard checks the value was actually found and fails loudly (instead of silently running `go mod tidy` under a stale/wrong image) if the Dockerfile's `ARG` line is ever renamed or removed.

### DIFF
--- a/codegen/update.sh
+++ b/codegen/update.sh
@@ -24,6 +24,14 @@ pushd "$THIS_DIR/.." >/dev/null
 
 docker build -t p4runtime-ci -f codegen/Dockerfile .
 
+# Read the Go version from the Dockerfile so this script and the CI image
+# can't drift out of sync with each other.
+GO_VERSION="$(sed -n 's/^ARG GO_VERSION=//p' codegen/Dockerfile)"
+if [ -z "$GO_VERSION" ]; then
+    echo "Could not determine GO_VERSION from codegen/Dockerfile" >&2
+    exit 1
+fi
+
 tmpdir="$(mktemp -d /tmp/p4rt.XXXXXX)"
 
 docker run --rm \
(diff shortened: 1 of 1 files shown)


chrispsommers:
@Devansh-567 Thank you for this PR. Do you think it'd "fix" https://github.com/p4lang/p4runtime/pull/627 and https://github.com/p4lang/p4runtime/pull/618?

Devansh-567:
> @Devansh-567 Thank you for this PR. Do you think it'd "fix" #627 and #618?

No, this PR doesn't address either of those. This PR specifically fixes the Go version drift in codegen/update.sh by dynamically deriving the Go version from codegen/Dockerfile (instead of using the hardcoded 1.20).

It doesn't cover:
- The AsciiDoc linter failures in docs/v1/P4Runtime-Spec.adoc (line length limits and missing backticks).
- The google.golang.org/grpc dependency bump (handled in #627).

chrispsommers (approved):

PR #628, opened 2026-08-18
tools: fix asciidoclint checks and add tests

Devansh-567:
## Problem

check_line_wraps() and check_keywords() in tools/asciidoclint.py are both gated by ContextAfterTitle, which only starts checking lines after it sees a line matching ^ *\[TITLE\] *$. That marker does not appear anywhere in the current docs/v1/P4Runtime-Spec.adoc (7,159 lines), I checked. Because ContextCompose short-circuits on the first False, this means the checker never actually enters either check on the real document, even though asciidoclint.py is run in CI on every PR (asciidoc-lint job) and is documented in CONTRIBUTING.md as the gate that "keep the spec uniform."

I confirmed this by running the tool directly:

    $ ./tools/asciidoclint.py docs/v1/P4Runtime-Spec.adoc
    Errors found: 0

Then appending a 120-character line (no URL in it) and a bare NOT_FOUND (no backticks) to a copy of the file and re-running still `Errors found: 0`.

There's a second, related bug in the same area: ContextSkipHeadings looks for Markdown-style headings (^ *#), but this document uses real AsciiDoc heading syntax (^=) 171 lines in the spec start with =, only 2 start with #. So fixing the title-detection bug on its own would immediately start flagging every long section heading as a line-wrap violation unless this is fixed too.

## Fix

- ContextAfterTitle now matches the actual document title line (^= , e.g. = P4Runtime Specification) instead of the obsolete
  [TITLE] sentinel.
- ContextSkipHeadings now matches AsciiDoc headings (^=+ ) instead of Markdown-style # headings.
- Both regexes are now raw strings, which also fixes two SyntaxWarning: invalid escape sequence warnings on import.

## Tests

Added tools/test_asciidoclint.py (no test suite existed for this file before). It runs the checker end-to-end against small synthetic .adoc fixtures rather than testing internals directly, so it also acts as a guard against the checks silently becoming no-ops again the way they did here. Covers:

- a long line after the title is flagged
- a long line in document metadata before the title is correctly ignored
- an un-backticked known keyword after the title is flagged; the correctly-backticked form is not
- a long AsciiDoc heading (=== ...) is skipped
- a long body line is still flagged even after a heading
- a #-prefixed line is not mistaken for a heading

I verified all 8 tests fail against the pre-fix code and pass against this fix.

Wired python3 tools/test_asciidoclint.py -v into the asciidoc-lint
CI job so this can't silently regress again.

Once the checks are actually running, they flag 171 pre-existing spots in the current spec text that were never caught while the checker was broken: 159 lines over 80 characters and 12 un-backticked keyword occurrences.

### DIFF
--- a/.github/workflows/spec.yml
+++ b/.github/workflows/spec.yml
@@ -31,6 +31,9 @@ jobs:
     runs-on: [ubuntu-latest]
     steps:
     - uses: actions/checkout@v3
+    - name: Run linter unit tests
+      run: |
+        python3 tools/test_asciidoclint.py -v
     - name: Run linter
       run: |
         ./tools/asciidoclint.py docs/v1/P4Runtime-Spec.adoc
--- a/tools/asciidoclint.py
+++ b/tools/asciidoclint.py
@@ -128,7 +128,7 @@ class ContextSkipBlocks(Context):
     Block = namedtuple('Block', ['num_tildes', 'name'])
 
     def __init__(self):
-        self.p_block = re.compile('^ *(?P<tildes>~+) *(?:(?P<cmd>Begin|End)(?: +))?(?P<name>\w+)?')
+        self.p_block = re.compile(r'^ *(?P<tildes>~+) *(?:(?P<cmd>Begin|End)(?: +))?(?P<name>\w+)?')
         self.blocks_stack = []
 
     def enter(self, line, filename, lineno):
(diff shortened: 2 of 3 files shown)


Devansh-567:
Closing this for now, while digging into the CI failure here, I found a related bug in ContextSkipBlocks (it only recognizes ~~~~-delimited blocks, but this document exclusively uses ---- listing blocks, |=== tables, and [...] block-attribute lines, so none of those were actually being skipped either). Rather than layer a second unrelated fix on top of this PR's diff, I'm going to fold both fixes into a single new PR along with the doc cleanup needed to get CI green, and submit that instead.

Will link the new PR here once it's up.

PR #639, opened 2026-09-18
ci: standardize actions/checkout to v6

Devansh-567:
This is a follow-up to PR #623, which I closed earlier while dealing with health issues and was unable to continue with the requested updates at the time.

I'm reopening the change now with the requested updates so it can be reviewed again.

### What changed

* Updated all `actions/checkout` references across the GitHub Actions workflows to `v6`.
* Kept the change limited to the workflow checkout versions.

The original PR discussion asked to use `actions/checkout@v6` rather than downgrading to an older version, so this replacement PR follows that direction.

This replaces the previously closed PR #623.

Thanks for taking another look.


jafingerhut (approved):
looks reasonable to me

